PR #2, opened 2022-12-31
Add `bors` binary with basic GitHub webhook parsing

Kobzol:
The server doesn't really do anything yet, it just parses and validates GitHub webhooks.

### DIFF
--- a/src/bin/bors.rs
+++ b/src/bin/bors.rs
@@ -1,0 +1,64 @@
+use std::net::SocketAddr;
+use std::sync::Arc;
+
+use anyhow::Context;
+use axum::http::StatusCode;
+use axum::response::IntoResponse;
+use axum::routing::post;
+use axum::Router;
+use clap::Parser;
+use tower::limit::ConcurrencyLimitLayer;
+
+use bors::config::Config;
+use bors::github::webhook::GitHubWebhook;
+
+#[derive(clap::Parser)]
+struct Opts {
+    /// Secret used to authenticate webhooks.
+    #[arg(long, env = "WEBHOOK_SECRET")]
+    webhook_secret: String,
+}
+
+async fn github_webhook(GitHubWebhook(event): GitHubWebhook) -> impl IntoResponse {
+    println!("Received GitHub webhook: {:?}", event);
+    (StatusCode::OK, "")
+}
+
+async fn server(config: Config) -> anyhow::Result<()> {
+    let state = Arc::new(config);
+
+    let app = Router::new()
+        .route("/github", post(github_webhook))
+        .layer(ConcurrencyLimitLayer::new(100))
+        .with_state(state);
+    let addr = SocketAddr::from(([0, 0, 0, 0], 8080));
+
+    axum::Server::bind(&addr)
+        .serve(app.into_make_service())
+        .await?;
+    Ok(())
+}
+
+fn try_main(opts: Opts) -> anyhow::Result<()> {
+    let config = Config::new(opts.webhook_secret);
+
+    let runtime = tokio::runtime::Builder::new_multi_thread()
+        .enable_all()
+        .build()
+        .context("Cannot build tokio runtime")?;
+
+    let server_fut = runtime.spawn(server(config));
+    runtime.block_on(server_fut)??;
+
+    Ok(())
+}
+
+fn main() {
+    env_logger::init();
+
+    let opts = Opts::parse();
+    if let Err(error) = try_main(opts) {
+        eprintln!("Error: {error:?}");
+        std::process::exit(1);
+    }
+}
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,0 +1,15 @@
+use secrecy::SecretString;
+
+/// Configuration of `bors` used to communicate with GitHub.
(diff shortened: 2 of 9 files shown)


config.toml.sample
```diff
@@ -0,0 +1,12 @@
+[github]
+# A GitHub personal access token.
+access_token = "<token>"
+
+# A GitHub oauth application for this instance of homu:
+app_client_id = "<id>"
+app_client_secret = "<secret>"
```
pietroalbini, line 7: In production I'd prefer we use environment variables for secrets rather than writing them in a TOML configuration file. That's the easiest way in our infrastructure to pass secrets into the container.

Options could be:

1. Do the same thing homu does, custom support for variable interpolation in the configuration file:

   ```toml
   access_token = "${GITHUB_TOKEN}"
   ```

2. Use a serde untagged enum, like Cargo does for version numbers:

   ```toml
   access_token = { env = "GITHUB_TOKEN" }
   ```

3. Keep the repository configuration in the file (maybe? see the GitHub App comment) but use only environmment variables (with `dotenvy` for local use) for secrets.

I'm personally leaning towards 3), preferring it over to 2). I don't think we should go with option 1) as it's more fragile.

pietroalbini, line 7: Homu used GitHub OAuth to authenticate, but is there a reason why we're not using a GitHub App? That provides granular permissions, configures webhooks and handles webhook secrets automatically for us.

With that, environment variables, and a `rust-bors.toml` at the root of each repository (similar to `triagebot.toml`) there should be no need to have a configuration file at all.

src/bin/bors.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    Ok(())
+}
+
+fn main() {
+    env_logger::init();
+
+    let opts = Opts::parse();
+    if let Err(error) = try_main(opts) {
+        eprintln!("Error: {error:?}");
```
pietroalbini, line 61: This returns with exit code 0 if there is an error.

src/config.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+#[derive(serde::Deserialize, Debug, Clone)]
+pub struct GithubConfig {
+    /// GitHub personal access token.
+    pub access_token: String,
+    /// GitHub OAuth application client ID
+    pub app_client_id: String,
+    /// GitHub OAuth application client secret
+    pub app_client_secret: String,
```
pietroalbini, line 26: We could use the https://docs.rs/secrecy/latest/secrecy/ crate to avoid accidentally leaking secrets.

pietroalbini (changes requested):

config.toml.sample
Kobzol: I *think* that @Mark-Simulacrum mentioned something in the sense that it would be nice if we could manage the deployment ourselves (I assume that GitHub apps run on GH infrastructure, but I may be wrong, I have never used them).

Kobzol: I assume that we will use a single, shared bors instance for all rust-XXX repositories, therefore passing lists in environment variables would probably be quite unwieldy. I'm not sure how 2) differs from 3). I assume that we would need to name the used environment variables in the config file and then pass them to the program? We could maybe auto-derive the env. name from the repository configuration in the file to avoid even mentioning the env. variable in the file.

pietroalbini: GitHub Apps are just a different way to authenticate with the API, it'd still be hosted by us. The main differences for us would be:

* Rather than setting up webhooks in each repository, there is a single webhook for the app that gets events for all the repositories the app is installed on.
* Permissions are more granular than personal access tokens.
* Enabling bors on a repository would just be installing the app on the new repository too, rather than having to add the bors user to the list of users with write access and configuring an ad-hoc webhook.

See https://docs.github.com/en/developers/apps/getting-started-with-apps/about-apps.

pietroalbini: Yeah we'll use a single bors instance for all repos.

If we do not use a GitHub App, then I'd auto-derive the env like `BORS_WEBHOOK_SECRET_rust-clippy`. If we use a GitHub App, there would be a single webhook (and thus a single secret) for all the repos, with no need to deriving variable names.

At that point though we could just configure everything with environment variables, as the list of repositories bors should care about could be fetched from GitHub (list of installed apps).

Mark-Simulacrum: FWIW the webhook could just be a global one across the org (triagebot does that). But I think preferring the GitHub app is definitely better.

src/config.rs
Kobzol: Good idea, I added it.

src/bin/bors.rs
Kobzol: Fixed, thanks.

Kobzol:
I agree that using a GitHub app is the way forward. I think that with an app, we don't even need to store (nor fetch) a list of activated repositories (at least not for authentication itself). If the (single, shared) webhook secret of the incoming webhook is validated, then the bors app should be installed on the repository.

I simplified the PR to only validate the webhook via a provided secret. I will add further authentication tokens once there is need for it (e.g. to send comments on PRs).

PR #3, opened 2023-01-04
Implement basic command handling of `bors ping`

Kobzol:
This PR adds basic Github app functionality.

Once the server is started, all repositories tied to the github app are discovered and stored in a hash map. Then an async service is started, which receives Bors commands and executes them, by responding with a comment on a given GitHub repository.

The http server -> service communication is performed using a tokio channel, it could also be done in a simpler way, if the webhook handler just took a mutex lock of a shared instance of the bors service, but I don't think it's that important at this moment.

Bors comments are marked with a HTML comment that lets us know that a comment was made by us, to repeat endless comment cycles. I wasn't able to find out easily what is the bot user ID of the github app account for a given repository (I spent a lot of time on this - if someone knows how to do this, I'd be glad for any inspiration), so this seemed like the next easiest way to find out if a comment was made by the bot.

If you think that it is worth it, I can replace the GitHub communication backend in the GH service with some trait, so that it can be mocked, and then we could write (integration?) tests for the GH logic.


Mark-Simulacrum:
> If you think that it is worth it, I can replace the GitHub communication backend in the GH service with some trait, so that it can be mocked, and then we could write (integration?) tests for the GH logic.

I think this may be worth it -- I'm less worried about testing *comments*, but I do think we want to think early on about testing the approval/merge/etc logic, which probably at least for the majority of tests doing it "locally" (rather than against the GH API) will be faster/easier.

Mark-Simulacrum:
Then I don't think that it is critical we get that tested; comments themselves aren't likely to be interesting.

src/bin/bors.rs
```diff
@@ -1,34 +1,65 @@
...
 struct Opts {
     /// Secret used to authenticate webhooks.
     #[arg(long, env = "WEBHOOK_SECRET")]
     webhook_secret: String,
+    /// Github App ID.
+    #[arg(long, env = "APP_ID")]
+    app_id: u64,
+    /// Path to a private key used to authenticate as a Github App.
+    #[arg(long, env = "PRIVATE_KEY_PATH")]
+    private_key_path: PathBuf,
```
Mark-Simulacrum, line 29: I think in production this probably will be an env variable, not a path to some file.

Mark-Simulacrum (approved):

src/bin/bors.rs
pietroalbini: Yeah, the private key content will be in the environment variable.

src/github/api.rs
```diff
@@ -0,0 +1,102 @@
...
+
+                // Load repositories of this installation
+                let repos: InstallationRepositories = installation_client
+                    .get::<_, _, ()>(repositories_url, None)
+                    .await?;
+                for repo in repos.repositories {
+                    let Some(owner) = repo.owner else { continue; };
+                    let key = GitHubRepositoryKey::new(&owner.login, &repo.name);
+                    let access = Repository {
+                        client: installation_client.clone(),
```
pietroalbini, line 51: Since I had to look it up, octocrab refreshes expired tokens automatically under the hood.

src/github/api.rs
```diff
@@ -0,0 +1,102 @@
...
+use octocrab::models::{AppId, InstallationRepositories};
+use octocrab::Octocrab;
+use secrecy::{ExposeSecret, SecretVec};
+
+use crate::github::GitHubRepositoryKey;
+
+/// Provides access to managed GitHub repositories.
+#[derive(Debug)]
+pub struct RepositoryAccess {
+    repositories: HashMap<GitHubRepositoryKey, Repository>,
```
pietroalbini, line 14: This should be a `RwLock<T>`, and we should listen to the `installation`/`installation_repositories` (not sure which are sent when) webhooks to update the `HashMap` when an installation is added or removed. I wouldn't want to restart the application every time we change the set of repositories bors uses.

pietroalbini (changes requested):
> I wasn't able to find out easily what is the bot user ID of the github app account for a given repository (I spent a lot of time on this - if someone knows how to do this, I'd be glad for any inspiration), so this seemed like the next easiest way to find out if a comment was made by the bot.

While for a normal user, the user's `html_url` is `https://github.com/username`, for GitHub App users the `html_url` is `https://github.com/apps/appname`. We could call `https://api.github.com/app` at startup to get the name of the app, and then compare the commenter's `html_url` with the cached one.

Kobzol:
> While for a normal user, the user's html_url is https://github.com/username, for GitHub App users the html_url is https://github.com/apps/appname. We could call https://api.github.com/app at startup to get the name of the app, and then compare the commenter's html_url with the cached one.

Seems quite hacky, but if there's not a better way, I can try it.

src/bin/bors.rs
```diff
@@ -1,34 +1,65 @@
...
+    webhook_secret: WebhookSecret,
 }
 
-async fn server(config: Config) -> anyhow::Result<()> {
-    let state = Arc::new(config);
+impl AsRef<WebhookSecret> for ServerState {
+    fn as_ref(&self) -> &WebhookSecret {
+        &self.webhook_secret
+    }
+}
```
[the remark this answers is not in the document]
Kobzol: The `AsRef` is here so that the webhook axum extractor doesn't need to know specifically how does the state look like. I can implement the extractor with the knowledge of the server state if you want.

src/github/api.rs
Kobzol: It does indeed, that is why I also cache these clients. I will add a comment to make it obvious.

src/github/api.rs
Kobzol: I considered "adding a new repository" to be a super rare event, but I agree that it would be nice if it could refresh automatically. I will add it in a future PR if you don't mind, to keep this one smaller.

pietroalbini: It would indeed be a super rare event, but not having to restart the instance would simplify the `sync-team` code a ton. I'm also totally ok doing it in a future PR, I agree we should keep PRs small-ish.

src/bin/bors.rs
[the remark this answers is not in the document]
pietroalbini: I'd implement the extractor with the knowledge of the server state, yeah.

src/github/api.rs
```diff
@@ -0,0 +1,102 @@
...
+/// Provides access to a single app installation (repository).
+#[derive(Debug)]
+pub struct Repository {
+    client: Octocrab,
+    key: GitHubRepositoryKey,
+}
+
+impl Repository {
+    /// Returns true if the comment was made by this bot.
+    pub fn is_comment_internal(&self, comment: &Comment) -> bool {
```
rylev, line 85: I find it odd that this is a method on `Repository`, since it doesn't need to be.

src/github/service.rs
```diff
@@ -0,0 +1,104 @@
...
+
+async fn handle_event(repo: &Repository, message: WebhookContent) -> anyhow::Result<()> {
+    match message.event {
+        WebhookEvent::Comment(event) => {
+            // We only care about pull request comments
+            if event.issue.pull_request.is_none() {
+                return Ok(());
+            }
+            // We want to ignore comments made by this bot
+            if repo.is_comment_internal(&event.comment) {
```
rylev, line 46: Might make sense to add `trace!`s here.

src/github/service.rs
```diff
@@ -0,0 +1,104 @@
...
+            log::info!(
+                "Received comment at https://github.com/{}/{}/issues/{}, commands: {:?}",
+                repo.repository().owner,
+                repo.repository().name,
+                pr_number,
+                commands
+            );
+            for command in commands {
+                match command {
+                    Ok(command) => match command {
```
rylev, line 64: `Ok(BorsCommand::Ping) =>`

src/github/service.rs
```diff
@@ -0,0 +1,104 @@
...
+                        BorsCommand::Ping => {
+                            execute_bors_command(repo, command, pr_number)
+                                .await
+                                .context("Cannot execute bors command")?;
```
rylev, line 68: Nit: errors in past tense
```suggestion
                                .context("Could not execute bors command")?;
```

src/github/service.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+async fn execute_bors_command(
+    repo: &Repository,
+    command: BorsCommand,
+    pr_number: u64,
+) -> anyhow::Result<()> {
+    match command {
+        BorsCommand::Ping => {
+            log::info!("Executing ping");
```
rylev, line 98: Is `info` the right level or `debug`?

Kobzol:
Addressed review comments.

pietroalbini (approved):

PR #5, opened 2023-02-04
Add GH API merge functionality

Kobzol:
Currently when the user executes `@bors try`, bors will just directly merge the PR :) I'll add integration with GH actions in a follow-up PR.


src/github/service/try_cmd.rs
```diff
@@ -0,0 +1,19 @@
...
+
+pub async fn try_command(repo: &RepositoryClient, pr: PullRequest) -> anyhow::Result<()> {
+    log::debug!("Executing try on {}/{}", repo.name(), pr.id);
+
+    // Just merge the PR directly right now. Let's hope that tests pass :)
+    merge_branches(
+        repo,
+        &pr.base.ref_field,
+        &pr.head.ref_field,
+        "Bors merge commit",
```
rylev, line 13: nit: can we already do the commit message bors uses?

> Auto merge of {number} - {author}:{branch_name}, r={reviewer}

src/github/api/mod.rs
```diff
@@ -116,18 +119,22 @@ pub struct RepositoryClient {
...
     client: Octocrab,
-    repository: GithubRepoName,
+    // We store the name separately, because repository has an optional owner, but at this point
+    // we must always have some owner of the repo.
+    repo_name: GithubRepoName,
+    repository: Repository,
 }
 
 impl RepositoryClient {
-    pub async fn post_pr_comment(&self, issue: u64, content: &str) -> octocrab::Result<Comment> {
```
rylev, line 123: I think it would be nice to still have this method that calls to the new free function.

src/github/api/operations.rs
```diff
@@ -0,0 +1,93 @@
+use octocrab::models::issues::Comment;
+use reqwest::StatusCode;
+
+use crate::github::api::RepositoryClient;
+
+#[derive(Debug)]
+pub enum MergeError {
+    NotFound,
+    Conflict,
+    AlreadyMerged,
```
rylev, line 10: Probably would be a good idea to hold on to the refs that could not be merged in this errors.

src/github/service/try_cmd.rs
Kobzol: Ok, I added the message and basic handling of merge errors.

src/github/api/mod.rs
Kobzol: I'd like to keep the client be a dumb holder of token/connection/repository ID. If there will be a lot of git operations we can put them behind into struct impl later.

src/github/api/operations.rs
Kobzol: The caller of the function passes the head/base refs, so I don't think that it's strictly needed here. I'll add them if there will be a callsite where it will be needed.

PR #6, opened 2023-02-05
Load repository config from `rust-bors.toml`

Kobzol:


src/permissions.rs
```diff
@@ -0,0 +1,111 @@
...
+    })
+}
+
+#[derive(serde::Deserialize)]
+struct UserPermissionsResponse {
+    github_users: Vec<String>,
+}
+
+/// Loads users that are allowed to perform try/review from the Rust Team API.
+async fn load_users(url: &str) -> anyhow::Result<Vec<String>> {
```
rylev, line 102: Nit: it feels strange that this takes the entire url instead of the parts that differ. Like this, the caller is required to pass the entire correct url instead of, for example, an enum which specifies which user permissions to fetch

src/permissions.rs
```diff
@@ -0,0 +1,111 @@
...
+        .map_err(|error| anyhow::anyhow!("Cannot load try users: {error:?}"))?;
+    Ok(UserPermissions {
+        review_users: review_users.into_iter().collect(),
+        try_users: try_users.into_iter().collect(),
+    })
+}
+
+#[derive(serde::Deserialize)]
+struct UserPermissionsResponse {
+    github_users: Vec<String>,
```
rylev, line 98: Can we deserialize directly to HashSets instead of Vec?

src/permissions.rs
```diff
@@ -0,0 +1,111 @@
+use std::collections::HashSet;
+use std::time::{Duration, SystemTime};
+
+use crate::github::GithubRepoName;
+
+/// For how long should the permissions be cached.
+const CACHE_DURATION: Duration = Duration::from_secs(60 * 5);
```
rylev, line 7: I think we could probably lower this further to a minute.

src/permissions.rs
Kobzol: Good point!

PR #9, opened 2023-04-06
Add branch merging functionality to try build

Kobzol:
Try build now resets the try branch to master and then merges the corresponding PR into it.


src/handlers/trybuild.rs
```diff
@@ -14,9 +19,73 @@ pub async fn command_try_build<Client: RepositoryClient, Perms: PermissionResolv
...
+    client
+        .set_branch_to_sha(TRY_BRANCH_NAME, &pr.base.sha)
+        .await?;
+    match client
+        .merge_branches(
+            TRY_BRANCH_NAME,
+            &pr.head.sha,
+            &format!("Merge {}", pr.number),
+        )
+        .await
```
Urgau, line 31: Should this have any protection against a race?

Kobzol: That is a good question, although I'm not really sure how we could do that. In general, there should be only a single bot running (per repository, at least), and it's single-threaded currently. If the `main` branch changes in the meantime, that's not such a problem for try builds.

For normal merges, it would be a problem, but if we assume that only the bot is able to perform merges, then it shouldn't be a problem. As long as it stays single-threaded, and as long as it is implemented correctly (the "merge queue" isn't implemented yet), then I think that we do not have to cares about these kinds of races.

Urgau: I here your arguments but wouldn't this scenario be still possible, where two users `bors try` nearly at the same time and due to latency in the request of `set_branch_to_sha` the second user could call it while the first one is still waiting his response ?
```
  User 1         User 2
    │              │
    │              │
    ▼              ▼
  bors try       bors try
     │              │
     │              │
     │              │
     │              │
     │              └──►set_branch_to_sha───────►merge_branch
     │                 │                │
     └───► set_branch_to_sha───────►merge_branch
         │             │   │            │
         └────┐ ┌──────┘   └─ ─┐  ┌─────┘
              │ │              │  │
            start req         end req


──────► time
```

Kobzol: No, with the current design, this cannot happen. Handling events and commands is strictly single-threaded. Until the handling of the first bors try command finishes, the second one will not begin executing.

The question is whether we want to keep this design, maybe we will want to allow some parallelism.

PR #11, opened 2023-05-04
Bootstrap database

Kobzol:
Uses `sqlx` and `sea-orm` to support migrations and single code across SQLite and Postgre.


Kobzol:
The database is hidden behind a trait, so even if we decide to change its implementation completely later, it shouldn't affect the rest of the code heavily. I'll use SeaORM for now.

PR #26, opened 2023-06-12
Modify parser to handle key value pairs

sladyn98:
This PR modifies the parser to handle key value pairs

### DIFF
--- a/src/bors/command/parser.rs
+++ b/src/bors/command/parser.rs
@@ -1,9 +1,9 @@
 //! Defines parsers for bors commands.
 
 use crate::bors::command::BorsCommand;
+use std::collections::HashMap;
 use std::iter::Peekable;
 use std::str::SplitWhitespace;
-use std::collections::HashMap;
 
 #[derive(Debug)]
 pub enum CommandParseError<'a> {
@@ -31,12 +31,13 @@ impl CommandParser {
         // The order of the parsers in the vector is important
         let parsers: Vec<fn(Tokenizer) -> ParseResult> =
             vec![parser_ping, parser_try_cancel, parser_try];
-        
+
         text.lines()
             .filter_map(|line| match line.find(&self.prefix) {
                 Some(index) => {
                     let command = &line[index + self.prefix.len()..];
-                    let (command_line, remaining) = command.split_once(' ').unwrap_or((command, ""));
+                    let (command_line, remaining) =
+                        command.split_once(' ').unwrap_or((command, ""));
                     // let key_value_pairs =  parse_key_value(remaining).unwrap_or_else(|_| HashMap::new());
                     let mut tokenizer = Tokenizer::new(command);
                     for parser in &parsers {
@@ -49,8 +50,7 @@ impl CommandParser {
                 None => None,
             })
             .collect()
-    } 
-    
+    }
 }
 
 struct Tokenizer<'a> {
@@ -74,8 +74,6 @@ impl<'a> Tokenizer<'a> {
 }
 
 type ParseResult<'a> = Option<Result<BorsCommand, CommandParseError<'a>>>;
-
-
 
 /// Parsers
 
@@ -134,6 +132,7 @@ fn parse_list<'a>(
     Some(Ok(result))
 }
 
+#[allow(unused)]
 pub fn parse_key_value(input: &str) -> Result<HashMap<String, String>, String> {
     let mut kv_pairs = HashMap::new();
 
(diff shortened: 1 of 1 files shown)


src/bors/command/parser.rs
```diff
@@ -70,6 +70,27 @@ impl<'a> Tokenizer<'a> {
 
 type ParseResult<'a> = Option<Result<BorsCommand, CommandParseError<'a>>>;
 
+
+//  parse a line of text into command and key-value pairs
+fn parse_key_value(line: &str) -> Option<(&str, Vec<(&str, &str)>)> {
```
sladyn98, line 75: @Kobzol We might also have to modify tokenizer and other functions to handle this change right ?

src/bors/command/parser.rs
```diff
@@ -30,19 +30,19 @@ impl CommandParser {
...
         let parsers: Vec<fn(Tokenizer) -> ParseResult> =
             vec![parser_ping, parser_try_cancel, parser_try];
-
+    
         text.lines()
-            .filter_map(|line| match line.find(&self.prefix) {
-                Some(index) => {
-                    let command = &line[index + self.prefix.len()..];
+            .filter_map(|line| {
+                if let Some((command, kv_pairs)) = parse_key_value(line) {
```
Kobzol, line 36: We still need to parse the bot prefix first, as before. Let's stick to the following command syntax for now:
```
@<prefix> <command> <key1>=<val1> <key2>=<val2>
@bors try
@bors try parent=foo
@bors try job=msvc parent=foo opts=debug,assertions
```
So first we need to look for the bot prefix (same as before), and if we find it, then try to parse a command and a list of key-value pairs after it.

src/bors/command/parser.rs
Kobzol: Yes. In fact I would rather get rid of the tokenizer and just write this using some parsing library (`chomsky`, `nom` etc.) to provide useful error messages. Or, we should implement error reporting by hand.

With the existing implementation, if there's a single "syntax error", the whole command will be just `None`, so we will not be able to provide useful feedback to users on what went wrong.

Kobzol, line 75: I think that it would make sense for this function to return the key-value pairs in a `HashMap<String, String>` (or `HashMap<&str, &str>`, doesn't really matter), the order shouldn't be important (at least we don't have a use-case for it being important now).

Kobzol (commented):
Hi, thanks for the PR! I left some comments with guidance on how to continue with it. It would be great if you could write a few unit tests in the `parser.rs` file that would test the `parse_key_value` functions, e.g. with the inputs that I have described in one of my other comments.

src/bors/command/parser.rs
```diff
@@ -127,10 +134,29 @@ fn parse_list<'a>(
     Some(Ok(result))
 }
 
+pub fn parse_key_value(input: &str) -> Result<HashMap<String, String>, String> {
```
Kobzol, line 137: You can add `#[allow(unused)]` on top of the function for now to resolve the warnings, I'll use it later in another PR.

src/bors/command/parser.rs
```diff
@@ -127,10 +134,29 @@ fn parse_list<'a>(
...
+
+    for pair in input.split_whitespace() {
+        if let Some((key, value)) = pair.split_once('=') {
+            kv_pairs.insert(key.to_string(), value.to_string());
+        }
+    }
+
+    if kv_pairs.is_empty() {
+        return Ok(HashMap::new());
+    }
```
Kobzol, line 148:
```suggestion
```
This is not needed, you can just return `Ok(kv_pairs)` directly.

Kobzol (commented):
Thanks! Please reformat the code using `cargo fmt` and resolve `cargo check/clippy` warnings and it's good to go, as an initial implementation.

src/bors/command/parser.rs
sladyn98: IMO we could go with hand written error reporting, would give us much more control

Kobzol: If you're up to it, feel free to implement it! :)

src/bors/command/parser.rs
sladyn98: Perfect !

src/bors/command/parser.rs
sladyn98: Perfect

Kobzol:
Thanks!

PR #44, opened 2024-03-06
feat: help command

vohoanglong0107:


Kobzol (changes requested):
Looks great! One small nit.

src/bors/handlers/help.rs
```diff
@@ -0,0 +1,34 @@
+use crate::bors::RepositoryClient;
+use crate::bors::RepositoryState;
+use crate::github::PullRequest;
+
+const HELP_MESSAGE: &str = r#"
+- try: Request that the PR be tested by buildbot, without accepting it.
```
Kobzol, line 6:
```suggestion
- try: Execute the `try` CI workflow on this PR (without approving it for a merge).
```

`buildbot` is actually the name of an existing CI service, which we do not use. This could be a bit confusing to users. We also can't just say to "test this PR on CI", because all PRs are tested on CI, just with different workflows xD So I suggest rewording a bit.

Kobzol (changes requested):
Once you reword the help text, could you please squash the commits? :) Let me know if you want help with that.

src/bors/handlers/help.rs
vohoanglong0107: Really sorry about that, I copy it from the old homu page without double checking it

Kobzol (approved):
Thank you! :heart:

PR #45, opened 2024-03-07
chore: bump rust version in Dockerfile

vohoanglong0107:
The current rust version in Dockerfile is too old, make it incompatible with cargo chef

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.70 as base
+FROM rust:1.76 as base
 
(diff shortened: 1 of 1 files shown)


Kobzol:
Thanks! It looks like there is a conflict. You will probably need to update your `bors:main` branch locally. I updated it yesterday, because it didn't actually contain the newest commits - the most recent branch was actually `staging`. It's a bit messy at the moment, because the repo is now designed to have two branches (`main` and `staging`), that will be deployed separately to two production environments.

Now I updated `main`, so it should hopefully be OK going forward.

Kobzol (approved):
Thank you! :)

Issue #29, opened 2023-06-25
Automatically select same parent as a previous try build


Kobzol:
The bot can perform a try build with a specific parent after https://github.com/rust-lang/bors/pull/28. It would be nice to specify something like `@bors try parent=<last>` to automatically do a try build where the parent commit will be based on the parent of the last try build (to achieve a more reproducible build, which is particularly useful for perf. runs).

If there was not previous try build, the bot should post a comment with an error message.

PR #48, opened 2024-03-14
feat: try last parent

vohoanglong0107:
# What

Implement `@bors try parent=last`, to use the last build parent as base.
Will post a comment to explain if no last build was found, and postpone.

Fixes: https://github.com/rust-lang/bors/issues/29

### DIFF
--- a/src/bors/command/mod.rs
+++ b/src/bors/command/mod.rs
@@ -2,6 +2,12 @@ mod parser;
 use crate::github::CommitSha;
 pub use parser::{CommandParseError, CommandParser};
 
+#[derive(Clone, Debug, PartialEq)]
+pub enum Parent {
+    CommitSha(CommitSha),
+    Last,
+}
+
 /// Bors command specified by a user.
 #[derive(Debug, PartialEq)]
 pub enum BorsCommand {
@@ -12,7 +18,7 @@ pub enum BorsCommand {
     /// Perform a try build.
     Try {
         /// Parent commit which should be used as the merge base.
-        parent: Option<CommitSha>,
+        parent: Option<Parent>,
     },
     /// Cancel a try build.
     TryCancel,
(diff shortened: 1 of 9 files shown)


src/bors/command/mod.rs
```diff
@@ -2,6 +2,12 @@ mod parser;
 use crate::github::CommitSha;
 pub use parser::{CommandParseError, CommandParser};
 
+#[derive(Clone, Debug, PartialEq)]
+pub enum Parent {
```
Kobzol, line 6: It would be nice to add some documentation here, to explain what `Last` means (i.e. the parent of the last try build on the same PR).

src/bors/handlers/trybuild.rs
```diff
@@ -52,10 +53,27 @@ pub(super) async fn command_try_build<Client: RepositoryClient>(
...
 
     let base_sha = match parent.clone() {
-        Some(parent) => parent,
+        Some(parent) => match parent {
+            Parent::Last => match last_parent {
+                None => {
+                    repo.client
+                        .post_comment(
+                            pr.number,
+                            ":exclamation: There was no previous build. Please set an explicit parent or remove the parent option to use a default value.",
```
Kobzol, line 68:
```suggestion
                            ":exclamation: There was no previous build. Please set an explicit parent or remove the `parent=last` argument to use the default parent.",
```

src/tests/state.rs
```diff
@@ -359,7 +359,16 @@ impl RepositoryClient for TestRepositoryClient {
     }
 
     async fn set_branch_to_sha(&mut self, branch: &str, sha: &CommitSha) -> anyhow::Result<()> {
-        self.add_branch_sha(branch, &sha.0);
+        let branch_history = self.branch_history.entry(branch.to_string()).or_default();
```
Kobzol, line 362: I don't think that this logic is needed, we should know how many resets have happened. Let's remove this logic and allow duplicates there (and modify the test so that it checks that the same SHA was really set twice).

Kobzol (commented):
Thanks a lot! Left a few nits.

I have to say that your Rust is reallly readable and easy to review! :)

Kobzol (approved):
Thank you!

PR #54, opened 2024-03-24
refactor: reload permission on refresh event

vohoanglong0107:


src/bors/handlers/refresh.rs
```diff
@@ -10,12 +10,21 @@ pub async fn refresh_repository<Client: RepositoryClient>(
     repo: &mut RepositoryState<Client>,
     db: &dyn DbClient,
 ) -> anyhow::Result<()> {
-    let timeout = repo.config.timeout;
+    cancel_timed_out_builds(repo, db).await?;
```
Kobzol, line 13: We should do these two things always, i.e. if cancelling somehow fails, we should still try to also reload permissions.

Kobzol (commented):
Thank you, this is a good start! The hard part is modifying the refresh logic so that handling of events can proceed concurrently with the refresh. That will require a much larger refactoring of the app (should be left for another PR).

PR #55, opened 2024-03-28
feat: reload config

### LINKED ISSUE #49: Reload repository config while the bot is running
The bot loads configuration from a `rust-bors.toml` file, located in the root of the repositories attached to the bot. Currently, it only reloads them when the bot itself is restarted.

It should be able to reload them without restarting itself. Either periodically (e.g. every 30 minutes or so), or it should react to pushes to the main branch on the repository, and reload the config after each push (or if it finds out that the `rust-bors.toml` file was modified).

vohoanglong0107:
# What

Fix #49

### DIFF
--- a/src/bors/handlers/refresh.rs
+++ b/src/bors/handlers/refresh.rs
@@ -12,6 +12,7 @@ pub async fn refresh_repository<Client: RepositoryClient>(
 ) -> anyhow::Result<()> {
     let res = cancel_timed_out_builds(repo, db).await;
     reload_permission(repo).await;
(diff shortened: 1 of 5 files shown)


Kobzol (approved):
Nice and clean :+1: Thank you!

Once the refreshing logic is separated from the main event/command handling loop, we will probably want to assign different refresh periods for the various things that are being refreshed (time will tell).

Also, we should probably explicitly set some reasonable timeouts for all the network communications happening here, but that's for another PR.

PR #56, opened 2024-03-29
feat: use user id instead of username

### LINKED ISSUE #47: Use GitHub IDs instead of usernames
The bot currently uses GitHub usernames instead of GitHub IDs for checking permissions. This is not secure, we should switch it to use IDs instead. Check `src/permissions.rs`.

vohoanglong0107:
# What

Fix #47

### DIFF
--- a/src/permissions.rs
+++ b/src/permissions.rs
@@ -1,4 +1,5 @@
 use axum::async_trait;
+use octocrab::models::UserId;
 use std::collections::HashSet;
 use tokio::sync::Mutex;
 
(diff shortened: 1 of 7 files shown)


src/permissions.rs
```diff
@@ -14,7 +15,7 @@ pub enum PermissionType {
 /// Decides if a GitHub user can perform various actions using the bot.
 #[async_trait]
 pub trait PermissionResolver {
-    async fn has_permission(&self, username: &str, permission: PermissionType) -> bool;
+    async fn has_permission(&self, _username: &UserId, _permission: PermissionType) -> bool;
```
Kobzol, line 18:
```suggestion
    async fn has_permission(&self, _user_id: &UserId, _permission: PermissionType) -> bool;
```
Same in a few other places.

Kobzol (commented):
One small nit, otherwise good. Thank you!

Kobzol (approved):

PR #57, opened 2024-03-30
feat: display workflow url

### LINKED ISSUE #46: Link to cancelled workflow when cancelling a try build
When a try build is cancelled, the bot currently sends a comment that says "Try build cancelled". It would be nice to add a link to this message that would point to the cancelled CI workflow (if there is any).

vohoanglong0107:
# What

Fix #46

### DIFF
--- a/src/bors/handlers/trybuild.rs
+++ b/src/bors/handlers/trybuild.rs
@@ -190,7 +190,7 @@ pub(super) async fn command_try_cancel<Client: RepositoryClient>(
 Cancelled workflows:"#
                 .to_string();
             for id in workflow_ids {
-                let url = repo.client.get_workflow_url(id).await;
+                let url = repo.client.get_workflow_url(id);
                 try_build_cancelled_comment += format!("\n- {}", url).as_str();
             }
             repo.client
@@ -301,7 +301,6 @@ async fn check_try_permissions<Client: RepositoryClient>(
 
 #[cfg(test)]
 mod tests {
-    use anyhow::anyhow;
     use sea_orm::EntityTrait;
 
     use entity::workflow;
--- a/src/bors/handlers/trybuild.rs
+++ b/src/bors/handlers/trybuild.rs
@@ -1,4 +1,5 @@
 use anyhow::{anyhow, Context};
+use octocrab::models::RunId;
 
 use crate::bors::command::Parent;
 use crate::bors::handlers::labels::handle_label_trigger;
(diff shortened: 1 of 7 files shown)


src/bors/handlers/trybuild.rs
```diff
@@ -167,21 +168,29 @@ pub(super) async fn command_try_cancel<Client: RepositoryClient>(
...
             "Could not cancel workflows for SHA {}: {error:?}",
             build.commit_sha
-        );
-    }
-
-    db.update_build_status(&build, BuildStatus::Cancelled)
-        .await?;
+        ),
+        Ok(workflow_ids) => {
+            db.update_build_status(&build, BuildStatus::Cancelled)
```
Kobzol, line 177: We definitely want to mark the build as cancelled, even if the workflow itself was not actually cancelled. That's why the code had `if let Err(error)` previously. Let's keep the original logic, and just store the IDs into an `Option`, and then include them in the comment message if they are `Some`.

Actually, it would be nice to have a test that checks if the build was marked as cancelled in DB even if the workflow cancellation has failed (not sure if there is one already somewhere).

src/github/api/client.rs
```diff
@@ -190,12 +190,16 @@ impl RepositoryClient for GithubRepositoryClient {
...
         let actions = self.client.actions();
 
         // Cancel all workflows in parallel
-        futures::future::join_all(run_ids.into_iter().map(|run_id| {
-            actions.cancel_workflow_run(self.repo_name.owner(), self.repo_name.name(), run_id)
+        futures::future::join_all(run_ids.iter().map(|run_id| {
+            actions.cancel_workflow_run(
+                self.repo_name.owner(),
+                self.repo_name.name(),
+                run_id.clone(),
```
Kobzol, line 201: Clippy warning :)

src/github/api/client.rs
```diff
@@ -247,6 +251,23 @@ impl RepositoryClient for GithubRepositoryClient {
...
+            .map(|url| url.to_string())
+            .unwrap_or_else(|| {
+                format!(
+                    "{}/{}/{}",
+                    base_github_html_url(),
+                    self.name().owner,
+                    self.name().name
+                )
+            });
+        format!("{}/actions/runs/{}", html_url, run_id)
```
Kobzol, line 269:
```suggestion
        format!("{html_url}/actions/runs/{run_id}")
```

src/github/api/mod.rs
```diff
@@ -23,6 +23,12 @@ type GHRepositoryState = RepositoryState<GithubRepositoryClient>;
 
 type RepositoryMap = HashMap<GithubRepoName, GHRepositoryState>;
 
+fn base_github_html_url() -> Url {
+    let url = std::cell::OnceCell::new();
```
Kobzol, line 27: Now that I see this, this actually creates the `OnceCell` everytime the function is called. It would need to be `static`, but then it would need to implement `Sync`, so it needs to be wrapped in a thread local or it would need to be `OnceLock` :laughing:

Yeah, my code below is totally wrong =D

Let's just returns `&'static str` here, we don't need the `Url` for this specific URL, I think.

Kobzol (commented):
Thanks! Left a few comments. You'll also need to rebase.

src/bors/handlers/trybuild.rs
vohoanglong0107: Oh got it, I missed that part. Thanks for pointing out

src/github/api/mod.rs
vohoanglong0107: Ah, that's explain why. I didn't understand the OnceCell part, so I just copy it. Do you want me to do the same for the other function too?

src/bors/handlers/trybuild.rs
```diff
@@ -167,21 +168,36 @@ pub(super) async fn command_try_cancel<Client: RepositoryClient>(
...
+                build.commit_sha
+            );
+            db.update_build_status(&build, BuildStatus::Cancelled)
+                .await?;
 
-    repo.client
-        .post_comment(pr_number, "Try build cancelled.")
-        .await?;
+            repo.client
+                .post_comment(pr_number, "There was an error cancelling try build!")
```
Kobzol, line 181: Let's reword this a bit, otherwise people might retry the cancelling too much. Something like "Try build was cancelled. It was not possible to cancel some workflows."

src/bors/handlers/trybuild.rs
```diff
@@ -167,21 +168,36 @@ pub(super) async fn command_try_cancel<Client: RepositoryClient>(
...
+        Ok(workflow_ids) => {
+            db.update_build_status(&build, BuildStatus::Cancelled)
+                .await?;
+            tracing::info!("Try build cancelled");
+
+            let mut try_build_cancelled_comment = r#"Try build cancelled.
+Cancelled workflows:"#
+                .to_string();
+            for id in workflow_ids {
+                let url = repo.client.get_workflow_url(id).await;
```
Kobzol, line 193: This doesn't need to be an `async` method, let's just keep it sync for simplicity.

Kobzol (commented):
Looks good. There are a few warnings, and left a few nits, otherwise should be good to go.

PR #62, opened 2024-04-20
refactor: dedicated comment Trait

vohoanglong0107:
Preparing for adding html comments to Bors PR comment


Kobzol (commented):
Thank you! It is nice to centralize the comments. I know that I told you to use a trait, but now that I see the code, I wonder if it would be easier to do this with just structures.

I don't think that the comments will have a lot of custom logic (feel free to prove me wrong if you found something like this in homu!). The reason why I wanted to add the Comment abtstraction is to make it more "discoverable" what comments we have, and also to support serializing "homu metadata" in the comments. If we won't need anything else, I think that we could design a struct representing a comment, something like this:
```rust
struct Comment {
   text: String,
   metadata: Option<Box<dyn Serialize> // data type to be modified once we have an actual use-case for metadata
}
```

and this struct would then know how to render itself (generically) into a GitHub comment.

Then we would create constructors for all the various comments that we have:
```rust
fn try_build_started(head_sha: CommitSha, merge_sha: CommitSha) -> Comment {
   Comment {
      text: format!("..."),
      metadata: None
   }
   // or e.g.
   CommentBuilder::text(format!("..."))
      .with_metadata(...)
      .build()
}
```

We wouldn't need to box everything, which is a bit annoying in code, and also the `impl Comment` blocks, which are also quite boilerplate-y for such a simple trait.
What do you think?

vohoanglong0107:
I agree. The way I'm doing now lead to too many type of comments. It also take some effort just to add a new comment

src/bors/mod.rs
```diff
@@ -105,3 +105,17 @@ pub struct RepositoryState<Client: RepositoryClient> {
     pub permissions: ArcSwap<UserPermissions>,
     pub config: ArcSwap<RepositoryConfig>,
 }
+
+/// A comment that can be posted to a pull request.
+pub struct Comment {
+    text: String,
```
vohoanglong0107, line 111: I could't find a use for the metadata, as currently homu just serialize the entire comment to the html tag. Once there is a concrete usecase I will update this

Kobzol (approved):
Thanks, this is a good first step.

PR #64, opened 2024-04-20
Temporarily disable concurrency of bors

Kobzol:
For now, bors does not perform any locking of per-repository GitHub API operations, which share a common resource (e.g. the try branches). This means that it is open to race conditions. Until such locking is introduced, we should handle commands serially.

The code is now prepared for concurrent execution (with `Arc`s etc.), so disabling/enabling concurrency is a matter of such a one-line change, which is great.

### DIFF
--- a/src/github/server.rs
+++ b/src/github/server.rs
@@ -74,16 +74,14 @@ pub fn create_bors_process(
             let state = state.clone();
             let ctx = ctx.clone();
 
-            tokio::spawn(async move {
-                let span = tracing::info_span!("Event");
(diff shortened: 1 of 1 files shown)


Kobzol:
CC @vohoanglong0107

Mark-Simulacrum:
Note that concurrency can also exist outside of bors. Our deployment strategy for ECS tasks spins up a new task and then shuts down the old one, so there's naturally a period of overlap.

Kobzol:
Yeah, this is a bit concerning, we might need some mechanism to deal with overlapping deployments.

PR #68, opened 2024-04-22
feat: add metadata to comment

### LINKED ISSUE #58: Add homu-compatible metadata to PR comments
The [homu](https://github.com/rust-lang/homu) bot adds some special HTML comments to its GH PR comments, which are then parsed by other tools, for example by [`rustc-perf`](https://github.com/rust-lang/rustc-perf/blob/3cd223ad19ca03912070684724728e676dff9164/site/src/github.rs#L370). I would like to eventually change the format, but for now we should keep compatibility with homu, and add these special markes to try build comments. This should be hopefully the last step needed to make try builds on `rust-lang/rust` working with the benchmark suite.

vohoanglong0107:
Fixes #58
We have much more unique comment than we do metadata, so instead of a dedicated Comment Struct for each type of comments, I created a Metadata struct for each type of Metadata

### DIFF
--- a/src/bors/comment.rs
+++ b/src/bors/comment.rs
@@ -1,0 +1,49 @@
+use serde::Serialize;
+
+use crate::github::CommitSha;
+
+/// A comment that can be posted to a pull request.
+pub struct Comment {
+    text: String,
+    metadata: Option<CommentMetadata>,
+}
+
+#[derive(Serialize)]
+#[serde(tag = "type")]
+pub enum CommentMetadata {
+    TryBuildCompleted { merge_sha: String },
+}
+
+impl Comment {
+    pub fn new(text: String) -> Self {
+        Self {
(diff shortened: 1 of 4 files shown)


Kobzol (approved):
I used `gh pr view 112049 --comments --json comments` on a `rust-lang/rust` PR to examine the HTML contents of the comments.
This is what homu puts in:
```
<!-- homu: {\"type\":\"TryBuildCompleted\",\"builders\":{\"checks-actions\":\"https://github.com/rust-lang-ci/rust/actions/runs/8767267903/job/24061841144\"},\"merge_sha\":\"cb5c3f06fa0b92756ca75f4a6bed6833359d1634\"} -->
```
It has the additional `builders` field, but we don't need that for rustc-perf, so this should be fine.

Thanks!

PR #69, opened 2024-04-22
refactor: move is_internal_comment to per repo

vohoanglong0107:
An effort to clean up the blur between global and per repo state

### DIFF
--- a/src/github/api/client.rs
+++ b/src/github/api/client.rs
@@ -1,7 +1,7 @@
 use anyhow::Context;
 use axum::async_trait;
 use base64::Engine;
-use octocrab::models::{Repository, RunId};
+use octocrab::models::{App, Repository, RunId};
 use octocrab::{Error, Octocrab};
 use tracing::log;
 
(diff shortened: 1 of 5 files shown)


src/github/api/client.rs
```diff
@@ -42,6 +43,17 @@ impl RepositoryClient for GithubRepositoryClient {
         self.name()
     }
 
+    /// Was the comment created by the bot?
+    async fn is_comment_internal(&self, comment: &PullRequestComment) -> anyhow::Result<bool> {
+        let app = self
```
Kobzol, line 48: Let's just pass `app.html_url` to the GH repository client when creating it, to avoid an unnecessary network call that could fail, and to remove `Result` from the return type. The URL should be immutable.

Kobzol (changes requested):
Thanks! Left one nit.

src/github/api/client.rs
```diff
@@ -13,6 +14,7 @@ use crate::github::{Branch, CommitSha, GithubRepoName, PullRequest, PullRequestN
 
 /// Provides access to a single app installation (repository) using the GitHub API.
 pub struct GithubRepositoryClient {
+    pub app: App,
```
vohoanglong0107, line 17: I tried to save just app_html_url, but it does not make sense, so I saved the entire App instead

Kobzol: I wouldn't mind caching just the URL, but storing the App is also fine. Thanks!

Kobzol (approved):

PR #70, opened 2024-04-22
refactor: split global event from per repo ones

vohoanglong0107:


src/bors/handlers/mod.rs
```diff
@@ -28,10 +30,26 @@ pub async fn handle_bors_event<Client: RepositoryClient>(
...
+    event: BorsRepositoryEvent,
+    state: Arc<dyn BorsState<Client>>,
+    ctx: Arc<BorsContext>,
 ) -> anyhow::Result<()> {
     let db = Arc::clone(&ctx.db);
     match event {
-        BorsEvent::Comment(comment) => {
+        BorsRepositoryEvent::Comment(comment) => {
             // We want to ignore comments made by this bot
             if let Some(repo) = get_repo_state(state, &comment.repository) {
```
Kobzol, line 54: It would be nice to attach the repository name to each `BorsRepositoryEvent`, so that we can move out `get_repo_state` out of the match

Kobzol (commented):
Heh, it seems like I can just think something and you will do it soon after, that is awesome :laughing: Left one nit.

src/bors/event.rs
```diff
@@ -15,6 +15,17 @@ pub enum BorsRepositoryEvent {
     CheckSuiteCompleted(CheckSuiteCompleted),
 }
 
+impl BorsRepositoryEvent {
+    pub fn repository(&self) -> &GithubRepoName {
```
Kobzol, line 19: Oh, this is even better than attaching it to the enum as a field. Nice!

src/bors/handlers/mod.rs
```diff
@@ -48,10 +48,10 @@ pub async fn handle_bors_repository_event<Client: RepositoryClient>(
     ctx: Arc<BorsContext>,
 ) -> anyhow::Result<()> {
     let db = Arc::clone(&ctx.db);
-    match event {
-        BorsRepositoryEvent::Comment(comment) => {
-            // We want to ignore comments made by this bot
-            if let Some(repo) = get_repo_state(state, &comment.repository) {
+    if let Some(repo) = get_repo_state(state, &event.repository()) {
```
Kobzol, line 51: Clippy warning + let's turn this into `let-else` to avoid the nesting:
```rust
let Some(repo) = get_repo_state(..) else {
   log::error!("Received event {event:?} for an unknown repository {}", event.repository());
   return Ok(());
}
```

Kobzol (approved):
Thanks! :heart:

Issue #66, opened 2024-04-20
Make refreshing more robust


Kobzol:
Currently, if refreshing the data (config/permissions) of a repository fails, it will essentially remove that repository from bors. Instead, we should just continue using the old data for that repository if it fails to update.

Also, the refreshment updates per repository should be granular, i.e. if refreshing the config succeeds, but refreshing permissions fails, we should update the config, but keep the old permissions.

PR #72, opened 2024-04-23
refactor: remove GithubAppState

vohoanglong0107:


src/bors/context.rs
```diff
@@ -1,13 +1,32 @@
...
-impl BorsContext {
-    pub fn new(parser: CommandParser, db: Arc<dyn DbClient>) -> Self {
-        Self { parser, db }
+impl<Client: RepositoryClient> BorsContext<Client> {
+    pub async fn new(
+        parser: CommandParser,
+        db: Arc<dyn DbClient>,
+        global_client: Arc<dyn GlobalClient<Client>>,
+    ) -> Self {
+        // this unwrap is making me nervous, but if lhe repos loading
```
Kobzol, line 21: We can just return `anyhow::Result<Self>` from the constructor, no need to `unwrap` here.

src/bin/bors.rs
```diff
@@ -72,14 +72,14 @@ fn try_main(opts: Opts) -> anyhow::Result<()> {
...
 
-    let state = runtime
-        .block_on(GithubAppState::load(
-            opts.app_id.into(),
-            opts.private_key.into_bytes().into(),
-        ))
-        .context("Cannot load GitHub repository state")?;
-    let ctx = BorsContext::new(CommandParser::new(opts.cmd_prefix), Arc::new(db));
-    let (repository_tx, global_tx, bors_process) = create_bors_process(state, ctx);
+    let client = create_github_client(opts.app_id.into(), opts.private_key.into_bytes().into())?;
```
Kobzol, line 75: This crashes at runtime, it needs to be executed within a tokio Runtime. Let's move it into an `async move` block in the `block_on` call below.

src/bors/context.rs
```diff
@@ -1,13 +1,32 @@
...
+use crate::{bors::command::CommandParser, database::DbClient, github::GithubRepoName};
+use std::{collections::HashMap, sync::Arc};
+
+use super::{GlobalClient, RepositoryClient, RepositoryState};
+
+pub struct BorsContext<Client: RepositoryClient> {
     pub parser: CommandParser,
     pub db: Arc<dyn DbClient>,
+    pub global_client: Arc<dyn GlobalClient<Client>>,
+    pub repositories: Arc<ArcSwap<HashMap<GithubRepoName, Arc<RepositoryState<Client>>>>>,
```
Kobzol, line 12: - Let's use `GithubRepositoryMap` (and move it into this file).
- The outer `Arc` looks unnecessary (the main program compiles without it).

But I think that here I would do a step back and think about how exactly we want to read from, and write to, the repository state.
What we want to do:
- Mutably add (even possibly remove) new repositories in refresh, if someone adds a new repo to the bot while it is running (I think that we should support this use-case, even though it is quite rare). It's probably better to just modify the map than to replace it altogether, so I would use a `RwLock` around the `HashMap`.
- We need to access a given repository state in per-repo event handling, and also in global event handling, but for that it should be enough that we have `Arc<BorsContext>`. We don't ever need to change the name of a repo, nor its client. Therefore name and client in repository state could be read-only. We do need to atomically replace the config and the permissions, so these two fields inside the state should be behind `ArcSwap`, and they already are, so that's good.

Apart from that, I don't see other use-cases for accessing the state. So I *think* that we can simplify this type just to `RwLock<HashMap<GithubRepoName, RepositoryState<Client>>>`, and that should work. Could you please try it?

src/bors/mod.rs
```diff
@@ -70,6 +71,16 @@ pub trait RepositoryClient: Send + Sync {
     fn get_workflow_url(&self, run_id: RunId) -> String;
 }
 
+/// Temporary trait to sastify the test mocking.
+/// TODO: Remove this trait once we move to mock REST API call.
+#[async_trait]
+pub trait GlobalClient<Client: RepositoryClient>: Send + Sync {
```
Kobzol, line 77: This is a bit ugly, as I'm sure you also see :D Apart from getting rid of mocking, what about removing the global client from `BorsContext` and only storing it in `consume_global_events` (or someone who calls this function)?

But we don't need to do that, it would require some additional mocking anyway. I would maybe just rename this to `RepositoryLoader`.

Kobzol (commented):
Thanks! I think that we're moving to a state (pun intended) where it's becoming much clearer how are the individual shared bits of the global state being used. It should also help with https://github.com/rust-lang/bors/issues/66, which would be a good follow-up step I think.

Left some comments. Could you also please update the `docs` documents after this change?

src/github/api/operations.rs
```diff
@@ -42,19 +41,7 @@ pub async fn merge_branches(
...
-        .map(|url| url.to_string())
-        .unwrap_or_else(|| {
-            format!(
-                "{}/repos/{}/{}/merges",
-                base_github_url(),
-                repo.name().owner,
-                repo.name().name
-            )
-        });
+    let merge_url = format!("/repos/{}/{}/merges", repo.name().owner, repo.name().name);
```
Kobzol, line 44: This (the fact that `octocrab` stopped supporting absolute URLs) is exactly the kind of error that can only be found with an integration test.

vohoanglong0107: Exactly.
Sorry, meant to split this into another PR for easier review, but I accidentally pushed it here

src/bors/mod.rs
vohoanglong0107: Yeah we need to mock this to test the installation change event.

src/bors/context.rs
vohoanglong0107: I see. I was forced to mindlessly spam `Arc<RepositoryState<Client>>` by the trait method `get_repo_state`, cause there was no way to get a reference into a repo inside the hashmap. Now that the trait is gone, I should have remove all those unnecessary Arc.

About RwLock, do you think using RwLock is better than ArcSwap, at least for the hashmap?

Kobzol: It's a bit more annoying because of the unwrap, but I think it makes more sense herr. We don't need to replace the whole hashmap during refresh. In fact, we don't want to, because if e.g. the refresh of a single repo fails, we should still keep its old data, and that would be annoying if we'd always need to rebuild the whole map. Also we don't want to replace the repo clients with cached GH tokens.

Btw, while thinking about this more, I realized that it might be more practical to put the repo state into an Arc after all, just to avoid holding on to the hashmap RwLock for the whole duration of a repo event handling (to avoid repo events from starving global events and vice-versa). But that should be an easy change that we can do later, so let's start with the easier approach without the Arc.

vohoanglong0107: I see. Since it's only just a matter of reverting a commit, I have added the Arc back to the repo state

src/bors/context.rs
```diff
@@ -1,13 +1,33 @@
...
-impl BorsContext {
-    pub fn new(parser: CommandParser, db: Arc<dyn DbClient>) -> Self {
-        Self { parser, db }
+impl<Client: RepositoryClient> BorsContext<Client> {
+    pub async fn new(
+        parser: CommandParser,
+        db: Arc<dyn DbClient>,
+        global_client: Arc<dyn RepositoryLoader<Client>>,
+    ) -> anyhow::Result<Self> {
+        // this unwrap is making me nervous, but if lhe repos loading
```
Kobzol, line 22: No unwrap here anymore :)

Kobzol (approved):
One nit, otherwise LGTM. https://github.com/rust-lang/bors/issues/66 is a logical next step, now that the global state has been made more reasonable.

src/bors/context.rs
vohoanglong0107: My bad 😅

PR #75, opened 2024-04-25
fix: load app with app client

vohoanglong0107:
Cause installation client couldn't be used to load app info


Kobzol (approved):
Thanks, this makes sense.

PR #77, opened 2024-04-25
fix: decode file content with octocrab builtin

vohoanglong0107:


Kobzol:
Nice find! Could you also please remove `base64` from `Cargo.toml`?

Kobzol (approved):
Thanks! :)

PR #78, opened 2024-04-25
fix: merge and try build url

vohoanglong0107:


Kobzol (commented):
There are two more usages of `base_github_url`. Since the `octocrab` update, we should probably update all the occurrences of the absolute URLs and replace them with relative ones?

Kobzol (approved):
Thanks :heart:

PR #79, opened 2024-04-26
docs: add necesarry webhook event to test bors

vohoanglong0107:
Bors also need to listen to `Workflow run` event


Kobzol (approved):
Good catch, thanks! :)

PR #83, opened 2024-05-04
feat: make refresh more robust

### LINKED ISSUE #66: Make refreshing more robust

vohoanglong0107:
Fix #66 
By comparing the set of repositories before reloading and the set of repositories got from reloading, I could infer which repository was reloaded, which was not and which was recently added.
I tried to add tests to verify this behavior but was limited by our current tests infra. It was hard to load more than one repo during reloading due to the way we are mocking the `RepositoryLoader`.
The refresh process of each repository (config refreshing, etc...) is already granular, and each refresh task run concurrently with the other using `tokio::join!`. Therefore, no further action is needed in this regard.

### DIFF
--- a/src/bors/handlers/mod.rs
+++ b/src/bors/handlers/mod.rs
@@ -125,10 +125,23 @@ pub async fn handle_bors_global_event<Client: RepositoryClient>(
             let span = tracing::info_span!("Repository reload").entered();
 
             match ctx.repository_loader.load_repositories().await {
-                Ok(repos) => {
+                Ok(reloaded_repos) => {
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Thanks!

PR #85, opened 2024-05-06
feat: arbitrary try build

### LINKED ISSUE #80: Implement support for arbitrary try builds
With https://github.com/rust-lang/rust/pull/124631, the `rust` CI will be able to run an arbitrary CI job in a try build. It would be great to add support for this in bors.

Arbitrary try build CI jobs are parsed from a special line with the format `try-job: <job-name>`, which should be in the commit message of the latest commit on the PR. Currently, bors sets the Auto merge commit message to the PR description, so users have to modify it. With new bors, it should instead append these special lines directly to the merge commit message.

Steps:
1) Extend the parsing of a `@bors try` command so that it will be able to receive the names of arbitrary CI jobs.
Something like `@bors try jobs=<name1>,<name2>,...`
2) For each job name in the command, output a single line with the contents `try-job: <job-name>` to the commit message of the try build

And that's pretty much it. We could also implement some sanity checks, e.g. if the user specifies 50 jobs, bors should error out (https://github.com/rust-lang/rust/pull/124631 currently allows specifying 10 jobs max.).

vohoanglong0107:
Fixes #80 
It works on my branch 🤣
https://github.com/vohoanglong0107/test-bors/commit/9eeaa0c106c10810b5e7bf77540a1a4ab26e6e84

### DIFF
--- a/src/bors/command/mod.rs
+++ b/src/bors/command/mod.rs
@@ -23,7 +23,7 @@ pub enum BorsCommand {
         /// Parent commit which should be used as the merge base.
         parent: Option<Parent>,
         /// The CI workflow to run.
-        jobs: Option<Vec<String>>,
+        jobs: Vec<String>,
     },
     /// Cancel a try build.
     TryCancel,
--- a/src/bors/command/mod.rs
+++ b/src/bors/command/mod.rs
@@ -22,6 +22,8 @@ pub enum BorsCommand {
     Try {
         /// Parent commit which should be used as the merge base.
         parent: Option<Parent>,
+        /// The CI workflow to run.
+        jobs: Option<Vec<String>>,
     },
     /// Cancel a try build.
     TryCancel,
--- a/src/bors/command/parser.rs
+++ b/src/bors/command/parser.rs
@@ -140,7 +140,7 @@ fn parser_try<'a>(command: &'a str, parts: &[CommandPart<'a>]) -> ParseResult<'a
     }
 
     let mut parent = None;
-    let mut jobs = None;
+    let mut jobs = Vec::new();
 
     for part in parts {
         match part {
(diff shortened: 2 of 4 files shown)


src/bors/command/mod.rs
```diff
@@ -22,6 +22,8 @@ pub enum BorsCommand {
     Try {
         /// Parent commit which should be used as the merge base.
         parent: Option<Parent>,
+        /// The CI workflow to run.
+        jobs: Option<Vec<String>>,
```
Kobzol, line 26:
```suggestion
        jobs: Vec<String>,
```
This should be easier. Passing an empty `jobs=` list should be an error.

src/bors/command/parser.rs
```diff
@@ -262,14 +280,26 @@ line two
...
+            })
+        ));
     }
 
     #[test]
     fn parse_try() {
         let cmds = parse_commands("@bors try");
         assert_eq!(cmds.len(), 1);
-        assert!(matches!(cmds[0], Ok(BorsCommand::Try { parent: None })));
+        assert!(matches!(
```
Kobzol, line 296: Rewriting these tests to `insta` would make it to bless them :laughing: (we don't have to do that in this PR).

src/bors/handlers/trybuild.rs
```diff
@@ -232,17 +233,27 @@ fn get_pending_build(pr: PullRequestModel) -> Option<BuildModel> {
...
 
 {pr_message}"#,
         pr_label = pr.head_label,
         pr_title = pr.title,
         pr_message = pr.message
-    )
+    );
+    if let Some(jobs) = jobs {
+        for job in jobs {
+            message.push_str(&format!("\nci-job: {}", job));
```
Kobzol, line 253:
```suggestion
            message.push_str(&format!("\ntry-job: {}", job));
```
We actually renamed the tag since I created the bors issue :sweat_smile: I modified it in the issue.

Kobzol (commented):
Looks great, thank you! Left a few nits.

Kobzol (approved):
Thanks! :)

PR #86, opened 2024-05-08
tests: ping comment with wiremock

vohoanglong0107:
This took so longgg!

This PR introduces BorsTester struct, which coordinates the set up of all the dependencies for Bors and mocks all outgoing Github requests with wiremock.

The old test infrastructure still remains, as I intended to gradually replace the tests for other handlers in subsequent PRs.

### DIFF
--- a/src/tests/mocks/issue.rs
+++ b/src/tests/mocks/issue.rs
@@ -1,0 +1,290 @@
+use std::{
+    collections::HashMap,
+    sync::{Arc, Mutex},
+};
+
+use serde::{Deserialize, Serialize};
+use url::Url;
+use wiremock::{
+    matchers::{method, path},
+    Mock, MockServer, Request, ResponseTemplate,
+};
+
+use super::{
+    repository::Repository,
+    user::{default_user, Author},
+};
+
+#[derive(Serialize)]
+struct Issue {
+    id: u64,
+    node_id: String,
+    user: Author,
+    url: Url,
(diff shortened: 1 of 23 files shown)


src/tests/mocks/issue.rs
```diff
@@ -0,0 +1,290 @@
...
+    1
+}
+
+pub(crate) struct InMemoryIssue {
+    comments: Arc<Mutex<HashMap<u64, Vec<String>>>>,
+}
+
+impl InMemoryIssue {
+    pub(crate) async fn check_comments(&self, expected_comments: &[&str]) -> bool {
+        let pr_number = default_pr_number();
```
Kobzol, line 242: This is quite hacky and could become flaky. What about creating a queue of comments and waiting until a comment is sent to it by a mocked impl?

Kobzol (commented):
I left a few comments, but it's quite difficult for me to make sense of this :sweat_smile: We should understand properly how are we gonna use the end-to-end tests, so that it can drive the implementation.

From a high-level point of view, we should describe the state existing on GitHub (including things like the contents of `rust-bors.toml` in a repo, the installations of the app etc.), and then derive the mocked responses from it.

Let's start with something smaller, like just initiating the state of the app, which should load config from a set of installed repos.

vohoanglong0107:
From our discussion, I will open smaller PRs to implement this.

PR #87, opened 2024-05-20
refactor: migrate from sea-orm to sqlx

vohoanglong0107:
This is the first step to migrate from `sea-orm` to `sqlx`. All sea-orm models have been removed. All interaction with the postgres database through raw sql string is defined in module `database/operations`.
The tests are expected to fail in this PR, because all those test share the same database connection. A new PR will be raised later to fix those tests by using `sqlx::test`, allowing each test to have its own database.


migrations/20240517094752_create_build.down.sql
```diff
@@ -0,0 +1,7 @@
+BEGIN;
```
Kobzol, line 1: I would expect that `sqlx` runs the migrations in a transaction by default?

src/database/mod.rs
```diff
@@ -32,22 +64,58 @@ pub enum BuildStatus {
     Timeouted,
 }
 
+impl sqlx::Type<Postgres> for BuildStatus {
```
Kobzol, line 67: Cool, I had no idea that `sqlx` has these conversions.

src/database/mod.rs
```diff
@@ -61,6 +129,34 @@ pub enum WorkflowType {
     External,
 }
 
+impl sqlx::Type<Postgres> for WorkflowType {
+    fn type_info() -> PgTypeInfo {
+        <String as sqlx::Type<Postgres>>::type_info()
+    }
+}
+
+impl sqlx::Decode<'_, Postgres> for WorkflowType {
```
Kobzol, line 138: Maybe this could be done even easier with the `sqlx` proc macro https://docs.rs/sqlx/latest/sqlx/trait.Type.html#enumeration?

src/github/mod.rs
```diff
@@ -44,6 +44,34 @@ impl Display for GithubRepoName {
...
+impl From<String> for GithubRepoName {
+    fn from(value: String) -> Self {
+        let mut parts = value.split('/');
+        let owner = parts.next().unwrap_or_default();
+        let name = parts.next().unwrap_or_default();
+        Self::new(owner, name)
+    }
+}
+
+impl sqlx::Type<sqlx::Postgres> for GithubRepoName {
```
Kobzol, line 56: Could we move these impls to the `DB` module?

src/github/mod.rs
```diff
@@ -96,6 +124,20 @@ impl From<u64> for PullRequestNumber {
     }
 }
 
+// to load from/ save to Postgres, as it doesn't have unsigned integer types.
+impl From<i64> for PullRequestNumber {
```
Kobzol, line 128: Same here.

Kobzol (commented):
Looks very good! Had a quick look and left a few comments.

migrations/20240517094752_create_build.down.sql
vohoanglong0107: Ah, I didn't know that. Apparently it does, but only for Postgres.

src/database/mod.rs
vohoanglong0107: I wish it could, but sqlx's support for postgres enum is quite limited. I had quite a difficult time attempting to query the BuildStatus enum within the BuildModel inside Workflow, only to be greeted by this issue: https://github.com/launchbadge/sqlx/issues/1920
> error occurred while decoding column variables: custom types in records are not fully supported yet: failed to retrieve type info for field 4 with type oid 25101253

So I just manually implemented `Decode` trait for each of those enums.

Kobzol: I don't really want to use Postgres enums, just map Rust enums to strings in the DB. But I suppose that's not easily possible with sqlx (?).

Kobzol:
LGTM, but we can't merge with tests failing. So either you need to comment out the tests in CI for now, or fix tests in this PR.

Btw, the Dockerfile needs updating, as it copies a directory that doesn't exist anymore.

src/database/mod.rs
vohoanglong0107: Yeah, I've just checked sqlx source code to be certain, and it appears under the hood `sqlx` use the database's native enum type to represent Rust enum.

Kobzol: I checked the [docs](https://docs.rs/sqlx/latest/sqlx/trait.Type.html#enumeration) again and they claim that without `#[repr(i32)]` or `#[sqlx(type_name)]`, it should just be using text. It's not obvious from the documentation, but this works fine:
```rust
#[derive(sqlx::Type, Debug)]
#[sqlx(type_name = "TEXT")] // <- this is the important bit
#[sqlx(rename_all = "lowercase")]
enum Color {
    Red,
    Green,
    Blue
}

#[derive(sqlx::FromRow, Debug)]
struct Row {
    color: Color
}

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    let pool = PgPool::connect("postgres://postgres:postgres@localhost/postgres").await?;
    sqlx::query("CREATE TABLE IF NOT EXISTS test(COLOR TEXT);").execute(&pool).await?;

    // DB contains { color: "red" }
    // sqlx::query("INSERT INTO test(COLOR) VALUES($1)").bind(Color::Red).execute(&pool).await?;
    let row: Row = sqlx::query_as("SELECT * FROM test").fetch_one(&pool).await?;
    println!("{row:?}");

    Ok(())
}
```
This would be ideal, however I haven't found a way how to get this working with `query!`, i.e. compile-time checked queries.

vohoanglong0107: Sorry, I wasn't clear earlier. What I meant was there was no way to use the enum within the query macro. I did not try the function version though

Kobzol: It does work for inserts:
```rust
sqlx::query!("INSERT INTO test(color) VALUES($1)", Color::Red as Color).execute(&pool).await?;
```
Not sure how to query it back from the DB yet, though.

Kobzol: Ok, this does the trick:
```rust
let row: Row = sqlx::query_as!(Row, r#"SELECT color AS "color!: Color" FROM test"#).fetch_one(&pool).await?;
```
it's a bit ugly, but at the same time it saves us from duplicating the encode/decode impls and the string constants within them.

Kobzol:
To avoid having to commit the `.sqlx` directory created by `cargo sqlx prepare`, I think that the easier thing would be to spin up a Postgres instance in our CI tests and export a `DATABASE_URL` variable pointing to them.

It can be done with the `services` key: https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers#running-jobs-in-containers

src/database/mod.rs
vohoanglong0107: Wow, `type_name=TEXT` really cooked. Not once had I thought that I could put TEXT as type name, I kept trying to create a new Postgres enum instead 🤣 Thanks Jakub, now I know what to do the next time I encounter this issue.

vohoanglong0107:
Got it. Once tried the approach of committing the `.sqlx` directory, I soon regreted it though.

Dockerfile
```diff
@@ -12,15 +13,19 @@ RUN cargo chef prepare --recipe-path recipe.json
...
 FROM base as build
 
+# require a running postgres instance on the host
+ENV DATABASE_URL=postgres://bors:bors@host.docker.internal/bors
+
 WORKDIR /app
 COPY --from=planner /app/recipe.json recipe.json
 
 RUN cargo chef cook --release --recipe-path recipe.json
+RUN cargo sqlx migrate run
```
Kobzol, line 23: At this point the `migrations` directory hasn't been copied to the image yet. I with we could just disable the compile-time check in the Dockerfile (is it possible?). It is annoying that a DB is needed for the Docker image build. We upload the built image to Amazon ECS, so we would also need to make Postgres available [here](https://github.com/rust-lang/bors/blob/main/.github/workflows/deploy.yml#L32).

Do you know if there is a way to completely disable the compile-time checks?

Kobzol: We will need to run the migrations when the image is executed (so that they run against the production DB), not during the build of the container itself.

Kobzol: Yeah, it should probably be an argument.

vohoanglong0107: I don't know how to connect to localhost during docker build 🥹

Kobzol: I guess that there are ways around that (https://www.reddit.com/r/rust/comments/18y4ts3/compile_time_issue_with_rustpostgres_docker/ maybe?), but I wonder if passing the `.sqlx` to it might not be an easier solution after all. We don't need to commit it, we could just prepare it on CI and then pass it to the Dockerfile.

What issues with `cargo sqlx prepare` did you encounter?

vohoanglong0107: Now that I think about it, most of the time, it was due to the the combination of mismatched versions between our dev machines and CI, along with cargo `-D warnings` flag, making each `cargo sqlx prepare` as miserable as it could be.

vohoanglong0107:
Since the tests fix is straightforward (by changing `#[tokio::test]` to `#[sqlx::test]` and handle the pool, I have also included the fix to this PR

.github/workflows/test.yml
```diff
@@ -12,6 +12,23 @@ jobs:
   test:
     name: Test
     runs-on: ubuntu-latest
+    env:
+      SQLX_OFFLINE: true
+      DATABASE_URL: postgres://postgres:password@localhost:5432/test
```
Kobzol, line 17: We shouldn't use `SQLX_OFFLINE` here if we also set up the database, right? Instead, we should add `cargo sqlx prepare --check` somewhere as a test step, to check that the `.sqlx` directory is up to date. And the DB will need to be migrated with `cargo sqlx migrate run` first.

Kobzol (commented):
Could you please also update `docs/development.md`? We should document the switch to `sqlx` and perhaps also prepare some guide for developers how to setup Postgres, i.e. with a simple `docker-compose.yml` file, and also the `.env` file so that compilation even works. Thanks!

.github/workflows/test.yml
vohoanglong0107: Are I see. I was under the impression that `.sqlx` is gonna be used as the source of trust. Will remove it

Kobzol: The main point is to check on CI whether `.sqlx` matches the queries that we have in code, but for that we need a live DB.

vohoanglong0107:
Do you think we should add a `docker-compose.yaml` file at the root of the project? That way we just need to add instructions on running the development environment instead of having to include the `docker-compose.yaml` content within the docs

docs/development.md
```diff
@@ -71,28 +69,35 @@ Everytime you want to run bors:
 - Try `@bors ping` on some PR on the test repository :)
 
 ## Database
-You must have `sea-orm-cli` installed for the following commands to work.
+You must have `sqlx-cli` installed for the following commands to work.
 ```console
-$ cargo install sea-orm-cli
+$ cargo install sqlx-cli --no-default-features --features native-tls,postgres
```
Kobzol, line 74: Please pin the `sqlx-cli` version in this command (`cargo install --version X.Y.Z`).

docs/development.md
```diff
@@ -71,28 +69,35 @@ Everytime you want to run bors:
...
 
-You must also set up a `DATABASE_URL` environment variable. **You can use SQLite for local testing,
-but when entities are regenerated, it should be done against a Postgre database!**
+The database can be set up with the docker-compose file in the root of the repository:
+
+```console
+$ docker-compose up -d
+```
+
+Then, set the `DATABASE_URL` environment variable to the connection string of the database. The content of the variable is can be found in the `.env.example` file.
```
Kobzol, line 83: Let's also tell the user that if `.env` is present, it will  be automatically picked up by `sqlx`.

docs/development.md
```diff
@@ -71,28 +69,35 @@ Everytime you want to run bors:
...
-    ```
-4) Re-generate entities, again against a **Postgre** DB.
-    ```console
-    $ sea-orm-cli generate entity -o database/entity/src --lib
+    $ sqlx migrate run
     ```
+
+### Generate `.sqlx` directory
+```console
+$ sqlx prepare -- --tests
```
Kobzol, line 102: Let's also mention that `.sqlx` needs to be committed after any changes.

Kobzol (commented):
Good! I think that the only change left is to add a CI step that checks on CI that `.sqlx` is updated.

Kobzol (approved):
Awesome! This looks really great, thanks a lot for the hard work! I think that now the DB side is much cleaner, and makes it easier for us to understand exactly what queries we are running against the DB.

Kobzol:
Looks like some GHA bug (https://www.githubstatus.com/). The workflows have finished, but the status hasn't been updated yet.

PR #90, opened 2024-05-21
build: install fixed sqlx-cli version with Just

vohoanglong0107:
Add `justfile` including a recipe to install development tools, allowing all environments to share the same sqlx-cli version.
Also include a recipe to prepare `.sqlx` directory


Kobzol:
I don't think we have the dev environment complex enough to add more similar tooling, so I'm gonna close this for now.

PR #91, opened 2024-05-21
Simplify docker-compose.yml

Kobzol:
I don't think that we need health checks, and the volume was not working for me locally. Explicit volume should not be needed, `docker-compose up` should remember the latest state, unless you clear docker temporary artifacts.

CC @vohoanglong0107

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -2,16 +2,9 @@ version: '3.7'
 services:
   db:
     image: postgres:16
-    volumes:
-      - db_data:/var/lib/postgresql/data
     environment:
       POSTGRES_USER: bors
       POSTGRES_PASSWORD: bors
       POSTGRES_DB: bors
     ports:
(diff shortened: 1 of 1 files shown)


vohoanglong0107:
Ah, my bad, I forgot to add the explicit volume add the end of `docker-compose.yaml`.

> docker-compose up should remember the latest state, unless you clear docker temporary artifacts.

Hmmm, I'm not sure about this. In my experience, without an explicit volume, once we run `docker-compose down`, the data stored inside our postgres instance will be cleared.

Kobzol:
Yeah, but if you don't run `docker-compose down`, and just kill the Docker (as I usually do, with Ctrl+C), zhr data stays there.

vohoanglong0107:
I see. I personally don't have any objection with this, as I'm running a local postgres instance, so I will leave the decision up to you.

PR #92, opened 2024-05-22
fix: span hold accross await point

vohoanglong0107:
This span turns the handler into non Send future, which disables the ability to run handlers concurrently.

### DIFF
--- a/src/bors/handlers/mod.rs
+++ b/src/bors/handlers/mod.rs
@@ -122,23 +122,8 @@ pub async fn handle_bors_global_event<Client: RepositoryClient>(
     let db = Arc::clone(&ctx.db);
     match event {
         BorsGlobalEvent::InstallationsChanged => {
-            let reloaded_repos = ctx.repository_loader.load_repositories().await?;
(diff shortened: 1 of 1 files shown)


src/bors/handlers/mod.rs
```diff
@@ -122,32 +122,23 @@ pub async fn handle_bors_global_event<Client: RepositoryClient>(
...
-                        if repositories
-                            .insert(repo.repository.clone(), repo.clone())
-                            .is_some()
-                        {
-                            tracing::info!("Repository {} was reloaded", repo.repository);
-                        } else {
-                            tracing::info!("Repository {} was added", repo.repository);
-                        }
-                    }
+            let reloaded_repos = ctx.repository_loader.load_repositories().await?;
```
Kobzol, line 125: We can wrap this whole thing in `async {}` (or even better, create an `async fn`), and then instrument it with a span, so that we keep the span, it would be nice to see that this is happening in logs.

Kobzol (changes requested):

Kobzol (approved):
Thanks! :)

PR #97, opened 2024-05-26
tests: mock team api

vohoanglong0107:
Follow up on https://github.com/rust-lang/bors/pull/94, this PR adds a mock server for team API

### DIFF
--- a/src/bors/mod.rs
+++ b/src/bors/mod.rs
@@ -15,6 +15,7 @@ use crate::config::RepositoryConfig;
 use crate::database::RunId;
 use crate::github::{CommitSha, GithubRepoName, MergeError, PullRequest, PullRequestNumber};
 use crate::permissions::UserPermissions;
+use crate::TeamApiClient;
 pub use command::CommandParser;
 pub use comment::Comment;
 pub use context::BorsContext;
@@ -78,6 +79,7 @@ pub trait RepositoryLoader<Client: RepositoryClient>: Send + Sync {
     /// Load state of repositories.
     async fn load_repositories(
         &self,
+        team_api_client: &TeamApiClient,
     ) -> anyhow::Result<HashMap<GithubRepoName, Arc<RepositoryState<Client>>>>;
 }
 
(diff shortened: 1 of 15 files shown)


src/bors/mod.rs
```diff
@@ -101,6 +102,6 @@ pub struct CheckSuite {
 pub struct RepositoryState<Client: RepositoryClient> {
     pub repository: GithubRepoName,
     pub client: Client,
-    pub permissions: ArcSwap<UserPermissions>,
+    pub team_api_client: Arc<TeamApiClient>,
```
Kobzol, line 105: Any specific reason for why to do this? I liked the design where the state only contained actual state (apart from the GH client, which can't easily be replaced with pure state), rather than dependencies from which it is loading stuff.

vohoanglong0107: To provide each handler with access to the team API, I could add the client as an argument to the handler itself, or add it to the RepositoryState. The handlers have one to many parameters now, and it feels like the client and permission share the same purpose in RepositoryState, so I just replace one with another. Do you have any suggestions on how to better handle this?

Kobzol: Why should handlers have access to the team API? :thinking: Shouldn't it be enough for them to just read the existing permissions stored in the repo? The only part of the code that needs access to the team API is the periodic refresh process (which could be made optional in the future, e.g. if we add a way to run with local permissions for easier testing on your own repo).

vohoanglong0107: Yeah, only the refresh handler needs access to the team API to update the permissions. In that case, would it be better to move the client to the refresh handler and leave the state unchanged?

Kobzol: Yeah, I would prefer that :)

vohoanglong0107:
Looking back, I really went overboard with the refactoring. Should have stopped at the base_url, instead I tried to encapsulate permissions into TeamApiClient 😅

Kobzol (approved):
Thanks! I feel like there's a lot of code in the end-to-end test suite already, so it would be nice to validate it with some tests. I'll try to play around with the code and write a few tests and send a PR for you to review, if you don't mind.

PR #100, opened 2024-05-27
Add end-to-end test for loading repositories

Kobzol:
I wanted to draft an API for building the simulated GitHub and team API state that we want to mock in end-to-end tests. I like to create small test-only APIs for constructing the expected state, and then let the testing machinery construct the full data (in this case, the response from the mocked GH API) automatically. I think that it's easier than using the GH structs in tests directly.

I didn't have a lot of time to clean this up, but I think that this can be used as a base from which we can expand the end-to-end testing API. @vohoanglong0107 let me know what do you think.

Note that this first test (`test_load_installed_repos`) isn't actually end-to-end, because we don't spawn the web server, however for testing the loading of the initial state, it's a bit awkward to do it fully end-to-end, because once we spawn the server, we can only communicate with it using webhooks, therefore it's not trivial to figure out what is its inner loaded state.


Cargo.toml
```diff
@@ -58,6 +58,7 @@ insta = "1.26"
 derive_builder = "0.20.0"
 wiremock = "0.6.0"
 base64 = "0.22.1"
+tracing-test = "0.2.4"
```
vohoanglong0107, line 61: Between this crate and `test_log` crate, which one do you prefer? I think test_log is slightly more popular, but I don't have a strong preference.

vohoanglong0107 (commented):
This overall looks very nice, particularly the Repo and Permission builder.
Personally, I really like moving the repo loader out of Context. It simplifies the context and provides a clear distinction between global event handlers and the per-repo handlers.
Regarding the test for repo loading, I agree that it's difficult to use e2e test for it. Perhaps we could improve the test once we have a concrete use case for it (like listing all supported repositories on bors website), but it's fine for now.

Cargo.toml
Kobzol: I just forgot the name of the crate that you were using xD And found something online. What I like a bit more about tracing-test is that it is an "official" `tracing-xxx` crate, and also that it is composable. Instead of modifying the existing `tokio/sqlx::test` attribute, you just slap an additional attribute on top of the function.

Kobzol:
I have a bit more WIP code on top of this, I'll post a PR soon.

PR #107, opened 2024-05-31
Implement comment checking in tests

Kobzol:
CC @vohoanglong0107

### DIFF
--- a/src/tests/mocks/github.rs
+++ b/src/tests/mocks/github.rs
@@ -1,21 +1,51 @@
 use octocrab::Octocrab;
+use std::collections::HashMap;
 use wiremock::MockServer;
 
 use crate::create_github_client;
+use crate::github::GithubRepoName;
 use crate::tests::mocks::app::{default_app_id, AppHandler};
-use crate::tests::mocks::repository::RepositoriesHandler;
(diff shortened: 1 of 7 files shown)


src/tests/mocks/github.rs
```diff
@@ -1,21 +1,83 @@
...
+    async fn get_comment(&mut self, pr: u64) -> Comment {
+        // First, try to resolve the comment from the pending comment list
+        if let Some(index) = self.pending_comments.iter().position(|c| c.pr == pr) {
+            return self.pending_comments.remove(index);
+        }
+        // If it is not there, wait until some comment is received
+        loop {
+            let comment = self
+                .comments_queue
+                .recv()
```
vohoanglong0107, line 37: Just to confirm, if the sender panic this loop will be broke out, right?

Kobzol: Good question. If all senders (well, there should be only one) panic, they should be unwound, therefore their Drop impls will run, which should drop the sender, and that should result in this `recv()` call returning `None`. If something went bad, the call of this method is wrapped in a `timeout`, so hopefully it shouldn't last more than a few seconds until we saw that in the test that something is wrong.

Issue #93, opened 2024-05-22
Test by mocking http request


vohoanglong0107:
Since this is a significant change, I think it's worth outlining my plan here.

To move from mocking dependencies with trait to using a http request mocking framework like wiremock, the http mock must support mocking both the Github API and Team API for resolving permissions.

For Github API, the mock must include the following endpoints:
- [x] GET `/app` -- checking if a comment belong to bors or a normal github user.
- [x] GET `/app/installations` and GET `/installation/repositories` -- used for listing all repositories managed by bors.
- [x] POST `/app/installations/{...}/access_tokens` -- used by `octocrab` to perform authentication, requires a dummy private key.
- [x] GET `/repos/{owner}/{name}/pulls/{number}` -- used to get informations for a pull requests, like head SHA.
- [x] POST `/repos/{owner}/{name}/issues/{number}/comments` -- used to post a new comment.
- [x] POST `/repos/{owner}/{name}/issues/{number}/labels` -- used to attach a label to an issue.
- [x] DELETE `/repos/{owner}/{name}/issues/{number}/labels/{name}` -- used to delete a label from an issue.
- [x] GET `/repos/{owner}/{name}/branches/{name}` -- used to get informations for a branch, like head SHA.
- [x] POST `/repos/{owner}/{name}/git/refs` -- used to create a new branch.
- [x] PATCH `/repos/{owner}/{name}/git/refs/heads/{branch_name}` -- used to update head of a branch.
- [x] POST `/repos/{owner}/{name}/merges` -- used to merge branches.
- [x] GET `/repos/{owner}/{name}/commits/{sha}/check-suites` -- used to get check suites for a commit.
- [x] POST `/repos/{owner}/{name}/actions/runs/{run_id}/cancel` -- used to cancel a github action workflow.
- [x] GET `/repos/{owner}/{name}/contents/rust-bors.toml` -- used to get bors config.

For Team API, the mock must include this endpoint: GET `/v1/permission/bors.{name}.{permission}.json`.

To implement this, I propose we create a `BorsTester` struct that facilitates the creation of the mock server, along with multiple queues used for events handling. This struct will also keep an in memory version github server, which will track all of the changes made through Github API.

@Kobzol Would love to hear your feedback on this.

Kobzol:
Wow, that's a lot of endpoints. But I still think it will be worth it. Yeah, this is a good plan, I'd just like to review it in smaller PRs, and not one big change xD

[...]

PR #109, opened 2024-06-05
Finish the move to end-to-end tests

Kobzol:
This PR implements the rest of the end-to-end mocking endpoints and moves most of existing tests to the end-to-end regime. The only remaining tests left to port are the refresh tests.

CC @vohoanglong0107

Closes: https://github.com/rust-lang/bors/issues/93


src/bors/mod.rs
```diff
@@ -1,87 +1,48 @@
...
+pub mod comment;
+mod context;
+pub mod event;
+mod handlers;
 
-    /// Get a workflow url.
-    fn get_workflow_url(&self, run_id: RunId) -> String;
+/// Loads repositories through a global GitHub client.
+pub struct RepositoryLoader {
+    client: Octocrab,
```
vohoanglong0107, line 28: I think we could remove this struct and just directly call load_repositories?

Kobzol: I started with that, but I decided to keep the struct in the end. Because we use a bunch of Octocrab clients in the app, but only the global one is usable for loading the repositories, and it's easy to accidentally mix them up. So this is now essentially just a typed wrapper to make it harder to use the wrong client for loading the repositories.

vohoanglong0107: I see. If that's the case, do you think RepositoryLoader and TeamApiClient should be combined? They are always used together to load/reload repositories, and like you said, both also use a global octocrab client

Kobzol: TeamApiClient doesn't use Octocrab, it doesn't communicate with GitHub :) But even then, I wouldn't combine them, they have different responsibilities.

src/tests/mocks/mod.rs
```diff
@@ -76,3 +112,29 @@ impl ExternalHttpMock {
...
+        .and(path_regex(regex))
+        .respond_with(move |req: &Request| {
+            let captured = parsed_regex
+                .captures(req.url.path())
+                .unwrap()
+                .extract::<N>()
+                .1;
+            f(req, captured)
+        })
+}
```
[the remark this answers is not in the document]
Kobzol: Yeah, that's the reason, so that we don't have to repeat the URL.

src/bors/mod.rs
vohoanglong0107: Hmmm, still, I don't really love this 
```rust
repo_loader.load_repositories(team_api_client).await
```
There should be other way to do this, right?

Kobzol: I don't know, is it that bad? Seems like normal dependency injection to me, the method needs a dependency, so we provide the dependency to it.

Btw, I probably misunderstood your "should be combined" suggestion. I thought that you meant to literally combine these two structs into one (which I wouldn't do), but you probably just meant to add `TeamApiClient` as a field into `RepositoryLoader`? We could do that, although then we'd need to either clone the client or put it behin a shared pointer, and I don't think doing that provides enough benefit here vs just passing it as an argument.

vohoanglong0107: > I don't know, is it that bad? Seems like normal dependency injection to me, the method needs a dependency, so we provide the dependency to it.

Yeah, I mean, it feels kinda awkward. Like, RepoLoader, load repositories for me, and then there is this TeamApiClient. But I'm just being nitpicky here.

>  We could do that, although then we'd need to either clone the client or put it behind a Rc pointer, and I don't think doing that provides enough benefit here vs just passing it as an argument.

Yeah it sounds more hassle than necessary. Let's just keep it that way.

src/tests/util.rs
```diff
@@ -0,0 +1,40 @@
...
+use tokio::sync;
+
+pub struct TestSyncMarker {
+    state: Lazy<TestSyncMarkerInner>,
+}
+
+impl TestSyncMarker {
+    pub const fn new() -> Self {
+        Self {
+            state: Lazy::new(|| TestSyncMarkerInner::new()),
```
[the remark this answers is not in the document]
Kobzol: Thanks!

vohoanglong0107 (approved):

Kobzol:
Resolved Clippy issues, will merge it now.

I'll take a look at https://github.com/rust-lang/bors/blob/263b7dfc2f1548a91f83d9b0b7009ebe81541e7a/src/bors/handlers/refresh.rs#L108, and then the Great End-to-End Migration™ should be completed.

PR #111, opened 2024-06-10
Fix refresh tests

Kobzol:
It turns out that they were broken because of a completely different reason than I thought. `TestSyncMarker` wasn't thread local, therefore it was being accessed by multiple tests in parallel.

This finally concludes the port to end-to-end tests. CC @vohoanglong0107


src/bors/handlers/refresh.rs
```diff
@@ -105,132 +105,131 @@ fn elapsed_time(date: DateTime<Utc>) -> Duration {
...
-//         MOCK_TIME.with(|time| {
-//             *time.borrow_mut() = None;
-//         });
-//     }
-// }
+#[cfg(test)]
+mod tests {
+    use crate::bors::handlers::refresh::MOCK_TIME;
+    use crate::bors::handlers::WAIT_FOR_WORKFLOW_STARTED;
+    use crate::bors::WAIT_FOR_REFRESH;
```
vohoanglong0107, line 112: I don't think it's necessary to wait for refresh event in any of the tests?

Kobzol: Well, we *do* need to wait, but the waiting is handled inside the `refresh` call, so this import is indeed unused. Good catch, thanks!

vohoanglong0107: Ah yeah, that's what I meant. Sorry for not expressing myself clearly😅

PR #112, opened 2024-06-10
Add a CI job for deploying to production

Kobzol:
This PR adds a CI job for deploying the bot to production. It works like this:
1) After a push to `main`, the bot is automatically deployed to the `staging` environment (this was pre-existing)
2) After a push to `main`, a workflow for deploying the bot to the `production` environment is started. This environment should be (manually) configured in this repo, so that it needs a review from someone (a bors admin, or multiple people).

The deployment history can be seen at https://github.com/rust-lang/bors/deployments, where admins can also approve the deployments.

Note that I have just copy-pasted the job from the staging environment for now. I'm not actually sure what should be changed to switch from staging to production - probably this: `arn:aws:iam::392478027976:role/gha-access`? I found (from some simpleinfra cache) that `392478027976` is the ID of the `bors-staging` account, but I'm not sure how to find this ID for the bors production account (it is not committed in simpleinfra, as far as I know).

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,9 +8,11 @@ on:
 concurrency: deploy
 
 jobs:
-  deploy:
+  deploy-to-staging:
     name: Deploy
     runs-on: ubuntu-latest
+    environment: staging
+    concurrency: staging
     permissions:
       id-token: write
       contents: read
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
351621253146 is the production account ID. It should have the same role provisioned as of ~now, and the app, db, etc. deployed in the same configuration as staging (modulo DNS name, which is bors-prod.rust-lang.net). If that's updated here my hope is that's enough for this to work, otherwise looks good to me.

PR #113, opened 2024-06-26
refactor: unify ways to get repository name

vohoanglong0107:
We currently have several methods to retrieve the repository name, which makes it a little bit confusing when deciding which one to use. This PR clean them up and adjust some visibility modifier along the way.

### DIFF
--- a/src/github/api/client.rs
+++ b/src/github/api/client.rs
@@ -13,27 +13,37 @@ use crate::github::{Branch, CommitSha, GithubRepoName, PullRequest, PullRequestN
 
 /// Provides access to a single app installation (repository) using the GitHub API.
 pub struct GithubRepositoryClient {
-    pub app: App,
+    app: App,
     /// The client caches the access token for this given repository and refreshes it once it
     /// expires.
-    pub client: Octocrab,
+    client: Octocrab,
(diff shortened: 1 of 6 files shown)


src/github/api/client.rs
```diff
@@ -266,19 +275,12 @@ impl GithubRepositoryClient {
...
-                    self.name().name
-                )
-            });
+            .unwrap_or_else(|| format!("{}/{}", base_github_html_url(), self.repository(),));
         format!("{html_url}/actions/runs/{run_id}")
     }
 
     fn format_pr(&self, pr: PullRequestNumber) -> String {
-        format!("{}/{}/{}", self.name().owner(), self.name().name(), pr)
+        format!("{}/{}", self.repository(), pr)
```
Kobzol, line 283: Let's hope no one changes the `Display` of `GithubRepoName` in the future :upside_down_face: But it should be fine to depend on it.

Kobzol (approved):
Thanks!

src/github/api/client.rs
vohoanglong0107: Let's hope we're be notified by a test failure when this happen 🤣

PR #114, opened 2024-06-28
feat: implement r+, r= and r-

vohoanglong0107:


src/bors/command/parser.rs
```diff
@@ -242,6 +287,38 @@ mod tests {
...
+        let cmds = parse_commands("@bors r+");
+        assert_eq!(cmds.len(), 1);
+        assert!(matches!(
+            cmds[0],
+            Ok(BorsCommand::Approve(Approver::Myself))
+        ));
+    }
+
+    #[test]
+    fn parse_approve_on_behalf() {
```
Kobzol, line 301: Maybe add another test for `bors r=foo,bar`, people sometimes use multiple names here.

src/config.rs
```diff
@@ -97,7 +101,20 @@ where
         }
     }
 
-    let triggers = HashMap::<Trigger, Vec<Modification>>::deserialize(deserializer)?;
+    let mut triggers = HashMap::<Trigger, Vec<Modification>>::deserialize(deserializer)?;
+    // If there are any `approve` triggers, add `unapprove` triggers as well.
```
Kobzol, line 105: This logic is taken from `homu`?

src/github/mod.rs
```diff
@@ -94,6 +104,29 @@ pub struct PullRequest {
     pub base: Branch,
     pub title: String,
     pub message: String,
+    pub author: GithubUser,
+}
+
+impl From<octocrab::models::pulls::PullRequest> for PullRequest {
```
Kobzol, line 110: Nice cleanup :+1:

src/bors/handlers/review.rs
```diff
@@ -0,0 +1,261 @@
...
+            pr.number,
+            Comment::new(format!(
+                "Commit {} has been approved by `{}`",
+                pr.head.sha, approver
+            )),
+        )
+        .await
+}
+
+fn insufficient_unapprove_permission(
```
Kobzol, line 101: I find it easier to follow if the logic is positive, rather than negative. Could you please change this into something like `has_[un]approve_permissions`?

src/bors/handlers/review.rs
```diff
@@ -0,0 +1,261 @@
...
+    let approver = match approver {
+        Approver::Myself => author.username.clone(),
+        Approver::Specified(approver) => approver.clone(),
+    };
+    tracing::info!("Approving PR {} by {}", pr.number, approver);
+    db.approve(repo_state.repository(), pr.number, approver.as_str())
+        .await?;
+    tracing::info!("Approved PR {}", pr.number);
+    handle_label_trigger(&repo_state, pr.number, LabelTrigger::Approved).await?;
+    tracing::info!("Label trigger handled for PR {}", pr.number);
```
Kobzol, line 36: This log seems superfluous, such logs are already in `handle_label_trigger`.

src/bors/handlers/review.rs
```diff
@@ -0,0 +1,261 @@
...
+    db: Arc<PgDbClient>,
+    pr: &PullRequest,
+    author: &GithubUser,
+) -> anyhow::Result<()> {
+    tracing::info!("Unapproving PR {}", pr.number);
+    if insufficient_unapprove_permission(repo_state.clone(), pr, author) {
+        deny_unapprove_request(&repo_state, pr, author).await?;
+        return Ok(());
+    };
+    db.unapprove(repo_state.repository(), pr.number).await?;
```
Kobzol, line 53: Add log with `Unapproved PR`, to mirror what approving does?

src/bors/handlers/review.rs
```diff
@@ -0,0 +1,261 @@
...
+        );
+        BorsBuilder::new(pool)
+            .world(world)
+            .run_test(|mut tester| async {
+                tester.post_comment("@bors r+").await?;
+                assert_eq!(
+                    tester.get_comment().await?,
+                    format!(
+                        "Commit pr-{}-sha has been approved by `{}`",
+                        default_pr_number(),
```
Kobzol, line 163: I think it's find to hardcode these values here to make the test simpler. If we ever change them in the future, the test will just fail.

src/bors/handlers/review.rs
```diff
@@ -0,0 +1,261 @@
...
+    repo.client
+        .post_comment(
+            pr.number,
+            Comment::new(format!("Commit {} has been unapproved", pr.head.sha)),
+        )
+        .await
+}
+
+#[cfg(test)]
+mod tests {
```
Kobzol, line 143: These tests only check the labels, we should also look inside the DB after the tests to see if the status has been changed there. You could introduce a helper that takes the `World` (or `Borstester`) and checks approved/unapproved status in the DB and also in the labels.

Kobzol (commented):
Great work! Left a couple of nits, but otherwise looks very clean.

src/bors/handlers/review.rs
vohoanglong0107: Ah I forgot to remove those 😅

src/config.rs
vohoanglong0107: I couldn't find any logical way to remove the labels added after approving a PR, so I just add a section labels to add/remove after unapproving a PR as well.

src/bors/handlers/review.rs
vohoanglong0107: Ah the reason I'm not hardcoding the values in was I didn't know where those values come from, and this was my attempt to explain the origin of those values for future me 😅

Kobzol: That's not a bad reason! I hope we could make the default values so obvious that this wouldn't be needed though, and it would be immediately obvious from the test that `1` or `default-user` etc. is a special value :)

Kobzol (approved):
Thanks! :)

PR #115, opened 2024-07-12
feat: unapprove on PR's base changed

vohoanglong0107:
This is homu's default behavior

### DIFF
--- a/src/bors/handlers/review.rs
+++ b/src/bors/handlers/review.rs
@@ -183,7 +183,8 @@ mod tests {
     use crate::{
         github::PullRequestNumber,
         tests::mocks::{
-            default_pr_number, default_repo_name, BorsBuilder, BorsTester, Permissions, User, World,
+            default_pr_number, default_repo_name, BorsBuilder, BorsTester, Permissions,
+            PullRequestChangeEvent, User, World,
         },
     };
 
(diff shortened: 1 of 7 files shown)


Kobzol:
Homu unapproves the PR when the PR description is modified? :thinking: I thought that it only does it when a new commit has been pushed (which we should also do in bors eventually :laughing:). Could you please link the original code from homu?

vohoanglong0107:
My bad, what I wanted to do was to unapprove pull request when the author push a new commit, but it appears that the correct event to used was not edit? [Here](https://github.com/rust-lang/homu/blob/f652a9e7ecc42a1a590f48e96d9d34df4596aedb/homu/server.py#L542) is the original code from homu

Kobzol:
Right, so this is actually a very special and rare case (that we should of course handle eventually, as with everything :sweat_smile:). The `edited` event being handled by `homu` deals with changing the base branch of a PR. This is a super rare operation, where you e.g. open a PR against `main`, and then you modify the PR so that it is opened e.g. against `beta`, or something like that. In that case the PR needs to be unapproved, of course.

We should definitely handle this too, so might as well keep the PR. We should be checking if the base ref [has changed](https://github.com/rust-lang/homu/blob/f652a9e7ecc42a1a590f48e96d9d34df4596aedb/homu/server.py#L546), though. Otherwise the PR would be unapproved on any edit.

Another thing that homu seems to do when the PR changes is to store its description and title into its DB. This is important, because these texts are used for commit messages made by homu. But we're not storing this info yet in bors.

vohoanglong0107:
Regarding the PR's description and title, since we are retrieving the PR information before we handle any operations, I think we postpone storing it for now

src/bors/handlers/review.rs
```diff
@@ -52,6 +55,21 @@ pub(super) async fn command_unapprove(
     notify_of_unapproval(&repo_state, pr).await
 }
 
+pub(super) async fn handle_pull_request_edited(
+    repo_state: Arc<RepositoryState>,
+    db: Arc<PgDbClient>,
+    payload: PullRequestEdited,
+) -> anyhow::Result<()> {
+    // If the base branch has changed, unapprove the PR
```
Kobzol, line 63: Very nice that we can get this directly from the GH webhook, instead of having to store the original base ref in our DB! Well, we will need to store it anyway, so that we know with which branch to actually merge :laughing: But still.

src/bors/handlers/review.rs
```diff
@@ -135,6 +153,23 @@ async fn notify_of_unapproval(repo: &RepositoryState, pr: &PullRequest) -> anyho
...
+async fn notify_of_edited_pr(
+    repo: &RepositoryState,
+    pr_number: PullRequestNumber,
+    head_sha: CommitSha,
+) -> anyhow::Result<()> {
+    repo.client
+        .post_comment(
+            pr_number,
+            Comment::new(format!(
+                r#":warning: The base branch changed to `{}`, and the
```
Kobzol, line 165: This message doesn't make sense if the PR was not approved in the first place. We should only send this message when the PR was originally approved.

src/bors/handlers/review.rs
```diff
@@ -52,6 +55,21 @@ pub(super) async fn command_unapprove(
...
+    db: Arc<PgDbClient>,
+    payload: PullRequestEdited,
+) -> anyhow::Result<()> {
+    // If the base branch has changed, unapprove the PR
+    let Some(_) = payload.from_base_sha else {
+        return Ok(());
+    };
+    let pr_number = payload.pull_request.number;
+    db.unapprove(repo_state.repository(), pr_number).await?;
+    handle_label_trigger(&repo_state, pr_number, LabelTrigger::Unapproved).await?;
```
Kobzol, line 69: Same here, the label should only be triggered if the PR was approved before.

Kobzol (commented):
Looks good! What is missing are tests, we should test what happens when the ref is changed when the PR is/isn't approved, that it sends a comment, that it gets unapproved in the DB etc.

src/bors/handlers/review.rs
```diff
@@ -52,6 +55,29 @@ pub(super) async fn command_unapprove(
...
+) -> anyhow::Result<()> {
+    // If the base branch has changed, unapprove the PR
+    let Some(_) = payload.from_base_sha else {
+        return Ok(());
+    };
+
+    let pr_model = db
+        .get_or_create_pull_request(repo_state.repository(), payload.pull_request.number)
+        .await?;
+    if pr_model.approved_by.is_none() {
```
Kobzol, line 71: We will probably ask this question in more places, so it would be good to add a more high-level query for this (simply `pr_model.is_approved()`. We can do that in some follow-up PR.

Kobzol (approved):
Cool, thank you!

PR #116, opened 2024-07-13
refactor: split webhook parser to multiple function

vohoanglong0107:
To prevent stack overflow in dev


Kobzol (approved):
Much more readable, thanks!

PR #117, opened 2024-07-24
fix: ignore clippy warning on variant's size

vohoanglong0107:
The global event variant is rarely used (compare to the per repository event), so there won't be much improvement if we try to fix this.

### DIFF
--- a/src/bors/event.rs
+++ b/src/bors/event.rs
@@ -38,6 +38,7 @@ pub enum BorsGlobalEvent {
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


Kobzol:
Well, it's kind of the purpose of the lint to warn about this exact situation, because even though the global event variant is rarely used, the fact that the enum is unnecessarily large can negatively affect the performance of any usage of that enum :)

But I don't think that this is performance critical in any way (I would be more worried about stack usage in Debug mode :sweat_smile:), so I don't have an issue with this :)

PR #119, opened 2024-07-24
refactor: allow more information for pull request creation

vohoanglong0107:
# Context

I'm working on un-approving pr whenever new commits were pushed against the branch, but it requires adding a branch column to `pull_request` table. New pr creation will need branch information, so I made this refactor to make the changes easier

### DIFF
--- a/src/bors/handlers/review.rs
+++ b/src/bors/handlers/review.rs
@@ -30,7 +30,7 @@ pub(super) async fn command_approve(
         Approver::Myself => author.username.clone(),
         Approver::Specified(approver) => approver.clone(),
     };
-    db.approve(repo_state.repository(), pr.number, approver.as_str())
+    db.approve(repo_state.repository(), pr, approver.as_str())
         .await?;
     handle_label_trigger(&repo_state, pr.number, LabelTrigger::Approved).await?;
     notify_of_approval(&repo_state, pr, approver.as_str()).await
@@ -49,7 +49,7 @@ pub(super) async fn command_unapprove(
         deny_unapprove_request(&repo_state, pr, author).await?;
         return Ok(());
     };
-    db.unapprove(repo_state.repository(), pr.number).await?;
+    db.unapprove(repo_state.repository(), pr).await?;
     handle_label_trigger(&repo_state, pr.number, LabelTrigger::Unapproved).await?;
     notify_of_unapproval(&repo_state, pr).await
 }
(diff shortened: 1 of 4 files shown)


src/bors/handlers/review.rs
```diff
@@ -30,7 +30,7 @@ pub(super) async fn command_approve(
         Approver::Myself => author.username.clone(),
         Approver::Specified(approver) => approver.clone(),
     };
-    db.approve(repo_state.repository(), pr.number, approver.as_str())
+    db.approve(repo_state.repository(), pr, approver.as_str())
```
Kobzol, line 33: I understand that we need to store more data about PRs in the database. First, I assumed that we will rely on the PR edited event to update data in our DB, and then just approve based on the PR number. But if I understand this correctly, you want to directly do PR update + approval in the DB, so that it stays atomic?

This sounds like a good idea, to avoid situations where we get a race condition and approve something else than what you see on GH. Just wanted to confirm that this was the motivation :)

vohoanglong0107:
Actually, when there is push event we aren't informed of which pr the push event belonged to, only the ref of the branch, so I need to store the branch name to perform pr extraction by name

src/bors/handlers/review.rs
vohoanglong0107: I didn't think of it like that 😅 My intention was only this
> I assumed that we will rely on the PR edited event to update data in our DB, and then just approve based on the PR number

Kobzol: Well :laughing: In any case, I think that using the latest information from the PR struct, since we get it from the webhook anyway, is a better idea, to be 100% sure of what commit we are approving.

Kobzol:
I think we can use the `synchronize` PR event to detect that: https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=synchronize#pull_request That should be much easier.

vohoanglong0107:
This is the first time I've heard about `synchronize` event. Let me try it. If it's possible to use it, perhaps this PR could be put on wait until we need to perform merge events

src/bors/handlers/review.rs
vohoanglong0107: Yeah, this pr instance was retrieved from either the webhook or by ourselves using gh_client.get_pull_request so it's 100% sure that the latest data was presented when we performed approval

Kobzol: Using a GH API call could actually lead to a race condition (approve -> push -> get PR from API). That's why I like reading the PR from the approval webhook directly :)

vohoanglong0107: That's a real issue.
Hmm, how can we fix it then, given that webhook payload sent to us when we approve pr (issue_comment) doesn't include the pull request information inside of it?

Kobzol: Well, we can't :) I thought that the PR payload is in the webhook, but since it isn't, we just need to do the GH API call, or depend on what we have in the DB. Let's keep it as it is now and do the API call. The race condition should be rare, and on GH you can check which SHA was approved, since the bot should include it in the "approved" comment.

Kobzol:
The code has changed considerably in the meantime, I think that passing the individual attributes is fine for now.

PR #120, opened 2024-07-24
feat: unapprove on push

vohoanglong0107:
Almost the same as unapprove on base edit


src/bors/event.rs
```diff
@@ -6,6 +6,8 @@ use octocrab::models::RunId;
 pub enum BorsRepositoryEvent {
     /// A comment was posted on a pull request.
     Comment(PullRequestComment),
+    /// When a new commit is pushed to the pull request branch.
+    PullRequestPushed(PullRequestPushed),
```
Kobzol, line 10:
```suggestion
    PullRequestCommitPushed(PullRequestCommitPushed),
```
or something like that? Otherwise it sounds like a PR was pushed, which does not make sense (it sounds more like a PR was created instead).

src/bors/handlers/review.rs
```diff
@@ -77,6 +79,26 @@ pub(super) async fn handle_pull_request_edited(
...
+        .get_or_create_pull_request(repo_state.repository(), pr.number)
+        .await?;
+
+    if !pr_model.is_approved() {
+        return Ok(());
+    }
+
+    let pr_number = pr_model.number;
+    db.unapprove(repo_state.repository(), pr_number).await?;
+    handle_label_trigger(&repo_state, pr_number, LabelTrigger::Unapproved).await?;
```
Kobzol, line 98: This seems exactly the same as in the case when the base changes. Maybe we can introduce a helper function for unapproving a PR if it's approved?

src/bors/handlers/review.rs
```diff
@@ -176,27 +198,37 @@ PR will need to be re-approved."#,
...
+async fn notify_of_pushed_pr(
+    repo: &RepositoryState,
+    pr_number: PullRequestNumber,
+    head_sha: CommitSha,
+) -> anyhow::Result<()> {
+    repo.client
+        .post_comment(
+            pr_number,
+            Comment::new(format!(
+                r#":warning: Branch's head was changed to `{}`, the
```
Kobzol, line 210: This sounds a bit cryptic, I'd mention the fact that a new commit was pushed to the branch, that will be more obvious.

src/github/webhook.rs
```diff
@@ -204,24 +204,31 @@ fn parse_issue_comment_event(body: &[u8]) -> anyhow::Result<Option<BorsEvent>> {
...
-    if payload.action == PullRequestEventAction::Edited {
-        let Some(changes) = payload.changes else {
-            return Err(anyhow::anyhow!(
-                "Edited pull request event should have `changes` field"
-            ));
-        };
-        Ok(Some(BorsEvent::Repository(
-            BorsRepositoryEvent::PullRequestEdited(PullRequestEdited {
-                repository: repository_name,
+    match payload.action {
```
Kobzol, line 207: I forgot about it before, but we have some tests in this file for actually parsing the GH webhooks for the individual webhooks. Could you please add new tests for PR edited + PR synchronize? It's also nice to have some actual reference data for the webhook's HTTP contents.

Kobzol (commented):
Thank you! Left a few notes, nothing major.

src/bors/handlers/review.rs
vohoanglong0107: I don't really like how this turned out https://github.com/rust-lang/bors/pull/120/commits/7f9606d8af1f89fe1043f49b4e1906378d4b81af. What do you think?

src/bors/handlers/review.rs
```diff
@@ -85,18 +78,10 @@ pub(super) async fn handle_push_to_pull_request(
...
-
-    if !pr_model.is_approved() {
-        return Ok(());
-    }
-
-    let pr_number = pr_model.number;
-    db.unapprove(repo_state.repository(), pr_number).await?;
-    handle_label_trigger(&repo_state, pr_number, LabelTrigger::Unapproved).await?;
-    notify_of_pushed_pr(&repo_state, pr_number, pr.head.sha.clone()).await
+    unapprove_if_approved(&repo_state, &db, pr, || async {
```
Kobzol, line 81: I meant just the shared part before the last line. Can't we just do
```rust
unapprove_if_approved(&repo_state, &db, pr).await?;
notify_of_pushed_pr(&repo_state, pr.number, pr.head.sha.clone()).await
```
? :sweat_smile:

vohoanglong0107: If we do that it will still post a comment event if this PR is not approved 😅

Kobzol: Right, so what about returning `Result<bool>`?

vohoanglong0107: Hmmm, I'd prefer we kept the original version. IMO condition function should be pure, so I'm not really comfortable with returning `Result<bool>`

Kobzol: I'm fine with going back to the original, but I'm intrigued by your last sentence :) What's a "condition function" and how does this interact with pure functions? Pretty sure no functions in this file are pure :D

Kobzol: Ah, I see. The function is already called `unapprove_if_approved`, so it seems natural to me that it signals to the caller if it actually did perform the unapproval or not. But up to you.

vohoanglong0107:
Sorry for the lack of updates on this 😓 I messed up my computer and it's taking awhile to get it up running again

Kobzol:
Absolutely no worries. Given that my PhD deadline is in ~30 days, I don't have plenty of time to work on bors atm myself..

Kobzol (approved):
Looks great, thank you!

PR #126, opened 2024-08-29
chore: cargo update

MarcoIeni:
Run `cargo update` to solve a compilation issue

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
+checksum = "6e4503c46a5c0c7844e948c9a4d6acd9f50cccb4de1c48eb9e291ea17470c678"
 dependencies = [
  "gimli",
 ]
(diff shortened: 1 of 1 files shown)


Kobzol (commented):
What compilation issue that was, out of curiosity? It would be nice to do this in a bit more targeted manner. This increases the amount of compiled packages by 2, which is not terrible (296 -> 298), but I noticed that we now compile three different versions of rustls, which isn't great (we already had two versions before).

MarcoIeni:
```
$ cargo check
    Checking hyper v1.2.0
   Compiling syn v1.0.109
    Checking time v0.3.34
    Checking regex-syntax v0.8.3
error[E0282]: type annotations needed for `Box<_>`
  --> /Users/marco/.cargo/registry/src/index.crates.io-6f17d22bba15001f/time-0.3.34/src/format_description/parse/mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
help: consider giving `items` an explicit type, where the placeholders `_` are specified
   |
83 |     let items: Box<_> = format_items
   |              ++++++++

    Checking regex-automata v0.4.6
For more information about this error, try `rustc --explain E0282`.
error: could not compile `time` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
```

> It would be nice to do this in a bit more targeted manner.

Imo doing `cargo update` in a surgical way is just a waste of our time. The `Cargo.toml` contains outdated dependencies already, and once we updated them, this time might be wasted (because new dependencies are pulled in).

My ideal setup is: renovate runs `cargo update` periodically and automatically merges the PR if tests pass.

So I would trade "inefficient compilations" for "just run cargo update, who cares". But the final decision is yours of course :)
Let me know if you want me to close this PR and open a new one with a precise cargo.lock update 👍

Kobzol:
I agree that running renovatebot is easier and ensures updated deps, but we shouldn't update them just for the sake of updating. When a dependency update unnecessarily includes a new version of the same dependency in our build tree, then we should either try to fix that or wait with the update until other deps catch up so that the duplicate version is removed. Just running cargo update is not great when it makes our dep tree worse.

Once I'm at a PC, I'll try to take a look if we can merge these deps somehow.

MarcoIeni:
Ok, I see you point!

> Once I'm at a PC, I'll try to take a look if we can merge these deps somehow.

I can also do it if you prefer :)

Kobzol (approved):
Thanks! :)

Issue #122, opened 2024-08-07
Automatic dependency update


MarcoIeni:
It would be nice to setup renovate to automatically update Cargo dependencies and Dockerfile versions (e.g. rust and ubuntu).

This would allow us to stay on top of our dependencies, so we can patch easily security vulnerabilities.

Examples:

- https://github.com/rust-lang/crates.io/blob/main/.github/renovate.json5
- https://github.com/jdno/infra-smoke-tests/blob/main/.github/renovate.json5

PR #127, opened 2024-08-29
Chore add renovate

MarcoIeni:
close https://github.com/rust-lang/bors/issues/122

This very simple configuration does 99% of what we need I think :)

Tested [here](https://github.com/MarcoIeni/renovate-playground/issues/3) 👍

If you see the PRs of the repository you can see what renovate does.

I'm using the hosted version of renovate as https://github.com/rust-lang/crates.io does.

I'm not sure which account we should use to add the GitHub app and authenticate into https://developer.mend.io/
We can ask the crates.io team how they do this if we have doubts.
Ideally everyone in the [infra-bors](https://github.com/rust-lang/team/blob/master/teams/infra-bors.toml) team should be able to access the renovate dashboard. But I guess we can fix this issue later? Also, you can control renovate by checking the boxes in the Dashboard [issue](https://github.com/MarcoIeni/renovate-playground/issues/3) 👍

It looks like this:

image


Kobzol:
We already have support for renovatebot in `team` (https://github.com/rust-lang/team/blob/7922f9bd1b0126b4fa403d313694f91e450f200d/repos/rust-lang/cargo.toml#L5), where we can install the renovatebot app to selected repositories. Is both the config file and the GitHub app installation needed?

MarcoIeni:
sync-team only [installs](https://github.com/rust-lang/sync-team/blob/f235bd801a9fce0c57d5c698019cb477e03bdb90/src/github/mod.rs#L436) the GitHub app.

From my understanding, if the configuration is missing, we will get a PR from renovate that adds it. Explained [here](https://docs.renovatebot.com/getting-started/installing-onboarding/#repository-onboarding).

I created https://github.com/rust-lang/team/pull/1539

We can:
1. Merge https://github.com/rust-lang/team/pull/1539
2. Wait for renovate to open the configuration pr
3. Check how the default configuration looks like
4. Merge this PR if we aren't satisfied with it 👍

What do you think?

MarcoIeni:
Here's the generated pr
https://github.com/rust-lang/bors/pull/128

.github/renovate.json5
```diff
@@ -0,0 +1,4 @@
+{
+  $schema: "https://docs.renovatebot.com/renovate-schema.json",
+  extends: ["config:recommended", ":maintainLockFilesWeekly"],
```
Kobzol, line 3: Does this mean that we would get a new PR every week? I'd prefer every month instead.

Issue #121, opened 2024-08-07
Add a container health check


MarcoIeni:
From the AWS console, the Health status of the task is Unknown:

<img width="1211" alt="image" src="https://github.com/user-attachments/assets/73a59ef2-3c4c-40b7-aa88-f00ab6fa09cb">

<img width="368" alt="image" src="https://github.com/user-attachments/assets/34a71c72-96a9-4ca3-a756-c28cf59bb205">

I found the docs on how to add an health check [here](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_HealthCheck.html).

We could define an health check to tell AWS when the app is ready to accept requests. E.g. the health check can be an http endpoint, so that we start receiving requests after the http server started.

What do you think?

Kobzol:
Yeah, that makes sense, should be easy to implement.

MarcoIeni:
the aws dashboard still shows "unknown" so I need to investigate

MarcoIeni:
https://github.com/rust-lang/simpleinfra/pull/559
I'll merge this tomorrow and check if it works in staging 👍

Kobzol:
Thanks! Bors is an important service (well, this specific bors instance hopefully will be important one day :laughing:), so if we can get a health check that could perhaps even alert us in some way if it goes offline, that would be nice.

MarcoIeni:
> this specific bors instance hopefully will be important one day 😆

It will 😁

This is now done. Closing this 🎉

PR #131, opened 2024-09-03
chore(docker): add healthcheck

MarcoIeni:
We should already have a test that covers this.
In fact, if [this](https://github.com/MarcoIeni/bors/blob/28390e0e7febe41ec8a496183e26a00ca932c34d/.github/workflows/test.yml#L77) CI job passes, it means that the healthcheck passed as well 👍 

Related to #121


Dockerfile
```diff
@@ -29,10 +29,14 @@ FROM ubuntu:22.04 as runtime
...
 
-RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y ca-certificates
+# curl is needed for healthcheck
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y ca-certificates curl
 
 COPY --from=build /app/target/release/bors .
 
 EXPOSE 80
 
+HEALTHCHECK --timeout=10s --start-period=1s \
```
Kobzol, line 39: I'm not familiar with Docker healthchecks, what does this actually do? Or rather, I understand what it does, but why do we need it?

MarcoIeni: To fix https://github.com/rust-lang/bors/issues/121

AWS uses this to check if the app is ready and it's running fine. E.g. it might be stuck due to high traffic or not ready yet.

Kobzol: I see, so AWS uses the `HEALTHCHECK` machinery from the Dockerfile? I thought that it just hardcores the `/health` endpoint :)

MarcoIeni: yes. this is my understanding from the docs: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_HealthCheck.html

Kobzol: Ok. I would increase the start period from 1s to 10s, otherwise LGTM.

Kobzol (approved):
Can you check after this is merged that AWS actually uses the healthcheck?

MarcoIeni:
of course!

PR #154, opened 2024-09-18
chore(renovate): don't update patches

### LINKED ISSUE #153: Avoid renovate sending patch updates too often
We don't want renovate to send PRs like https://github.com/rust-lang/bors/pull/150 or https://github.com/rust-lang/bors/pull/149
These kind of patches should only be done for security purposes.

MarcoIeni:
Close #153 

This should prevent Cargo.lock updates from `a.b.c` to `a.b.(c+1)`.

From my understanding, unfortunately disabling `minor` as well means that we'll stop get updates from `0.x.y` to `0.(x+1).0`, so I haven't disabled minor.

However ideally we want to disable updates from `a.b.c` to `a.(b+1).0`. Maybe something we can investigate further in the future.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -5,4 +5,13 @@
     ":maintainLockFilesMonthly",
     "docker:disable",
   ],
+  packageRules: [
+    {
+      matchCategories: ["rust"],
+      updateTypes: ["patch"],
+      // Disable patch updates for single dependencies because patches
+      // are updated periodically with lockfile maintainance.
+      enabled: false,
+    },
+  ],
 }


Kobzol:
We could probably group the Rust PRs into groups that are merged less frequently and enable only security vulnerability patches to be applied, but this seems like a quick solution that should work.

However, I'm not sure if this also doesn't disable security patch updates, that would be bad.

Maybe we also need to enable these explicitly? https://github.com/renovatebot/renovate/discussions/15490#discussioncomment-2702421

Kobzol:
I don't have access to settings on this repo :) I agree that we can merge this for now, thanks!

PR #161, opened 2024-10-11
chore(deps): update secrecy 0.8.0 -> 0.10.3

alexpovel:
I diffed the actual crates.io contents via

```bash
curl -L https://crates.io/api/v1/crates/secrecy/${VERSION}/download | tar -xf -
```

with `VERSION` as `0.8.0` and `0.10.3`, then
diffing via git. It revealed nothing suspicious,
with changes as claimed per their CHANGELOG.md.

0.10.0 had some breaking changes. `Secret<T>` is
no longer a thing, instead there's `SecretBox<T>`. It turned out secrecy usage in bors is limited,
and private keys are fed as PEM-encoded `String`s. For this, there's `SecretString`, which is just an alias for `SecretBox<str>`. Accepting it and
converting to bytes in `create_github_client`
simplifies code at the call sites a bit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -291,7 +291,7 @@ dependencies = [
  "parking_lot",
  "regex",
  "reqwest",
- "secrecy",
+ "secrecy 0.10.3",
  "serde",
  "serde_json",
  "sha2",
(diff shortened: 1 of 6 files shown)


MarcoIeni (approved):
Looks great, thanks! 🙏

Kobzol:
Thank you! :) PEM should hopefully be "just text", so UTF-8 encoding should handle it just fine.

Issue #84, opened 2024-05-05
Add a minimum CI run duration


Kobzol:
Sometimes, when [someone](https://github.com/rust-lang/rust/pull/124631) does a stupid thing and it completely removes all tests from CI, the CI workflows can finish very quickly, and e.g. run no jobs. Bors should have some configurable default to avoid accepting a merge when CI took less than a pre-defined time.

MarcoIeni:
In case anyone wants to work on this, here's how you can add the field to the config: https://github.com/rust-lang/bors/compare/main...MarcoIeni:bors:min-ci

PR #164, opened 2024-10-14
feat: added min duaration checking in RepositoryConfig

sriganeshres:
Fixes: https://github.com/rust-lang/bors/issues/84

I have added a serilazed min_ci_duartion field which takes care that atleast 60 seconds of testing is being done.

### DIFF
--- a/src/bors/event.rs
+++ b/src/bors/event.rs
@@ -1,5 +1,6 @@
 use crate::database::{WorkflowStatus, WorkflowType};
 use crate::github::{CommitSha, GithubRepoName, GithubUser, PullRequest, PullRequestNumber};
+use chrono::Duration;
 use octocrab::models::RunId;
 
 #[derive(Debug)]
@@ -88,6 +89,7 @@ pub struct WorkflowCompleted {
     pub commit_sha: CommitSha,
     pub run_id: RunId,
     pub status: WorkflowStatus,
+    pub running_time: Option<Duration>,
 }
 
 #[derive(Debug)]
--- a/src/bors/handlers/workflow.rs
+++ b/src/bors/handlers/workflow.rs
@@ -1,4 +1,5 @@
 use std::sync::Arc;
+use std::time::Duration;
 
 use crate::bors::comment::try_build_succeeded_comment;
 use crate::bors::event::{CheckSuiteCompleted, WorkflowCompleted, WorkflowStarted};
@@ -61,10 +62,27 @@ pub(super) async fn handle_workflow_started(
 pub(super) async fn handle_workflow_completed(
     repo: Arc<RepositoryState>,
     db: Arc<PgDbClient>,
-    payload: WorkflowCompleted,
+    mut payload: WorkflowCompleted,
 ) -> anyhow::Result<()> {
     if !is_bors_observed_branch(&payload.branch) {
         return Ok(());
+    }
+
+    if let Some(running_time) = payload.running_time {
+        let running_time_as_duration =
+            chrono::Duration::to_std(&running_time).unwrap_or(Duration::from_secs(0));
+        if let Some(min_ci_time) = repo.config.load().min_ci_time {
+            if running_time_as_duration < min_ci_time {
+                payload.status = WorkflowStatus::Failure;
+                tracing::warn!(
+                    "Workflow running time is less than the minimum CI duration: {:?} < {:?}",
+                    running_time_as_duration,
+                    min_ci_time
+                );
+            }
+        }
+    } else {
+        tracing::warn!("Running time is not available.");
     }
 
     tracing::info!("Updating status of workflow to {:?}", payload.status);
--- a/src/config.rs
+++ b/src/config.rs
@@ -19,10 +19,21 @@ pub struct RepositoryConfig {
     pub timeout: Duration,
     #[serde(default, deserialize_with = "deserialize_labels")]
     pub labels: HashMap<LabelTrigger, Vec<LabelModification>>,
+    #[serde(default, deserialize_with = "deserialize_duration_from_secs_opt")]
+    pub min_ci_time: Option<Duration>,
 }
 
 fn default_timeout() -> Duration {
     Duration::from_secs(3600)
+}
+
+fn deserialize_duration_from_secs_opt<'de, D>(deserializer: D) -> Result<Option<Duration>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    // Allow null values for the option
+    let maybe_seconds = Option::<u64>::deserialize(deserializer)?;
+    Ok(maybe_seconds.map(Duration::from_secs))
 }
 
 fn deserialize_duration_from_secs<'de, D>(deserializer: D) -> Result<Duration, D::Error>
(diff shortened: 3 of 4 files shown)


Kobzol:
Hi! This is a good start, but the bot should actually use this value, and check that the CI workflow has not in fact completed in a shorter time than this.

sriganeshres:
I will check on that

src/bors/handlers/workflow.rs
```diff
@@ -77,6 +78,21 @@ pub(super) async fn handle_workflow_completed(
         branch: payload.branch,
         commit_sha: payload.commit_sha,
     };
+
+    if let Some(running_time) = payload.running_time {
```
Kobzol, line 82: When we find out that the workflow ran too quickly, we should change its status to failed, and then act as it was an actual failure (with some custom error code).

Kobzol (commented):
Thanks for the PR, it's a good start. Left one comment about the actual logic that should happen on failure.

We should also add some basic tests for this. You can check the existing tests to see how we define them.

src/config.rs
```diff
@@ -142,6 +151,20 @@ mod tests {
         assert_eq!(config.timeout.as_secs(), 3600);
     }
 
+    #[test]
+    fn deserilaize_min_ci_time_empty() {
```
Kobzol, line 155:
```suggestion
    fn deserialize_min_ci_time_empty() {
```

src/config.rs
```diff
@@ -142,6 +151,20 @@ mod tests {
...
+        let content = "";
+        let config = load_config(content);
+        assert_eq!(config.min_ci_time, default_min_ci_duration());
+    }
+
+    #[test]
+    fn deserilaize_min_ci_time() {
```
Kobzol, line 162:
```suggestion
    fn deserialize_min_ci_time() {
```

src/bors/event.rs
```diff
@@ -1,5 +1,6 @@
 use crate::database::{WorkflowStatus, WorkflowType};
 use crate::github::{CommitSha, GithubRepoName, GithubUser, PullRequest, PullRequestNumber};
+use chrono::Duration;
```
Kobzol, line 3: Let's use `std::time::Duration` here, to avoid the casting between chrono and std.

src/bors/handlers/workflow.rs
```diff
@@ -77,6 +78,25 @@ pub(super) async fn handle_workflow_completed(
...
+        let running_time_as_duration =
```
Kobzol, line 83: The status of the workflow should be modified before we store the status into the DB (`update_workflow_status`). I would also post a comment to the PR (see `find_pr_by_build`), because this should be a rare occurrence that should be immediately visible.

Kobzol (commented):
I left a few more comments. You can find tests e.g. at the bottom of the `workflow.rs` file.

To run tests locally, you'll need to setup a DB, which I tried to describe in `docs/development.md`. Let me know if you have any questions about that.

src/bors/event.rs
sriganeshres: i used chrono beacuase it is what used by github.

src/bors/handlers/workflow.rs
```diff
@@ -61,12 +62,30 @@ pub(super) async fn handle_workflow_started(
...
+                "Workflow running time is less than the minimum CI duration: {:?} < {:?}",
+                payload.running_time,
+                repo.config.load().min_ci_time
+            );
+            return Err(anyhow::anyhow!(
+                "Workflow running time is less than the minimum CI duration"
+            ));
+        }
+    } else {
+        tracing::info!("Running time is not available.");
```
Kobzol, line 86:
```suggestion
        tracing::warn!("Running time is not available.");
```

src/bors/handlers/workflow.rs
```diff
@@ -61,12 +62,30 @@ pub(super) async fn handle_workflow_started(
...
+        let running_time_as_duration =
+            chrono::Duration::to_std(&running_time).unwrap_or(Duration::from_secs(0));
+        if running_time_as_duration < repo.config.load().min_ci_time {
+            payload.status = WorkflowStatus::Failure;
+            tracing::warn!(
+                "Workflow running time is less than the minimum CI duration: {:?} < {:?}",
+                payload.running_time,
+                repo.config.load().min_ci_time
+            );
+            return Err(anyhow::anyhow!(
```
Kobzol, line 81: We should not bail out here, but rather continue with the execution of the function. Currently, we won't be able to distinguish a normal failure with a "too short duration" failure, but I suppose that we can do that in another PR, e.g. by generalizing `WorkflowStatus` to include this information there.

Kobzol (commented):
Sorry for the long review time, I was (and still am) very busy.

Left a few comments, and we still need tests. If you don't want to work on them, it's fine, I can add them later.

src/config.rs
```diff
@@ -19,12 +19,21 @@ pub struct RepositoryConfig {
     pub timeout: Duration,
     #[serde(default, deserialize_with = "deserialize_labels")]
     pub labels: HashMap<LabelTrigger, Vec<LabelModification>>,
+    #[serde(
+        default = "default_min_ci_duration",
+        deserialize_with = "deserialize_duration_from_secs"
+    )]
+    pub min_ci_time: Duration,
```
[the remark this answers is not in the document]
sriganeshres: So I have made it zero seconds if not mentioned so the result would be similar. What do you say?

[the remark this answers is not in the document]
Kobzol: That's very non-obvious, we should model it in the type system explicitly.

src/bors/handlers/workflow.rs
sriganeshres: But the requirement is that if it is failing to give us some minimum time build should not go through rather it should fail. Can you give more information about what should happen. Should it not return a Error.

src/bors/handlers/workflow.rs
sriganeshres: Changed this.

src/bors/handlers/workflow.rs
Kobzol: It does not need to return an error. The fact that a workflow has failed is not an error, that's normal operation :) We store that information into the DB and continue executing.

sriganeshres:
@Kobzol I have made the changes can you review them and can you give me a idea of tests to write regarding this.

Kobzol (commented):
I implemented a simple test, that also changed internal logic for determining if a check suite was failed or not.

In the future, we should also generalize the workflow status to have explicit information about running for too short, as the error message in this case is confusing. But let's not do that in this PR.

src/config.rs
```diff
@@ -21,17 +21,26 @@ pub struct RepositoryConfig {
     pub labels: HashMap<LabelTrigger, Vec<LabelModification>>,
     #[serde(
         default = "default_min_ci_duration",
```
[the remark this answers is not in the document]
sriganeshres: Done

Kobzol:
Thank you! :)

PR #166, opened 2024-10-18
Include repo name in merge commit

### LINKED ISSUE #99: Include repository name in commit message
Homu issue: https://github.com/rust-lang/homu/issues/185

NULLx76:
This should fix #99 

I did notice that in the  `tests/data/*` files the commit name is also saved, so if those files are used as test validation, I expect them to fail. I wasn't entirely sure how to run the tests, so I've left those files as they are for now.


Kobzol (approved):
Looks great, thanks! :)

PR #174, opened 2024-11-15
chore: update octocrab to version 0.42

MarcoIeni:
EDIT: tests are failing. the regression is introduced in 0.39


Kobzol:
I'm quite sure that updating `octocrab` breaks bors (it does compile, but some behavior changes). We'd need to change code to make it work.

Kobzol (approved):
Tests pass, so LGTM :)

PR #180, opened 2024-11-22
refactor: clean up RepositoryLoader

vohoanglong0107:
## Why

Previously RepositoryLoader was created to ease the migration from trait to sqlx. Now that the migration is complete, we can safely clean it up.

### DIFF
--- a/src/github/api/mod.rs
+++ b/src/github/api/mod.rs
@@ -10,6 +10,7 @@ use secrecy::{ExposeSecret, SecretString};
 use client::GithubRepositoryClient;
 
 use crate::bors::RepositoryState;
+use crate::config::RepositoryConfig;
 use crate::github::GithubRepoName;
 use crate::permissions::TeamApiClient;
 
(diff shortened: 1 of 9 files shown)


src/github/api/mod.rs
```diff
@@ -83,17 +84,21 @@ pub async fn load_repositories(
...
                 team_api_client,
                 repo.clone(),
                 name.clone(),
             )
             .await
-            .map_err(|error| {
-                anyhow::anyhow!("Cannot load repository {:?}: {error:?}", repo.full_name)
-            });
+            {
+                Ok(repo_state) => repo_state,
```
Kobzol, line 96: The complicated handling of results here was by design, because we wanted to have a hard error when the repos fail to load when the bot starts, but only log a warning when we reload the state during the bot's execution. By moving the error handling here, it is not possible to distinguish these two cases. So I would prefer keeping the previous approach of returning `Result<RepositoryState>`.

Kobzol (commented):
Left one comment, otherwise the cleanup looks nice.

vohoanglong0107:
Oh that makes perfect sense. You want me to document this into the code, to
prevent future "clean up" attempts? 😆

Kobzol:
Yes please :)

Kobzol (approved):
Thanks!

PR #181, opened 2024-11-27
chore(ci): remove deprecated actions-rs actions

MarcoIeni:
Here's why I chose `actions-rust-lang/setup-rust-toolchain`:
- automatically setup `Swatinem/rust-cache@v2`.
- respects `rust-toolchain` file in case we add it in the future
- it has a [problem matcher](https://github.com/actions-rust-lang/setup-rust-toolchain/blob/main/rust.json)
- it has the `toolchain` field, which decouples the version of the action from the rust toolchain version. I.e. I think `action/action@stable` isn't great, because you can't pin the action commit.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,3 @@
-## Based on https://github.com/actions-rs/meta/blob/master/recipes/quickstart.md
 name: Test
 
 on:
@@ -30,14 +29,10 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install stable toolchain
-        uses: actions-rs/toolchain@v1
+        uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          profile: minimal
           toolchain: 1.80.0
-          override: true
           components: clippy, rustfmt
-
-      - uses: Swatinem/rust-cache@v2
 
       - name: Install sqlx-cli
         run: cargo install sqlx-cli@0.7.4 --no-default-features --features native-tls,postgres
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -30,15 +29,11 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install stable toolchain
-        uses: actions-rs/toolchain@v1
+        uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          profile: minimal
           toolchain: 1.80.0
-          override: true
```
MarcoIeni, line 37: override true is default.
profile is not present as an option.

.github/workflows/test.yml
```diff
@@ -49,28 +44,17 @@ jobs:
         run: cargo sqlx prepare --check -- --tests
 
       - name: Build
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: --all --all-targets
```
MarcoIeni, line 55: `--all` is deprecated in favor of `--workspace`

.github/workflows/test.yml
```diff
@@ -49,28 +44,17 @@ jobs:
...
-        with:
-          command: clippy
-          args: --all
+        run: cargo clippy --workspace
 
       - name: Check formatting
-        uses: actions-rs/cargo@v1
-        with:
-          command: fmt
-          args: --all -- --check
```
MarcoIeni, line 73: the `--` is not required anymore

MarcoIeni:
I got an error from CI.
I can reproduce with `cargo +1.80.0 install sqlx-cli@0.7.4 --no-default-features --features native-tls,postgres --force`

but I can't with rust 1.82.0 so I think the previous action wasn't working properly.

MarcoIeni:
Another failure not detected (unused functions). I don't want to know what the old action was doing 🙈

Kobzol (approved):
Thanks! :)

PR #186, opened 2024-12-08
feat: add github client dedicated for ci

vohoanglong0107:
# What

This PR adds a new github client to interact with CI repo.

# Why

The CI repo for Rust belongs to a different organization than the original repo, since we can't create a fork of a repo inside the same organization. However, Github doesn't support cross-orgs operations with Github App. We have to use another mechanism to authenticate with the CI repo.

# How

This PR introduces a new Github client, authenticated with Personal Access Token. This is the same method used by homu ([ref](https://github.com/rust-lang/homu/blob/master/cfg.production.toml#L7))

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,10 +53,10 @@ sqlx = { version = "0.8.1", features = ["runtime-tokio", "tls-rustls", "postgres
 chrono = "0.4"
 
 itertools = "0.13.0"
+derive_builder = "0.20.0"
 
 [dev-dependencies]
 insta = "1.26"
-derive_builder = "0.20.0"
 wiremock = "0.6.0"
 base64 = "0.22.1"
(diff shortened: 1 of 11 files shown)


Kobzol:
Oh, crap, that is unfortunate. I'll discuss this with the infra team.

Skgland:
> The CI repo for Rust belongs to a different organization than the original repo, since we can't create a fork of a repo inside the same organization.

Is that still the case? Based on https://github.com/github/roadmap/issues/330 and https://github.blog/changelog/2022-06-27-improved-innersource-collaboration-and-enterprise-fork-policies/ I would have assumed this is now possible.

vohoanglong0107:
I'm not sure. Need someone with admin permission in rust-lang org to test this 😅

PR #187, opened 2024-12-08
refactor: centralize build and workflow comments

vohoanglong0107:
# What

Centralize build and workflow comments.

Also refactor try build handler to multiple steps, to make it easier to see which steps use the Github app client, which step needs the CI client.


Kobzol (approved):
Thank you!

PR #195, opened 2025-01-07
chore: update to axum 0.8

MarcoIeni:
Changes: https://github.com/tokio-rs/axum/releases/tag/axum-v0.8.0

This is the biggest breaking change:

> breaking: Upgrade matchit to 0.8, changing the path parameter syntax from /:single and /*many to /{single} and /{*many}; the old syntax produces a panic to avoid silent change in behavior (https://github.com/tokio-rs/axum/pull/2645)

but I haven't seen this used in the codebase


Kobzol (approved):
Thanks!

PR #200, opened 2025-02-21
feat: added check for invalid username and if it is invalid find related names

### LINKED ISSUE #191: Check that reviewers actually exist
When someone does `@bors r=abc`, the bot should check if `abc` is a valid (GitHub) username (or a GH team?). To avoid situations where the name of the reviewer is mispelled, which happens quite often, and is problematic for https://thanks.rust-lang.org/.

It is common to specify names separated by commas, so bors should understand that and check all usernames separately.

We might also include some kind of opt-out, where you really want to use `r=` for a non-existent GH account.

Gmin2:
Fixes #191

### DIFF
--- a/src/github/api/client.rs
+++ b/src/github/api/client.rs
@@ -1,6 +1,7 @@
 use anyhow::Context;
 use octocrab::models::{App, Repository};
 use octocrab::{Error, Octocrab};
+use strsim::levenshtein;
 use tracing::log;
 
 use crate::bors::event::PullRequestComment;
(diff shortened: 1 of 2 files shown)


src/github/api/client.rs
```diff
@@ -1,6 +1,7 @@
 use anyhow::Context;
 use octocrab::models::{App, Repository};
 use octocrab::{Error, Octocrab};
+use strsim::levenshtein;
```
Kobzol, line 4: The `strsim` dependency is missing.

src/github/api/client.rs
```diff
@@ -287,6 +288,72 @@ impl GithubRepositoryClient {
...
+            }
+        }
+        Ok(None)
+    }
+
+    /// Searches for GitHub usernames similar to the provided username using Levenshtein distance.
+    async fn find_similar_usernames(&self, username: &str) -> anyhow::Result<Vec<String>> {
+        const MAX_DISTANCE: usize = 2;
+
+        let search_results = self
```
Kobzol, line 336: I wouldn't use the GitHub search API for this, that could be slow, and unknown GH users are not relevant anyway. Instead, we should try to search for known users in the team database. It will also be easier to test it like this.

But we can also skip the "find similar" functionality in this PR, if you want.

Kobzol (commented):
Thank you for the PR! It looks pretty good already. Left some comments.

src/github/api/client.rs
Gmin2: oh shit, forgotten to push it, done now !

src/github/api/client.rs
Gmin2: > I wouldn't use the GitHub search API for this, that could be slow, and unknown GH users are not relevant anyway. Instead, we should try to search for known users in the team database. It will also be easier to test it like this.

so we should use [this](https://github.com/rust-lang/bors/blob/417c8e7e3b0f9a2d6cc98b99d9bbb4ba9789ee7c/src/permissions.rs#L51) right ?

Kobzol: Yes, the reviewer list from the team API.

Kobzol:
Hi, are you still interested in finishing this PR? :)

Kobzol:
This was implemented in https://github.com/rust-lang/bors/pull/675. Thank you for the initial work on this!

PR #203, opened 2025-02-26
fix: ignore unrelated comments

### LINKED ISSUE #124: Do not post a comment for unrelated errors
After every comment on a PR, the bot tries to load the PR information from GitHub. However, this can fail, and in that case the bot starts spamming the PR with [error messages](https://github.com/rust-lang/rust/pull/128861#issuecomment-2286217324). The bot should post a comment when it encounters an error during its operation, but not when it just deals with a comment that is completely unrelated to it. We should first find out whether the comment actually contains a bors command, and if not, we should just always ignore it.

I originally included the PR info loading logic unconditionally to reduce code duplication (so that we always load the PR regardless of the event), but it was a bad idea. We are also needlessly spamming GitHub now, even though most comments are unrelated to bors.

Sakib25800:
Fixes #124

### DIFF
--- a/src/bors/handlers/mod.rs
+++ b/src/bors/handlers/mod.rs
@@ -185,6 +185,11 @@ async fn handle_comment(
     let pr_number = comment.pr_number;
     let commands = ctx.parser.parse_commands(&comment.text);
 
+    // Bail if no commands
+    if commands.is_empty() {
+        return Ok(());
+    }
+
     tracing::debug!("Commands: {commands:?}");
     tracing::trace!("Text: {}", comment.text);
 
@@ -320,6 +325,16 @@ mod tests {
                 .post_comment(Comment::from("@bors ping").with_author(User::bors_bot()))
                 .await?;
             // Returning here will make sure that no comments were received
+            Ok(tester)
+        })
+        .await;
+    }
+
+    #[sqlx::test]
+    async fn do_not_load_pr_on_unrelated_comment(pool: sqlx::PgPool) {
+        run_test(pool, |mut tester| async {
+            tester.default_repo().lock().pull_request_error = true;
+            tester.post_comment("no command").await?;
             Ok(tester)
         })
         .await;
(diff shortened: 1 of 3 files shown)


src/tests/mocks/repository.rs
```diff
@@ -58,6 +58,7 @@ pub struct Repo {
     pub cancelled_workflows: Vec<u64>,
     pub workflow_cancel_error: bool,
     pub pull_requests: HashMap<u64, PullRequest>,
+    pub pull_request_error: bool,
```
Kobzol, line 61: Could you please add a comment here to explain what the field does? Thanks!

src/bors/handlers/mod.rs
```diff
@@ -324,4 +329,15 @@ mod tests {
         })
         .await;
     }
+
+    #[tracing_test::traced_test]
```
Kobzol, line 333:
```suggestion

```
This is mostly for local debugging, we don't need to commit it.

src/bors/handlers/mod.rs
```diff
@@ -324,4 +329,15 @@ mod tests {
...
+    #[sqlx::test]
+    async fn pr_fetch_error(pool: sqlx::PgPool) {
```
Kobzol, line 335:
```suggestion
    async fn do_not_load_pr_on_unrelated_comment(pool: sqlx::PgPool) {
```
Something like this would make it easier to understand what's the test designed for.

Kobzol (commented):
I checked the code again and it seems like we didn't spam the repo with error comments anymore :sweat_smile: We would produce an error in the log though, and mainly we would spam GitHub, both of which is not needed, so this PR is still a nice improvement.

Thank you! Left a few nits.

Kobzol:
Thanks! Please squash the changes and we're good to go. Looks like I'll also need to fix CI.

Sakib25800:
This is part of what using bors would have solved right? 😆

Kobzol (approved):

PR #205, opened 2025-02-27
feat: add priority option

Sakib25800:
This PR adds a priority option to `@bors r=<username> <p=priority>` and `@bors r+ <p=priority>`.


Sakib25800:
Prerequisite to #202

src/database/operations.rs
```diff
@@ -88,7 +93,7 @@ pub(crate) async fn unapprove_pull_request(
     pr_id: i32,
 ) -> anyhow::Result<()> {
     sqlx::query!(
-        "UPDATE pull_request SET approved_by = NULL WHERE id = $1",
+        "UPDATE pull_request SET approved_by = NULL, priority = NULL WHERE id = $1",
```
Kobzol, line 96: This should not modify the priority.

src/database/operations.rs
```diff
@@ -72,11 +73,15 @@ pub(crate) async fn approve_pull_request(
     executor: impl PgExecutor<'_>,
     pr_id: i32,
     approver: &str,
+    priority: Option<u32>,
 ) -> anyhow::Result<()> {
+    let priority_i32 = priority.map(|p| p as i32);
+
     sqlx::query!(
-        "UPDATE pull_request SET approved_by = $1 WHERE id = $2",
+        "UPDATE pull_request SET approved_by = $1, priority = $2 WHERE id = $3",
```
Kobzol, line 81: Note that priority can be also configured outside of approving, via `@bors p=5` (not yet implemented).

It's not possible to remove priority, but when it's unset during `r+`, the existing priority should not be modified (as it could have been set before already, see my other comment).

I would suggest to use something like `priority = COALESCE($2, priority)`, to override priority if it was set, but keep the original priority when the person just did `@bors r+`.

src/bors/command/parser.rs
```diff
@@ -118,26 +118,36 @@ fn parse_parts(input: &str) -> Result<Vec<CommandPart>, CommandParseError> {
...
     if let Some(CommandPart::KeyValue { key, value }) = parts.first() {
         if *key != "r" {
             Some(Err(CommandParseError::UnknownArg(key)))
         } else if value.is_empty() {
             return Some(Err(CommandParseError::MissingArgValue { arg: "r" }));
         } else {
-            Some(Ok(BorsCommand::Approve(Approver::Specified(
-                value.to_string(),
-            ))))
+            match parse_priority(parts) {
```
Kobzol, line 144: This should only pass `&parts[1..]`, to skip the already parsed `r`.

Kobzol (commented):
Looks pretty good! Left some comments. And there should also be some tests (e.g. run `@bors r+ p=...` twice in a row and similar cases).

Then we'll add more test cases once `@bors p=5` is implemented.

Sakib25800:
Thanks! I'll make those changes and @ you when ready.

src/database/operations.rs
Sakib25800: I went with `priority = COALESCE($2, priority)`.

Sakib25800:
@Kobzol do we need a `max_priority` option like Homu has?

src/bors/command/mod.rs
```diff
@@ -39,4 +47,6 @@ pub enum BorsCommand {
     },
     /// Cancel a try build.
     TryCancel,
+    /// Set the priority of a commit.
+    Priority(Priority),
```
Kobzol, line 51:
```suggestion
    SetPriority(Priority),
```

src/bors/command/parser.rs
```diff
@@ -293,40 +373,265 @@ mod tests {
...
+    }
+
+    #[test]
+    fn parse_approve_negative_priority_invalid() {
+        let cmds = parse_commands("@bors r+ p=-1");
+        assert_eq!(cmds.len(), 1);
+        insta::assert_debug_snapshot!(cmds[0], @r###"
+        Err(
+            ValidationError(
+                "Priority must be a valid number",
```
Kobzol, line 473:
```suggestion
                "Priority must be a non-negative integer",
```

src/bors/handlers/mod.rs
```diff
@@ -205,18 +206,31 @@ async fn handle_comment(
...
                     }
                     BorsCommand::Unapprove => {
                         let span = tracing::info_span!("Unapprove");
                         command_unapprove(repo, database, &pull_request, &comment.author)
                             .instrument(span)
                             .await
                     }
+                    BorsCommand::Priority(priority) => {
+                        let span = tracing::info_span!("Priority");
+                        command_priority(repo, database, &pull_request, &comment.author, priority)
```
Kobzol, line 230:
```suggestion
                        command_set_priority(repo, database, &pull_request, &comment.author, priority)
```

src/bors/handlers/review.rs
```diff
@@ -105,24 +128,10 @@ fn sufficient_approve_permission(repo: Arc<RepositoryState>, author: &GithubUser
...
-            Comment::new(format!(
-                "@{}: :key: Insufficient privileges: not in review users",
-                author.username
-            )),
-        )
-        .await
+fn sufficient_priority_permission(repo: Arc<RepositoryState>, author: &GithubUser) -> bool {
+    repo.permissions
+        .load()
+        .has_permission(author.id, PermissionType::Try)
```
Kobzol, line 134: Setting priority should require review permissions. Unless you saw this in homu?

src/bors/handlers/review.rs
```diff
@@ -488,4 +516,62 @@ approve = ["+approved"]
...
+            tester.post_comment("@bors r+ p=10").await?;
+            tester.expect_comments(1).await;
+            let pr = tester
+                .db()
+                .get_or_create_pull_request(
+                    &default_repo_name(),
+                    PullRequestNumber(default_pr_number()),
+                )
+                .await?;
+            assert_eq!(pr.priority, Some(10));
```
Kobzol, line 532: This will get duplicated a lot, let's create some helper function on tester to get the default PR.

Sakib25800: Abstracted to: `tester.get_default_pr().await?;`

src/bors/handlers/review.rs
Sakib25800: I've changed it to review permissions.

Yeah, I saw in homu that it requires the `AuthState.TRY` permission so I thought we were copying that.

src/bors/command/mod.rs
```diff
@@ -2,6 +2,9 @@ mod parser;
 use crate::github::CommitSha;
 pub use parser::{CommandParseError, CommandParser};
 
+/// Priority of a commit.
+pub type Priority = u32;
```
Sakib25800, line 6: I think using the `newtype` idiom seems better, so I'm going to change this.

Kobzol: Using `Priority` is already much better than just `u32`. Usually I just start with a type alias, and only once I see that some errors could occur (e.g. multiple numbers being passed to a function next to each other that could be swapped), or the type is very important, I switch to newtypes.

src/bors/command/parser.rs
```diff
@@ -279,6 +279,14 @@ fn parse_priority<'a>(parts: &[CommandPart<'a>]) -> ParseResult<'a> {
             return Some(Err(CommandParseError::MissingArgValue { arg: "p" }));
         }
 
+        if let Ok(p_i64) = value.parse::<i64>() {
```
Kobzol, line 282: Sorry, I didn't mean that it has to be handled in such a complex way. I just meant that he default error message should be changed from "valid number" to "non-negative integer", just to make it a bit more precise :)

src/bors/handlers/review.rs
Kobzol: Interesting. Hmm, let's keep it as review, since I think it makes more sense, and I will create an issue to track this.

Kobzol:
> @Kobzol do we need a max_priority option like Homu has?

I don't think so, I haven't ever seen it being useful/used anywhere.

src/bors/command/parser.rs
Sakib25800: Sorry, I misunderstood. I've changed it.

src/bors/command/mod.rs
Sakib25800: Hmmzzz...I guess there isn't much point in using the `newtype` idiom in this case then.

Sakib25800:
If there's any further changes need, lmk.

Kobzol (commented):
Please also add two following tests:
- User sets priority with `@bors p=X`, and then does `@bors r+`, and check that the priority is still `X`.
- User sets priority with `@bors p=X`, and then does `@bors r+ p=Y`, and check that the priority is now `Y`.

Then I think we're done here :)

Sakib25800:
Added `priority_preserved_after_approve` and `priority_overridden_on_approve_with_priority` 🫡.

Kobzol (approved):
Thank you! This is really impressive work.

Issue #206, opened 2025-03-02
Differences between bors and homu


Kobzol:
This tracking issue enumerates (known) differences between the new bors and the original [homu](https://github.com/rust-lang/homu) repository. We can decide later whether to keep the differences or keep compatibility with homu.

- Setting priority of PRs requires `review` permissions in bors, but `try` in homu.
  - https://github.com/rust-lang/bors/pull/205

PR #210, opened 2025-03-05
feat: add delegation commands

### LINKED ISSUE #209: Implement delegation commands
[homu](https://bors.rust-lang.org/) has commands for "delegating" a PR. This means that when a reviewer does `@bors delegate+`, they provide the author of the PR with permissions to approve that PR (essentially they grant them reviewer permissions only for that single PR).

Bors can't currently do merges, but we could implement parsing of the command, checking of permissions etc., storing of the rollup state in the DB for a specific PR etc.

Sakib25800:
Fixes #209


Sakib25800:
I've modified the undelegation logic to require either reviewer permission or for the user to be a delegated user. However, the original **Homu** implementation required the try permission - seems like they never got around to it.   

```python
elif command.action == 'undelegate':
        # TODO: why is this a TRY?
        if not _try_auth_verified():
            continue
        state.delegate = ''
        state.save()
```
[Source reference](https://github.com/rust-lang/homu/blob/d9ed287e059b34661cb14e4e893e305812c42524/homu/main.py#L671)

So I guess #206 will need updating.

Kobzol (commented):
Thank you! Implementation looks good on a first glance, but I have to say that I forgot about `@bors delegate=<username>`. That sounds a bit.. useless, to be honest, and could be a potential security problem if someone mispells the name, since some other user could then get approval rights. Or a user could rename themselves to get approval rights (in theory). In general, we should try to store user GitHub IDs instead of usernames, but in this case I think we can just store a boolean property (delegated/not delegated).

I think that `delegate+` and `delegate-` should be enough. I asked others about this [here](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Use-case.20for.20bors.20delegate.3D.3Cname.3E).

Kobzol:
Yeah let's just keep `delegate+` and `delegate-`, to keep it simpler.

Sakib25800:
I've removed it. 

DB wise: `delegated_to` -> `delegated`.

src/bors/command/mod.rs
```diff
@@ -49,4 +49,8 @@ pub enum BorsCommand {
     TryCancel,
     /// Set the priority of a commit.
     SetPriority(Priority),
+    /// Delegate approval authority.
```
Kobzol, line 52:
```suggestion
    /// Delegate approval authority to the pull request author.
```

src/bors/handlers/review.rs
```diff
@@ -25,7 +25,7 @@ pub(super) async fn command_approve(
     priority: Option<u32>,
 ) -> anyhow::Result<()> {
     tracing::info!("Approving PR {}", pr.number);
-    if !sufficient_approve_permission(repo_state.clone(), author) {
+    if !has_review_or_delegated_permission(&repo_state, author, pr, &db).await? {
```
Kobzol, line 28:
```suggestion
    if !has_approve_permissions(&repo_state, author, pr, &db).await? {
```
I think that name is clearer (and doesn't leak implementation details :) ).

src/bors/handlers/review.rs
```diff
@@ -79,6 +79,39 @@ pub(super) async fn command_set_priority(
         .await
 }
 
+/// Delegate approval authority of a pull request.
```
Kobzol, line 82:
```suggestion
/// Delegate approval authority of a pull request to its author.
```

src/bors/handlers/review.rs
```diff
@@ -122,10 +155,27 @@ pub(super) async fn handle_push_to_pull_request(
...
         .load()
         .has_permission(author.id, PermissionType::Review)
+    {
+        true
+    } else {
+        let pr_model = db
+            .get_or_create_pull_request(repo_state.repository(), pr.number)
+            .await?;
+
+        pr_model.delegated
```
Kobzol, line 175: We also need to check if the author of the comment is the same as the author of the PR :sweat_smile: By comparing their GitHub IDs. And add a test for this.

src/bors/handlers/review.rs
```diff
@@ -598,4 +667,164 @@ approve = ["+approved"]
...
+        })
+        .await;
+    }
+
+    #[sqlx::test]
+    async fn delegatee_can_approve(pool: sqlx::PgPool) {
+        BorsBuilder::new(pool)
+            .world(create_world_with_approve_config())
+            .run_test(|mut tester| async {
+                let pr_author = User::default();
```
Kobzol, line 695: We should use something else than `User::default()` here, because that is already the default author of comments in tests and also the author of PRs. In general, for the delegation tests, we should mostly test a situation where the author of the PR is *not* a reviewer, and the commenter is a reviewer.

Since most (all?) existing tests only dealt with a single user so far, we might need some new test helper utilities to quickly create comments for a specific user.

src/bors/handlers/review.rs
```diff
@@ -598,4 +667,164 @@ approve = ["+approved"]
...
+    }
+
+    #[sqlx::test]
+    async fn undelegate_by_delegatee(pool: sqlx::PgPool) {
+        run_test(pool, |mut tester| async {
+            tester.post_comment("@bors delegate+").await?;
+            tester.expect_comments(1).await;
+
+            tester
+                .post_comment(Comment::from("@bors delegate-").with_author(User::default()))
```
Kobzol, line 775:
```suggestion
                .post_comment(Comment::from("@bors delegate-").with_author(User::default()))
```
Again here, `User::default()` is the author of the PR, so this whole test deals just with a single user.

Kobzol (commented):
There is kind of a glaring security hole in the current implementation :laughing: But should be easy to fix.

The tests will also require some modifications to actually test a situation with multiple users. In particular, there should be a test where person A creates a PR, person B delegates it, and then person C tries to approve, but it fails. This would find the "security hole" xD

src/bors/handlers/review.rs
Sakib25800: No wonder all the tests were passing first try - I knew it was too good to be true. The tests were literally a reviewer delegating themselves 🤦‍♂️.

Yeah, all existing tests deal with a single user so I've added utilities. 

I've gone ahead by adding a `create_world_with_delegate_config`, `reviewer()` and `as_reviewer()` utils.

src/bors/handlers/review.rs
Sakib25800: I'm a doughnut, not too sure what I was thinking in some parts of the code 😅. Sorry for the silly errors, they should be fixed.

Kobzol:
Btw, when we are implementing review remarks, it's fine to keep pushing individual commits, we can always squash at the very end just before merging. It is easier to review the changes more separately, rather than checking a big force push that changes the whole PR.

src/bors/handlers/review.rs
```diff
@@ -598,4 +646,220 @@ approve = ["+approved"]
...
+    async fn delegatee_can_try(pool: sqlx::PgPool) {
+        BorsBuilder::new(pool)
+            .world(create_world_with_delegate_config())
+            .run_test(|mut tester| async {
+                tester.post_comment(as_reviewer("@bors delegate+")).await?;
+                tester.expect_comments(1).await;
+
+                tester.post_comment("@bors try").await?;
+                tester.expect_comments(1).await;
+                tester
```
Kobzol, line 726: This actually sends an event to bors that the workflow has started (it doesn't check that it has started). Check e.g. `try_merge_branch_history` for how can we check that a try build has been requested.

src/bors/handlers/review.rs
```diff
@@ -598,4 +646,220 @@ approve = ["+approved"]
...
+    #[sqlx::test]
+    async fn undelegate_by_delegatee(pool: sqlx::PgPool) {
+        BorsBuilder::new(pool)
+            .world(create_world_with_delegate_config())
+            .run_test(|mut tester| async {
+                tester.post_comment(as_reviewer("@bors delegate+")).await?;
+                tester.expect_comments(1).await;
+
+                tester.post_comment("@bors delegate-").await?;
+                tester.expect_comments(1).await;
```
Kobzol, line 786: Undelegation doesn't send a comment, right? So what is this comment?

src/bors/handlers/trybuild.rs
```diff
@@ -190,7 +193,7 @@ pub(super) async fn command_try_cancel(
     author: &GithubUser,
 ) -> anyhow::Result<()> {
     let repo = repo.as_ref();
-    if !check_try_permissions(repo, pr, author).await? {
+    if !has_try_permission(repo, author, pr, &db).await? {
```
Kobzol, line 196: This now no longer sends a comment when the user doesn't have enoug privileges. And since CI is green, we probably didn't have a test for it :)

src/bors/handlers/trybuild.rs
```diff
@@ -322,31 +325,26 @@ handled during merge and rebase. This is normal, and you should still perform st
...
     author: &GithubUser,
+    pr: &PullRequest,
+    db: &PgDbClient,
 ) -> anyhow::Result<bool> {
-    let result = if !repo
+    if repo_state
         .permissions
         .load()
-        .has_permission(author.id, PermissionType::Try)
+        .has_permission(author.id, PermissionType::Review)
```
Kobzol, line 337: This should be `Try`, not `Review`. Clearly our test coverage of try permissions is bad :)

This is also duplicated with `has_approve_permission`. Let's just merge the two into a single function, parametrized by `PermissionType`.

Kobzol (commented):
Thank you! Left a few more comments.

src/bors/handlers/trybuild.rs
Sakib25800: I'm assuming this coverage will be improved in a separate PR, so I didn't add more tests for the permissions in this PR.

src/bors/handlers/review.rs
Sakib25800: So I was a little unsure as to what to do here. As I understand it, this allows for the test to synchronise at this point in time otherwise the database doesn't update in time and the test fails?

Kobzol: Yes, we need to synchronize somehow, usually through comments. I just wondered what is this comment? It has to print something, otherwise the test would fail. So what was it? :laughing:

Kobzol:
I think that it works now, but could you please also add a test that a (non-PR-author) user with try permissions cannot delegate? Thanks!

src/bors/handlers/review.rs
Kobzol: Hmm, something is quite wrong with the test suite. I'll take a look at it.

Kobzol: So `expect_comments` behaved in a weird way, which was hiding potential issues. We still need a way to synchronize with the service though. Maybe we could wait before the bors returns to the main event loop.

Sakib25800: I've merged the PR in and fixed the test.

Sakib25800:
Once this has been merged, I'll open a PR for the following: 
- Refactoring tests by introducing a few prebuilt users such as `DEFAULT_PR_AUTHOR`, `UNPRIVILEGED_PR_AUTHOR`, `REVIEWER`, etc and use them across the tests.
- Improve test coverage for the permissions. I guess only one test comes to mind, but I didn't want to do it under this PR.

Sakib25800:
Added `delegatee_can_set_priority` test.

Kobzol (approved):
Thank you, awesome work!

Issue #207, opened 2025-03-04
Implement tree closing and opening


Kobzol:
[homu](https://bors.rust-lang.org/) has a command closed `@bors treeclosed=<p>` and `@bors treeclosed-`, which close and open the tree, respectively.

A closed for priority P tree means that no PR with priority less than P won't be attempted to be merged.

Bors can't currently do merges, but we could implement parsing of the command, checking of permissions etc. We will also need to store this information in the DB somewhere, because it is attached to a repository, rather than to a PR, so we will need to represent repositories in the DB somehow.

It would be great to check out how [homu](https://github.com/rust-lang/homu) does this.

Rhythm1710:
Can I work on this?

Kobzol:
Sure :)

PR #211, opened 2025-03-05
Tree Closing and Opening

Rhythm1710:
Issue #207

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2518,9 +2518,9 @@ checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
 
 [[package]]
 name = "tokio"
-version = "1.43.0"
+version = "1.44.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3d61fa4ffa3de412bfea335c6ecff681de2b609ba3c77ef3e00e521813a9ed9e"
+checksum = "9975ea0f48b5aa3972bf2d888c238182458437cc2a19374b81b25cdf1023fb3a"
 dependencies = [
  "backtrace",
  "bytes",


Rhythm1710:
Issue #207

migrations/20250304172534_add_repo_model.down.sql
```diff
@@ -0,0 +1,8 @@
+-- Add down migration script here
+CREATE TABLE repo_model (
```
Kobzol, line 2: Just `repository`, to be consistent with the rest of the tables.

migrations/20250304172534_add_repo_model.down.sql
```diff
@@ -0,0 +1,8 @@
...
+    id SERIAL PRIMARY KEY,
+    repo TEXT NOT NULL,
```
Kobzol: `repo` -> `name`

migrations/20250304172534_add_repo_model.down.sql
```diff
@@ -0,0 +1,8 @@
...
+    treeclosed TEXT NOT NULL CHECK (treeclosed IN ('open', 'closed')),
```
Kobzol: Homu stored a numeric value P that states "closed for PRs below priority P". So I would store this column as `treeclosed INT NULL`, where `NULL` means "not closed" and a non-null value means a closed tree with the priority P.

src/bors/command/parser.rs
```diff
@@ -321,6 +323,30 @@ fn parse_priority_arg<'a>(parts: &[CommandPart<'a>]) -> Result<Option<u32>, Comm
     Ok(priority)
 }
 
+fn parse_tree_open<'a>(command: &'a str, _parts: &[CommandPart<'a>]) -> ParseResult<'a> {
```
Kobzol, line 331: Let's use the original [homu](https://bors.rust-lang.org/) commands, so `treeclosed-` for opening a tree.

src/bors/event.rs
```diff
@@ -18,6 +18,8 @@ pub enum BorsRepositoryEvent {
     /// A check suite has been completed, either as a workflow run on Github Actions, or as a
     /// workflow from some external CI system.
     CheckSuiteCompleted(CheckSuiteCompleted),
+    /// When a repository's tree state changes (open/closed)
+    TreeStateChanged(TreeStateChanged),
```
Kobzol, line 22: I don't think that we need any event for this? It should be enough to just handle the comment and store information about closed/opened repo in the DB.

src/bors/handlers/queue.rs
```diff
@@ -0,0 +1,192 @@
...
+            PrStatus::Empty => 3,
+            PrStatus::Error => 4,
+            PrStatus::Failure => 5,
+            PrStatus::Success => 6,
+        }
+    }
+}
+
+/// Process the queue of pull requests for each repository
+pub async fn process_queue(
```
Kobzol, line 41: This seems like it implements a basic merge functionality or something? Let's not do that in this PR :) Just store the information about the tree being closed/open in the DB and that's it.

Kobzol (commented):
Thank you! Left a bunch of comments.

Once we have the repository in the DB, we should link PRs to repositories through foreign keys, so that we can easily join them in queries (we will need to figure out if a repo for a given PR is closed or not, and get PRs belonging to a given repo). This doesn't need to happen in this PR, but probably we'll want to change the repository field of PRs to become a FK to the repository table, and then always try to insert a repo into the DB when creating new PRs.

src/bors/handlers/queue.rs
Rhythm1710: Will we need this in future? I can just comment it out and will be helpful in the future

Kobzol: I don't know how the merge functionality will look like, that will require some up-front planning, design and discussions. So let's remove this for now, it's not needed for basic tree open/close functionality.

Rhythm1710: Ok, Removing it

migrations/20250304172534_add_repository_model.down.sql
```diff
@@ -0,0 +1,8 @@
+-- Add down migration script here
+CREATE TABLE repository (
+    id SERIAL PRIMARY KEY,
+    name TEXT NOT NULL UNIQUE,
+    treeclosed INT NULL,
```
Kobzol, line 5:
```suggestion
    treeclosed BOOLEAN NULL,
```
Postgres supports booleans directly :)

src/bors/command/parser.rs
```diff
@@ -112,6 +119,20 @@ fn parse_parts(input: &str) -> Result<Vec<CommandPart>, CommandParseError> {
             break;
         }
 
+        // Special handling for r- command
+        if item == "r-" {
```
Kobzol, line 123: I would prefer to avoid special cases here. We can parse the two commands in a much simpler way, e.g.:
```rust
fn parse_tree_open<'a>(command: &'a str, _parts: &[CommandPart<'a>]) -> ParseResult<'a> {
    if command != "treeclosed-" {
        return None;
    }
    Some(Ok(BorsCommand::TreeOpen))
}

fn parse_tree_closed<'a>(command: &'a str, parts: &[CommandPart<'a>]) -> ParseResult<'a> {
    if command != "treeclosed" {
        return None;
    }

    match parts {
        [CommandPart::KeyValue { value, .. }] => match validate_priority(value) {
            Ok(priority) => Some(Ok(BorsCommand::TreeClosed(priority))),
            Err(e) => Some(Err(e)),
        },
        _ => Some(Err(CommandParseError::MissingArgValue {
            arg: "treeclosed",
        })),
    }
}
```

src/bors/handlers/help.rs
```diff
@@ -65,6 +67,12 @@ fn get_command_help(command: BorsCommand) -> String {
         BorsCommand::TryCancel => {
             "`try cancel`: Cancel a running try build"
         }
+        BorsCommand::TreeOpen => {
+            "`tree open`: Open the repository tree for merging"
```
Kobzol, line 71:
```suggestion
            "`treeclosed-`: Open the repository tree for merging"
```

src/config.rs
```diff
@@ -21,6 +21,8 @@ pub struct RepositoryConfig {
     pub labels: HashMap<LabelTrigger, Vec<LabelModification>>,
     #[serde(default, deserialize_with = "deserialize_duration_from_secs_opt")]
     pub min_ci_time: Option<Duration>,
+    #[serde(default)]
+    pub treeclosed: Option<u32>,
```
Kobzol, line 25: What does this do?

src/database/operations.rs
```diff
@@ -373,3 +375,47 @@ FROM workflow
...
+
+/// Gets the treeclosed state for a repository.
+pub(crate) async fn get_repository_treeclosed(
+    executor: impl PgExecutor<'_>,
+    repo: &GithubRepoName,
+) -> anyhow::Result<Vec<RepoModel>> {
+    let row = sqlx::query_as!(
+        RepoModel,
+        r#"
+        SELECT id, name, COALESCE(treeclosed, 0) as "treeclosed: TreeState", treeclosed_src, created_at
```
Kobzol, line 387: The `COALESCE` shouldn't be needed, since we're loading the treestate as `Option<i32>`, right? :thinking:

Kobzol (commented):
Left a few comments :)

src/database/operations.rs
Rhythm1710: Tried that, was giving some error, asked AI about it and it suggested this, Will change this.

src/config.rs
Rhythm1710: In this we were setting tree closed, changing that to tree open by default

src/database/operations.rs
Rhythm1710: Fixed this as well

src/bors/handlers/help.rs
Rhythm1710: Changed this

src/config.rs
Kobzol: This is a struct that represents a configuration of bors per each repository. Why should it mention treeclosed?

Kobzol:
You have a merge commit here. You should rebase onto `master` instead. Something like this:
```
git fetch origin
git rebase origin/main
git push --force
```

Issue #213, opened 2025-03-06
Implement `@bors info` command


Kobzol:
This command should reply with the status of the current PR, including information such as:
- Is the PR delegated?
- What is its priority?
- What is its merge status? (not implemented yet, so this is a future extension)
- Are there any try build workflows currently running on it?

Plus include a link to the bors queue (but again, this is not implemented yet, so this is a future extension).

Kobzol:
Sure :)

PR #215, opened 2025-03-06
@bors info command

Rhythm1710:
Fixes: https://github.com/rust-lang/bors/issues/213

### DIFF
--- a/src/bors/handlers/info.rs
+++ b/src/bors/handlers/info.rs
@@ -1,0 +1,172 @@
+use crate::bors::Comment;
+use crate::bors::RepositoryState;
+use crate::database::PgDbClient;
+use crate::github::PullRequest;
+use std::sync::Arc;
+
+pub(super) async fn command_info(
+    repo: Arc<RepositoryState>,
+    pr: &PullRequest,
+    db: Arc<PgDbClient>,
+) -> anyhow::Result<()> {
+    // Geting PR info from database
+    let pr_model = db
+        .get_or_create_pull_request(repo.client.repository(), pr.number)
+        .await?;
+
+    // Building the info message
+    let mut info_lines = Vec::new();
+
+    // Approval info
+    if let Some(approved_by) = pr_model.approved_by {
+        info_lines.push(format!("- **Approved by:** @{}", approved_by));
+    } else {
+        info_lines.push("- **Not Approved:**".to_string());
+    }
+
+    // Priority info
+    if let Some(priority) = pr_model.priority {
+        info_lines.push(format!("- **Priority:** {}", priority));
+    } else {
+        info_lines.push("- **Priority:** Not set".to_string());
+    }
+
+    // Build status
+    if let Some(try_build) = pr_model.try_build {
+        info_lines.push(format!("- **Try build branch:** {}", try_build.branch));
+
+        if let Ok(urls) = db.get_workflow_urls_for_build(&try_build).await {
+            info_lines.extend(
+                urls.into_iter()
+                    .map(|url| format!("- **Workflow URL:** {}", url)),
+            );
+        }
+    }
+
+    // Joining all lines
+    let info = info_lines.join("\n");
+
+    // Post the comment
+    repo.client
+        .post_comment(pr.number, Comment::new(info))
+        .await?;
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::tests::mocks::{create_world_with_approve_config, BorsBuilder, World};
+
+    #[sqlx::test]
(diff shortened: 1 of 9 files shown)


src/bors/handlers/info.rs
```diff
@@ -0,0 +1,58 @@
...
+            info_lines.push(format!("- **Approved by:** @{}", approved_by));
+        } else {
+            info_lines.push("- **Approved by:** No one".to_string());
+        }
+
+        // Priority info
+        if let Some(priority) = pr.priority {
+            info_lines.push(format!("- **Priority:** {}", priority));
+        } else {
+            info_lines.push("- **Priority:** None".to_string());
```
Kobzol, line 30:
```suggestion
            info_lines.push("- **Priority:** not set".to_string());
```

src/bors/handlers/info.rs
```diff
@@ -0,0 +1,58 @@
...
+    db: Arc<PgDbClient>,
+) -> anyhow::Result<()> {
+    // Geting PR info from database
+    let pr_model = db.get_pull_request(repo.repository(), pr.number).await?;
+
+    // Building the info message
+    let mut info_lines = Vec::new();
+
+    // Approval info
+    if let Some(pr) = pr_model {
```
Kobzol, line 19: Let's use `get_or_create_pull_request`, there's no point in posting that the PR was "not found" when it clearly exists :)

src/bors/handlers/info.rs
```diff
@@ -0,0 +1,58 @@
...
+
+        // Priority info
+        if let Some(priority) = pr.priority {
+            info_lines.push(format!("- **Priority:** {}", priority));
+        } else {
+            info_lines.push("- **Priority:** None".to_string());
+        }
+
+        // Build status
+        if let Some(try_build) = pr.try_build {
```
Kobzol, line 34: We could include some information about the try build, notably a link to a running workflow, if present in the DB (the `workflow` DB table). If there's no try build, then let's not put anything into the info message about try builds.

src/bors/handlers/info.rs
```diff
@@ -0,0 +1,58 @@
...
+            info_lines.push(format!("- **Try build branch:** {}", try_build.branch));
+        } else {
+            info_lines.push("- **Try build:** None".to_string());
+        }
+    } else {
+        info_lines.push("- **PR not found in database**".to_string());
+    }
+
+    // Merge status (placeholder for future)
+    info_lines.push("- **Merge status:** Not implemented yet".to_string());
```
Kobzol, line 44:
```suggestion

```

src/bors/handlers/info.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    // Queue link (placeholder for future)
+    info_lines.push("- **Queue link:** Not implemented yet".to_string());
```
Kobzol, line 47:
```suggestion

```

src/bors/handlers/info.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    // Joining all lines
```
Kobzol, line 49: These obvious comments (likely AI generated) are useless, let's remove them.

Kobzol (commented):
Looks fine. It would be nice to add a few tests.

src/bors/handlers/info.rs
Rhythm1710: Should we embed the workflow url in the run_id or just giving workflow_url is fine?

Rhythm1710:
If everything is fine @Kobzol, I will create a PR with clean commit history.

src/bors/handlers/info.rs
Kobzol: Printing a list of workflow URLs attached to the try build should be enough.

Kobzol:
You don't need to create a new PR, just rebase on `origin/main` and force push here :)

src/bors/handlers/info.rs
Rhythm1710: I have done this, should I change to get all builds from that build_id?

```
pub(crate) async fn get_workflow_url_for_build(
    executor: impl PgExecutor<'_>,
    build_id: i32,
) -> anyhow::Result<Option<String>> {
    let result = sqlx::query!(
        r#"
SELECT url
FROM workflow
WHERE build_id = $1
LIMIT 1
"#,
        build_id
    )
    .fetch_optional(executor)
    .await?;

    Ok(result.map(|r| r.url))
}
```

Kobzol: Usually there will only be a single workflow per build, but we should return all of them anyway.

Kobzol (approved):
Thank you, this is a good start!

PR #217, opened 2025-03-07
Improve tests regarding permissions

### LINKED ISSUE #212: Improve tests regarding permissions
Most of the existing tests create a default PR on a default repo by a default repo. That is fine for many use-cases, but it's quite opaque for things that need to check permissions. It would be nice to create a set of prebuilt users in a `World` configuration designed for checking privileges/permissions:
- Default PR author: no permissions
- Default reviewer: review permissions
- Try reviewer: try permissions
- Unprivileged user: no permissions

Also, it seems that we don't have a lot of tests regarding permissions of try builds, in particular checking that only a try (not review) permission is needed to start a try build.

Sakib25800:
Fixes #212

### DIFF
--- a/src/bors/handlers/trybuild.rs
+++ b/src/bors/handlers/trybuild.rs
@@ -334,7 +334,7 @@ mod tests {
     use crate::database::BuildStatus;
     use crate::github::CommitSha;
     use crate::tests::mocks::{
-        default_pr_number, default_repo_name, run_test, BorsBuilder, Permissions, Workflow,
+        default_pr_number, default_repo_name, run_test, BorsBuilder, Comment, User, Workflow,
         WorkflowEvent, World,
     };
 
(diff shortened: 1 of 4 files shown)


Sakib25800:
For PR #210, we could do something simple like:
```rust
fn create_world_with_delegate_config() -> World {
        let world = World::default();
        world.default_repo().lock().set_config(
            r#"
[labels]
approve = ["+approved"]
"#,
        );
        // Remove all permissions from PR author. 
        world
            .default_repo()
            .lock()
            .permissions
            .users
            .remove(&User::default());

        world
    }
```
This achieves the state we want.

src/bors/handlers/trybuild.rs
```diff
@@ -402,20 +402,47 @@ mod tests {
...
-                    "@default-user: :key: Insufficient privileges: not in try users"
-                );
-                Ok(tester)
-            })
-            .await;
+    #[sqlx::test]
+    async fn try_reviewer_insufficient_permissions(pool: sqlx::PgPool) {
+        run_test(pool, |mut tester| async {
+            tester
+                .post_comment(Comment::from("@bors try").with_author(User::reviewer()))
```
Kobzol, line 422: Try permissions always imply review. In other words, `reviewer` should have both review and try, it doesn't make sense for someone to have try, but not review. So this test currently doesn't make much sense :)

Kobzol (commented):
This is a nice clean-up, thanks. I think that the API could be still a bit nicer and more bulletproof, but that's for a larger refactoring. Also we should somehow check that the unprivileged user won't ever have any privileges in tests, but that might be tricky to verify, and it shouldn't happen by accident, so it should be fine.

Left one comment, otherwise LGTM.

Kobzol (approved):
Thanks!

src/bors/handlers/trybuild.rs
Sakib25800: I thought it made no sense, but I kinda just assumed that in tests we assume anything is possible.

PR #219, opened 2025-03-07
feat: track mergeability 

### LINKED ISSUE #218: Detect unmergeable status of PRs
Bors should track the mergeable status of PRs.

- [x] Track mergeability status: https://github.com/rust-lang/bors/pull/219
- [ ] Implement some refresh mechanism to update the status from GitHub

Sakib25800:
Related issue: #218 

This PR tracks merge ability. It adds a new column: `merge_state`. 

**Homu Implementation**
- Mergeable status is tracked via the `mergeable` column in the `pull` table
- **On PR open / reopen / edit / push**: `mergeable` is set to `None` 
- **On PR edit**: `mergeable` is set to `None`  only if base branch has changed
- **On branch push event**: 
    - All PRs targeting the given branch have `merge_state` set to `None`
    - If still null, then wait 5s 
    - If still null, then add to queue 
- **Async**: all PRs with a `mergeable` of `None` are scheduled to refetch in a background task via a queue

**Proposed Implementation**
- Mergeable status is tracked via the `merge_state` column in the `pull_request` table
- **On PR open / reopen / push**: `merge_state` is set to `unknown` 
- **On PR edit**: `merge_state` is set based on payload value
    - The reason we don't do this only if base branch changes is because the payload value can actually be something else other than `unknown`, whereas with the other events it is **always** `unknown` (due to the nature of the events)
- **On branch push event**: all PRs targeting the given branch have `merge_state` set to `unknown`
- **Async**: needs further thought -> left for another PR

### DIFF
--- a/migrations/20250315140840_add_merge_state_to_pr.up.sql
+++ b/migrations/20250315140840_add_merge_state_to_pr.up.sql
@@ -1,0 +1,2 @@
+-- Add up migration script here
+ALTER TABLE pull_request ADD COLUMN mergeable_state TEXT NOT NULL DEFAULT 'unknown';
--- a/src/bors/handlers/pr_events.rs
+++ b/src/bors/handlers/pr_events.rs
@@ -1,6 +1,7 @@
-use crate::bors::event::{PullRequestEdited, PullRequestOpened, PullRequestPushed};
+use crate::bors::event::{PullRequestEdited, PullRequestOpened, PullRequestPushed, PushToBranch};
 use crate::bors::handlers::labels::handle_label_trigger;
 use crate::bors::{Comment, RepositoryState};
+use crate::database::MergeableState;
 use crate::github::{CommitSha, LabelTrigger, PullRequestNumber};
 use crate::PgDbClient;
 use std::sync::Arc;
@@ -12,16 +13,20 @@ pub(super) async fn handle_pull_request_edited(
 ) -> anyhow::Result<()> {
     let pr = &payload.pull_request;
     let pr_number = pr.number;
+    let pr_model = db
+        .get_or_create_pull_request(
+            repo_state.repository(),
+            pr_number,
+            &pr.base.name,
+            pr.mergeable_state.clone().into(),
+        )
+        .await?;
 
     // If the base branch has changed, unapprove the PR
     let Some(_) = payload.from_base_sha else {
         return Ok(());
     };
 
-    let pr_model = db
-        .get_or_create_pull_request(repo_state.repository(), pr_number, &pr.base.name)
-        .await?;
-    db.update_pr_base_branch(&pr_model, &pr.base.name).await?;
     if !pr_model.is_approved() {
         return Ok(());
     }
@@ -37,15 +42,20 @@ pub(super) async fn handle_push_to_pull_request(
     payload: PullRequestPushed,
 ) -> anyhow::Result<()> {
     let pr = &payload.pull_request;
+    let pr_number = pr.number;
     let pr_model = db
-        .get_or_create_pull_request(repo_state.repository(), pr.number, &pr.base.name)
+        .get_or_create_pull_request(
+            repo_state.repository(),
+            pr_number,
+            &pr.base.name,
+            pr.mergeable_state.clone().into(),
+        )
         .await?;
 
     if !pr_model.is_approved() {
         return Ok(());
     }
 
-    let pr_number = pr_model.number;
     db.unapprove(&pr_model).await?;
     handle_label_trigger(&repo_state, pr_number, LabelTrigger::Unapproved).await?;
     notify_of_pushed_pr(&repo_state, pr_number, pr.head.sha.clone()).await
@@ -62,6 +72,24 @@ pub(super) async fn handle_pull_request_opened(
         &payload.pull_request.base.name,
     )
     .await
+}
+
+pub(super) async fn handle_push_to_branch(
+    repo_state: Arc<RepositoryState>,
+    db: Arc<PgDbClient>,
+    payload: PushToBranch,
+) -> anyhow::Result<()> {
+    let rows = db
+        .update_mergeable_states_by_base_branch(
+            repo_state.repository(),
+            &payload.branch,
+            MergeableState::Unknown,
+        )
+        .await?;
+
+    tracing::info!("Updated mergeable_state to `unknown` for {} PR(s)", rows);
+
+    Ok(())
 }
 
 async fn notify_of_edited_pr(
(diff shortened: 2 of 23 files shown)


Sakib25800:
@Kobzol is the goal now just to track merge ability, more specifically `mergeable_state` in the database?

Kobzol:
Yes, I think that is a good start :)

Sakib25800:
Due to this PR becoming a beast, part of it has been split - see #230.

Once that has been merged, this one will continue on from there.

src/database/mod.rs
```diff
@@ -168,6 +169,31 @@ impl<'r> sqlx::Decode<'r, sqlx::Postgres> for ApprovalStatus {
     }
 }
 
+/// State of a pull request merge.
+#[derive(Debug, PartialEq, sqlx::Type)]
+#[sqlx(type_name = "TEXT")]
+#[sqlx(rename_all = "snake_case")]
+pub enum MergeState {
```
Sakib25800, line 176: I chose an explicit unknown enum variant for `merge_state` as I found it nicer to work with rather than `None`. Will change it if you disagree though.

Kobzol: Sure, why not :) Both works.

migrations/20250315140840_add_merge_state_to_pr.up.sql
```diff
@@ -0,0 +1,2 @@
+-- Add up migration script here
+ALTER TABLE pull_request ADD COLUMN merge_state TEXT NOT NULL DEFAULT 'unknown';
```
Kobzol, line 2: I would also call it `mergeable_state`, because `merge_state` sounds like it was already merged.

src/database/mod.rs
```diff
@@ -168,6 +169,31 @@ impl<'r> sqlx::Decode<'r, sqlx::Postgres> for ApprovalStatus {
     }
 }
 
+/// State of a pull request merge.
```
Kobzol, line 172: Describes if a pull request can be merged or not.

src/database/mod.rs
```diff
@@ -168,6 +169,31 @@ impl<'r> sqlx::Decode<'r, sqlx::Postgres> for ApprovalStatus {
...
+    Unknown,
+}
+
+impl From<MergeableState> for MergeState {
+    fn from(state: MergeableState) -> Self {
+        match state {
+            MergeableState::Behind
+            | MergeableState::Blocked
+            | MergeableState::Dirty
+            | MergeableState::Draft => MergeState::HasConflicts,
```
Kobzol, line 188: I would probably consider `Draft` to be `Unknown`. Also, `Behind` just sounds like the PR isn't based on the latest `main` branch? That doesn't necessarily mean that there are conflicts. I really wish we had some better documentation of these options..

src/bors/handlers/pr_events.rs
```diff
@@ -12,15 +12,17 @@ pub(super) async fn handle_pull_request_edited(
 ) -> anyhow::Result<()> {
     let pr = &payload.pull_request;
     let pr_number = pr.number;
+    let pr_model = db
```
Kobzol, line 16: It's mostly microoptimization at this point, but this function already does a bunch of mostly useless SQL queries (especially since `get_or_create_pull_request` does up to three queries, which isn't ideal - that's on me :) ) and is unnecessarily complex because of that.

What we could do:
- Change `get_or_create_pull_request` to do an upsert instead (`INSERT ... ON CONFLICT DO UPDATE SET...)
-  Pass `get_or_create_pull_request` both the base branch and the mergeable state, so that it can be created or updated in a single go.
- Remove calls to `update_pr_merge_state` and `update_pr_base_branch` from this function.

I *think* that the mergeable state might be up-to-date when a PR is being created (and in most cases it will be mergeable anyway), but even if it wasn't, this should simplify the code a bit.

src/database/operations.rs
```diff
@@ -111,6 +111,28 @@ pub(crate) async fn update_pr_merge_state(
     Ok(())
 }
 
+pub(crate) async fn update_merge_states_by_base_branch(
+    executor: impl PgExecutor<'_>,
+    repo: &GithubRepoName,
+    base_branch: &str,
+    merge_state: MergeState,
+) -> anyhow::Result<u64> {
+    let result = sqlx::query!(
```
Kobzol, line 120: This query is a bit scary (although in practice pushes to the base branch should be rare, but still, there could be a lot of pull requests).

Let's add a FIXME comment here with a note to 1) Add a database index and 2) Filter PRs by state (it doesn't make sense to modify merge_state of already merged PRs), once we have PR state tracking.

src/tests/mocks/pull_request.rs
```diff
@@ -301,3 +301,19 @@ pub struct PullRequestChangeEvent {
...
+    pub repository: GitHubRepository,
+    #[serde(rename = "ref")]
+    pub ref_field: String,
+}
+
+impl GitHubPushEventPayload {
+    pub fn new(branch_name: &str) -> Self {
+        GitHubPushEventPayload {
+            repository: default_repo_name().into(),
+            ref_field: format!("refs/heads/{}", branch_name),
```
Kobzol, line 316: `format!("refs/heads/{branch_name}")`

Kobzol (commented):
Thank you for the PR! Left a few comments.

src/database/mod.rs
Sakib25800: `Draft` -> `Unknown`
`Behind` -> `Mergeable`

Correct, `Behind` means head ref out of date ([definitions](https://docs.github.com/en/graphql/reference/enums#mergestatestatus)).

src/bors/handlers/pr_events.rs
Sakib25800: - `get_or_create_pull_request` now takes in `base_branch` and `mergeable_state`
- `get_or_create_pull_request` now upserts, and updates `base_branch` and `mergeable_state` if the PR already exists

Yeah, it simplifies it nicely. 

Although, I do `mergeable_state: pr.mergeable_state.clone().into()` quite a few times. I was going to change `get_or_create_pull_request` to instead take in Octocrab's `MergeableState`, but I figured that DB methods should take DB model params.

Kobzol: The domain/DB model split isn't super clear in Bors right now, but I think it's also not worth to extend it further, it would just add complexity for little gain. Passing the DB model is fine, and the clones are also fine, that doesn't cost anything.

src/database/operations.rs
```diff
@@ -77,38 +78,51 @@ VALUES ($1, $2, $3) ON CONFLICT DO NOTHING
...
-    sqlx::query!(
-        "UPDATE pull_request SET mergeable_state = $1 WHERE id = $2 AND repository = $3",
-        mergeable_state as _,
-        pr_id,
-        repo as &GithubRepoName
+) -> anyhow::Result<PullRequestModel> {
+    let record = sqlx::query_as!(
+        PullRequestModel,
+        r#"
+        WITH upserted_pr AS (
```
Kobzol, line 91: Cool query! :)

Kobzol (commented):
Thank you! Please squash and we can merge it.

Sakib25800:
Squashed.

Kobzol (approved):
Thank you! I modified the PR description, as this doesn't fully fix https://github.com/rust-lang/bors/issues/218 yet.

PR #221, opened 2025-03-07
Added rollup parsing

### LINKED ISSUE #208: Implement rollup commands
[homu](https://bors.rust-lang.org/) has commands for marking a PR as being `rollup=[never/iffy/maybe/always]`.

Bors can't currently do merges, but we could implement parsing of the command, checking of permissions etc., storing of the rollup state in the DB for a specific PR etc.

It would be great to check out how [homu](https://github.com/rust-lang/homu) does this.

geetanshjuneja:
Closes #208


src/bors/command/mod.rs
```diff
@@ -31,6 +31,8 @@ pub enum BorsCommand {
         approver: Approver,
         /// Priority of the commit.
         priority: Option<Priority>,
+        // Rollup status of the commit.
+        rollup: String,
```
Kobzol, line 35: This should be `Option<RollupMode>`, to distinguish the rollup parameter being set/unset. The default value will be chosen elsewhere.

src/bors/command/mod.rs
```diff
@@ -53,4 +55,6 @@ pub enum BorsCommand {
     Delegate,
     /// Revoke any previously granted delegation.
     Undelegate,
+    /// Rollup status.
+    Rollup(String),
```
Kobzol, line 59: `rollup` is pretty much just four states (never/iffy/maybe/always), so it would be nicer to model this with an enum.

src/bors/command/parser.rs
```diff
@@ -75,13 +77,28 @@ impl CommandParser {
...
 
+                                        if key == "rollup" {
+                                            if ROLLUP_VALUES.contains(&value) {
+                                                return Some(Ok(BorsCommand::Rollup(
+                                                    value.to_string(),
+                                                )));
+                                            } else {
+                                                return Some(Err(
+                                                    CommandParseError::ValidationError(
+                                                        "rollup status can be never/iffy/maybe/always"
```
Kobzol, line 95: This could be generated from `ROLLUP_VALUES`, to ensure that it doesn't go out of sync.

Kobzol (commented):
Hi, thanks for the PR! Left a few comments. It would also be nice to have tests for the DB/command functionality.

src/bors/command/mod.rs
geetanshjuneja: Whenever we receive an `Approve` command shouldn't `rollup` field be set always? I am not able to see when `rollup` would be `none`.

geetanshjuneja:
what would be the behaviour if we set rollup mode before approving the PR?

src/bors/command/mod.rs
Kobzol: You can just say `@bors r+`, in that case there is no information about the rollup mode.

Kobzol:
Setting rollup mode and approving is independent. Setting the rollup mode just modifies the column in the DB, that's it :)

src/bors/command/mod.rs
geetanshjuneja: > You can just say `@bors r+`, in that case there is no information about the rollup mode.

but by default wouldn't the rollup mode would be `maybe` here

geetanshjuneja: Ohh I think I now get it. If in the db rollup mode is null and we get a`@bors r+` command then we need to change it to `maybe` else do nothing

Kobzol: Yes, but even if that wasn't the case, the command should faithfully represent what the user entered, not eagerly fill in default values. These should be decided later, not in the parser.

src/bors/handlers/mod.rs
```diff
@@ -285,9 +285,15 @@ async fn handle_comment(
...
                         let span = tracing::info_span!("Rollup");
-                        command_set_rollup(repo, database, &pull_request, &comment.author, &rollup)
-                            .instrument(span)
-                            .await
+                        command_set_rollup(
+                            repo,
+                            database,
+                            &pull_request,
+                            &comment.author,
+                            &rollup.to_string(),
```
Kobzol, line 293: Why is it converted to a string here? It should still be passed as an enum.

src/bors/handlers/review.rs
```diff
@@ -41,7 +42,7 @@ pub(super) async fn command_approve(
         pr.number,
         approver.as_str(),
         priority,
-        rollup.as_str(),
+        rollup.map(|r| r.to_string()).as_deref(),
```
Kobzol, line 45: It should still be passed as an enum here. The DB layer should be responsible for how it will represent the enum in the DB, it could be done via an integer, not a string, for example. The outside code shouldn't know about these implementation details :)

src/bors/command/mod.rs
```diff
@@ -22,6 +24,41 @@ pub enum Approver {
     Specified(String),
 }
 
+const ROLLUP_VALUES: [&str; 4] = ["always", "iffy", "maybe", "never"];
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum RollupMode {
```
Kobzol, line 31: We can also implement `Copy` for it, to make working with it easier.

src/database/client.rs
```diff
@@ -31,10 +31,19 @@ impl PgDbClient {
...
-        approve_pull_request(&self.pool, pr.id, approver, priority, rollup).await
+        if rollup.is_none() {
+            if pr.rollup.is_some() {
+                // don't change rollup mode in db if it is already set before
+                rollup = pr.rollup.as_deref();
+            }
+            else{
+                rollup = Some("always");
+            }
+        }
```
```rust
let rollup = rollup.or(pr.rollup).unwrap_or(RollupMode::Maybe);
```

It would be a bit cleaner to set the rollup state separately from the approval, but it would be kind of wasteful to perform two queries here instead of one, so let's keep it like this.

src/bors/handlers/review.rs
```diff
@@ -913,4 +914,202 @@ approve = ["+approved"]
...
+            assert_eq!(pr.priority, Some(10));
+            assert_eq!(pr.rollup, Some("never".to_string()));
+            assert!(pr.is_approved());
+            Ok(tester)
+        })
+        .await;
+    }
+
+    #[sqlx::test]
+    async fn approve_with_priority_rollup2(pool: sqlx::PgPool) {
```
Kobzol, line 972: I don't think that we need to test all combinations here, just one priority + rollup is enough, IMO.

src/bors/handlers/review.rs
```diff
@@ -913,4 +914,202 @@ approve = ["+approved"]
...
+            assert_eq!(pr.rollup, Some("always".to_string()));
+            assert_eq!(pr.priority, Some(10));
+            assert_eq!(pr.approved_by, Some("user1".to_string()));
+            Ok(tester)
+        })
+        .await;
+    }
+
+    #[sqlx::test]
+    async fn approve_on_behalf_with_priority_rollup2(pool: sqlx::PgPool) {
```
Kobzol, line 1028: What's missing though is a test that just does `@bors r+`; we should check that the default rollup mode is "maybe" in that case.

src/bors/handlers/review.rs
```diff
@@ -913,4 +914,202 @@ approve = ["+approved"]
...
+            assert_eq!(pr.rollup, Some("maybe".to_string()));
+            assert_eq!(pr.priority, Some(10));
+            assert_eq!(pr.approved_by, Some("user1".to_string()));
+            Ok(tester)
+        })
+        .await;
+    }
+
+    #[sqlx::test]
+    async fn set_rollup1(pool: sqlx::PgPool) {
```
Kobzol, line 1044: `set_rollup1` -> `set_rollup_default` to keep the test names a bit more descriptive. Also elsewhere.

Kobzol (commented):
Thanks for the changes! The conversion of the enum to a string should happen later, and the tests could use a bit of tweaking, but otherwise it looks to be pretty close to a working version.

src/bors/handlers/review.rs
geetanshjuneja: So should I change `ROLLUP ` column schema to int? It would be better to represent it in int.

Kobzol: No we can keep it as a string, that doesn't really matter :) I just wanted to state that this is an implementation detail of the DB, the rest of the code shouldn't know about it. We send the domain objects to the DB, and the DB returns domain objects back.

src/database/client.rs
```diff
@@ -31,9 +34,16 @@ impl PgDbClient {
         pr_number: PullRequestNumber,
         approver: &str,
         priority: Option<u32>,
+        rollup: Option<RollupMode>,
     ) -> anyhow::Result<()> {
         let pr = self.get_or_create_pull_request(repo, pr_number).await?;
-        approve_pull_request(&self.pool, pr.id, approver, priority).await
+        let rollup = rollup
+            .or(pr
+                .rollup
```
Kobzol, line 42: `pr.rollup` should also be of type `RollupMode`.

src/database/client.rs
```diff
@@ -31,9 +34,16 @@ impl PgDbClient {
...
+                .map(|r| RollupMode::from_str(&r).expect("Rollup mode in the db should be valid")))
+            .unwrap_or(RollupMode::Maybe);
+
+        approve_pull_request(&self.pool, pr.id, approver, priority, &rollup.to_string()).await
```
Kobzol, line 46: And this function should also receive `RollupMode`, not a string. The DB decides how is the value stored in the DB, outside code shouldn't know it.

src/bors/command/mod.rs
```diff
@@ -27,7 +27,8 @@ pub enum Approver {
 
 const ROLLUP_VALUES: [&str; 4] = ["always", "iffy", "maybe", "never"];
 
-#[derive(Copy, Clone, Debug, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq, sqlx::Type)]
+#[sqlx(type_name = "TEXT", rename_all = "lowercase")]
```
Kobzol, line 31: Please implement the `sqlx` conversion logic in the `src/database` module. The command module should have no relation to the database, otherwise changes here (e.g. just renaming the enum variant) would change how the data is read/stored from the DB, which is bad.

geetanshjuneja: done. Added impls for rollup mode in db layer

Kobzol (approved):
Thanks! :) Please squash the commits now and we can merge it.

Kobzol (approved):
Thank you, great work!

PR #223, opened 2025-03-08
Tree Closing and Opening

Rhythm1710:
Fixes: https://github.com/rust-lang/bors/issues/207

### DIFF
--- a/src/bors/handlers/review.rs
+++ b/src/bors/handlers/review.rs
@@ -7,6 +7,7 @@ use crate::bors::handlers::has_permission;
 use crate::bors::handlers::labels::handle_label_trigger;
 use crate::bors::Comment;
 use crate::bors::RepositoryState;
+use crate::database::TreeState;
 use crate::github::GithubUser;
 use crate::github::LabelTrigger;
 use crate::github::PullRequest;
@@ -129,6 +130,78 @@ pub(super) async fn command_set_rollup(
     db.set_rollup(repo_state.repository(), pr.number, rollup)
         .await
 }
+pub(super) async fn command_close_tree(
+    repo_state: Arc<RepositoryState>,
+    db: Arc<PgDbClient>,
+    pr: &PullRequest,
+    author: &GithubUser,
+    priority: u32,
+) -> anyhow::Result<()> {
+    if !sufficient_approve_permission(repo_state.clone(), author) {
+        deny_request(&repo_state, pr, author, PermissionType::Review).await?;
+        return Ok(());
+    };
+    db.upsert_repository(
+        repo_state.repository(),
+        TreeState::Closed {
+            priority,
+            source: format!(
+                "https://github.com/{}/pull/{}",
+                repo_state.repository(),
+                pr.number
+            ),
+        },
+    )
+    .await?;
+    notify_of_tree_closed(&repo_state, pr, priority).await
+}
+
+pub(super) async fn command_open_tree(
+    repo_state: Arc<RepositoryState>,
+    db: Arc<PgDbClient>,
+    pr: &PullRequest,
+    author: &GithubUser,
+) -> anyhow::Result<()> {
+    if !sufficient_delegate_permission(repo_state.clone(), author) {
+        deny_request(&repo_state, pr, author, PermissionType::Review).await?;
+        return Ok(());
+    }
+
+    db.upsert_repository(repo_state.repository(), TreeState::Open)
+        .await?;
+    notify_of_tree_open(&repo_state, pr).await
+}
+
+async fn notify_of_tree_closed(
+    repo: &RepositoryState,
+    pr: &PullRequest,
+    priority: u32,
+) -> anyhow::Result<()> {
+    repo.client
+        .post_comment(
+            pr.number,
+            Comment::new(format!(
+                "Tree closed for PRs with priority less than {}",
+                priority
+            )),
+        )
+        .await
+}
+
+async fn notify_of_tree_open(repo: &RepositoryState, pr: &PullRequest) -> anyhow::Result<()> {
+    repo.client
+        .post_comment(
+            pr.number,
+            Comment::new("Tree is now open for merging".to_string()),
+        )
+        .await
+}
+
+fn sufficient_approve_permission(repo: Arc<RepositoryState>, author: &GithubUser) -> bool {
+    repo.permissions
+        .load()
+        .has_permission(author.id, PermissionType::Review)
+}
 
 fn sufficient_delegate_permission(repo: Arc<RepositoryState>, author: &GithubUser) -> bool {
     repo.permissions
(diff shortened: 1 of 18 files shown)


src/bors/handlers/queue.rs
```diff
@@ -0,0 +1,58 @@
...
+use crate::database::{PgDbClient, PullRequestModel, TreeState};
+use crate::github::GithubRepoName;
+use anyhow::Result;
+use std::sync::Arc;
+
+use tracing::debug;
+
+/// Process the queue of pull requests for each repository
+#[allow(dead_code)]
+pub async fn process_queue(
```
Kobzol, line 11: This file is still present in the PR, it should be removed.

src/bors/handlers/review.rs
```diff
@@ -486,12 +549,17 @@ PR will need to be re-approved."#,
...
         let world = World::default();
-        world.default_repo().lock().set_config(
+        let repo = world.default_repo();
+        repo.lock().set_config(
             r#"
 [labels]
 approve = ["+approved"]
 "#,
         );
+        repo.lock()
```
Kobzol, line 559: Why was this added here?

src/bors/handlers/review.rs
```diff
@@ -610,6 +678,61 @@ approve = ["+approved"]
...
+        let world = create_world_with_approve_config();
+        BorsBuilder::new(pool)
+            .world(world)
+            .run_test(|mut tester| async {
+                // First ensure the tree is open
+                let repo_models = tester
+                    .db()
+                    .get_repository_treeclosed(&default_repo_name())
+                    .await?;
+                if repo_models.is_empty() {
```
Kobzol, line 692: The test shouldn't modify the DB in this way, just read from it.

src/bors/handlers/review.rs
```diff
@@ -156,6 +156,68 @@ pub(super) async fn handle_push_to_pull_request(
...
+    db: Arc<PgDbClient>,
+    pr: &PullRequest,
+    author: &GithubUser,
+) -> anyhow::Result<()> {
+    if !sufficient_delegate_permission(repo_state.clone(), author) {
+        deny_request(&repo_state, pr, author, PermissionType::Review).await?;
+        return Ok(());
+    }
+
+    db.update_repository_treeclosed(repo_state.repository(), 0)
```
Kobzol, line 186: Let's create two functions, `close_tree` and `open_tree`, on the DB. Let the DB layer decide how it will store the representation of closed trees.

src/database/mod.rs
```diff
@@ -83,8 +83,44 @@ impl sqlx::Type<sqlx::Postgres> for PullRequestNumber {
...
+    fn decode(value: sqlx::postgres::PgValueRef<'_>) -> Result<Self, sqlx::error::BoxDynError> {
+        let priority = <Option<i32> as sqlx::Decode<sqlx::Postgres>>::decode(value)?;
+        match priority {
+            None => Ok(TreeState::Open),
+            Some(p) => Ok(TreeState::Closed(p as u32)),
+        }
+    }
+}
+
+impl From<Option<TreeState>> for TreeState {
```
Kobzol, line 116: This is a pretty weird impl, it's very unobvious if someone uses `.into()` to go from `Option<TreeState>` to `TreeState` and it magically selects this default through this impl. This should be set explicitly somewhere else.

src/database/mod.rs
```diff
@@ -101,7 +137,7 @@ pub enum BuildStatus {
 }
 
 /// Represents a single (merged) commit.
-#[derive(Debug, sqlx::Type)]
+#[derive(Debug, Clone, sqlx::Type)]
```
Kobzol, line 140: What are the clones needed for?

Kobzol (commented):
Thanks for the changes. Left a few more comments.

I wonder if we should pre-fill the DB with known repositories when bors starts, because it does know the list of repositories at the beginning. Then we wouldn't have to assume everywhere in code that the repository might be missing. Although there are some edge cases around a repository being added to bors while it is running (although we will ~never hit that in practice). Anyway, that can be done in a separate PR.

src/bors/handlers/review.rs
Rhythm1710: I think this was added when I was resolving conflicts. I will remove this.

src/bors/handlers/queue.rs
Rhythm1710: I will remove this.

Rhythm1710:
> Thanks for the changes. Left a few more comments.
> 
> I wonder if we should pre-fill the DB with known repositories when bors starts, because it does know the list of repositories at the beginning. Then we wouldn't have to assume everywhere in code that the repository might be missing. Although there are some edge cases around a repository being added to bors while it is running (although we will ~never hit that in practice). Anyway, that can be done in a separate PR.

> I think that is a better approach. If you want I can add this in this PR only .

src/database/mod.rs
Rhythm1710: I was getting error in the try_build of Pull_Request_Model for not using clone in Build_Model.

src/bors/handlers/review.rs
Rhythm1710: Should we notify about tree open and tree closed?

src/database/mod.rs
Kobzol: Sure, but then why `PullRequestModel` needs to be clone now? :D

src/bors/handlers/review.rs
Rhythm1710: Should we notify about tree open and tree closed in this layer only?

Kobzol: I'm not sure what do you mean by that. The notification is implemented in the command handler, which is where it should be, IMO.

Kobzol:
I rebased the PR and refactored some parts a bit. I left one TODO for you to resolve, to fill in the source of the tree closed message. It should be a URL pointing to the PR comment that closed the tree.

Rhythm1710:
> I rebased the PR and refactored some parts a bit. I left one TODO for you to resolve, to fill in the source of the tree closed message. It should be a URL pointing to the PR comment that closed the tree.

Implemented this using

Function:- `source: format!(
                "https://github.com/{}/pull/{}",
                repo_state.repository(),
                pr.number
            ),`

Test:-`source: format!(
                            "https://github.com/{}/pull/{}",
                            default_repo_name(),
                            default_pr_number()
                        ),`

Kobzol:
We want to have a link to the specific comment that closed the tree, not just the PR, that could have a lot of comments (we want to see the reason why the tree was closed).
The `issue_comment` webhook contains a property `comment.html_url`, which contains a HTML URL link to the comment. It should be parsed from the webhook and stored in the parsed pull request.

We can then use that as the tree closed source.

Kobzol:
Thank you!

PR #226, opened 2025-03-11
Store approved commit 

### LINKED ISSUE #224: Store approved commit
When a PR is approved, bors should remember the exact commit SHA that was approved. We should check how homu does this.

Sakib25800:
Fixes #224 

Homu's flow:
1. Reviewer approves PR 
2. Homu checks that current `head_sha` matches the commit at approval time 
3. If match, `approved_by` is set (thus marked as approved)
4. When new commits are pushed, `head_sha` is updated and `approved_by` cleared

`head_sha` essentially serves as both the approved sha as well of course being the head sha.

Bors' flow:
- Use `approved_sha` column

Also, updated commands docs because making another PR is a little long.


Kobzol:
Hmm, how so? I'm not sure if we can depend on the PR being unapproved after a push, that can be racy. We should definitely be storing the approved SHA explicitly.

Sakib25800:
> Hmm, how so? I'm not sure if we can depend on the PR being unapproved after a push, that can be racy. We should definitely be storing the approved SHA explicitly.

I updated the PR description with a little more detail on Bors' flow. I was going to say that we could instead store `head_sha`, like Bors does, which would serve two purposes. But having gone through Bors' code and looked at where they use `head_sha` I don't think we really need it.

Kobzol:
2) I don't think that we need to check anything, because in the PR comment webhook we receive information about the PR's SHA, and we should set that to be the `approved_sha`. I don't think that we necessarily need to remember `head_sha`, any push or change of base branch should unconditionally unapprove.

src/database/mod.rs
```diff
@@ -149,6 +149,7 @@ pub struct PullRequestModel {
     pub repository: GithubRepoName,
     pub number: PullRequestNumber,
     pub approved_by: Option<String>,
+    pub approved_sha: Option<String>,
```
Kobzol, line 152: Could we store this in a way that is more bulletproof, e.g. `Option<ApprovalStatus>`?
```rust
struct ApprovalStatus {
    approver: String,
    sha: String
}
```
The models don't necessarily need to represent 1:1 what lies in the DB, since we are using these structs as domain models anyway. See https://github.com/rust-lang/bors/pull/223/files#diff-23b4c110a535fa6df354073bab85cf1645400d08a9a202311db1d2898dc406bbR202 for an example.

src/bors/handlers/review.rs
```diff
@@ -876,4 +876,33 @@ approve = ["+approved"]
...
+            tester.push_to_pull_request(default_pr_number()).await?;
+            tester.expect_comments(1).await;
+
+            tester.post_comment("@bors r+").await?;
+            tester.expect_comments(1).await;
+
+            let pr = tester.get_default_pr().await?;
+            assert_eq!(
+                pr.approved_sha,
+                Some(format!("pr-{}-sha", default_pr_number()))
```
Kobzol, line 901: Hmm, in the current shape the SHA is the same before and after push, that's not ideal :) `push_to_pull_request` should perhaps always use a different SHA, e.g. a monotonically increasing number?

src/database/mod.rs
Sakib25800: Fixed.

I didn't even know we could that.

src/bors/handlers/review.rs
Sakib25800: Fixed. 

TBH, I thought the same but I figured `push_to_pull_request` erases the `approval_status` so still tests it properly but yeah, should be more robust.

src/bors/handlers/review.rs
```diff
@@ -33,11 +34,14 @@ pub(super) async fn command_approve(
         Approver::Myself => author.username.clone(),
         Approver::Specified(approver) => approver.clone(),
     };
+    let approval_status = ApprovalStatus {
+        approver: approver.clone(),
+        sha: pr.head.sha.to_string().as_str().to_string(),
```
Kobzol, line 39: `sha: pr.head.sha.to_string()`

src/database/operations.rs
```diff
@@ -20,42 +21,67 @@ pub(crate) async fn get_pull_request(
     repo: &GithubRepoName,
     pr_number: PullRequestNumber,
 ) -> anyhow::Result<Option<PullRequestModel>> {
-    let pull_request = sqlx::query_as!(
-        PullRequestModel,
+    let record = sqlx::query!(
```
Kobzol, line 24: Remark: it's a bit annoying that we can't use `query_as` anymore when we want to create an attribute from two SQL table field. I haven't investigated this thoroughly, but I wonder if there's a way to still use `query_as` in this case. If you're interested, it would be nice to investigate :)

src/tests/mocks/bors.rs
```diff
@@ -32,6 +34,9 @@ use crate::{
 
 use super::pull_request::{GitHubPullRequestEventPayload, PullRequestChangeEvent};
 
+/// Counter used to generate incrementing SHA values for PR pushes.
+static PR_PUSH_SHA_COUNTER: AtomicU64 = AtomicU64::new(0);
```
Kobzol: We should store this in `BorsTester`, not as a global variable. Tests run in parallel, so this could race and make tests flaky.

src/tests/mocks/bors.rs
```diff
@@ -294,6 +299,15 @@ impl BorsTester {
     }
 
     pub async fn push_to_pull_request(&mut self, pr_number: u64) -> anyhow::Result<()> {
+        let repo = self.default_repo();
+        let mut repo = repo.lock();
+        let count = PR_PUSH_SHA_COUNTER.fetch_add(1, SeqCst);
+
+        if let Some(pr) = repo.pull_requests.get_mut(&pr_number) {
```
Kobzol, line 302: Let's crash here if the PR does not exist, the test should make sure that the PR has been initialized.

Kobzol (commented):
Haha, I left you one review remark and then noticed you fixed it in the following commit :D Probably you encountered the race.

Kobzol:
I played with this locally a bit and we can simplify the SQL queries regarding the try build. It is still possible to do this:
```sql
(
        build.id,
        build.repository,
        build.branch,
        build.commit_sha,
        build.status,
        build.parent,
        build.created_at
    )
    AS "try_build: BuildModel"
```
This will reuse the `BuildModel` decode logic and you thus don't need all the unwraps. You can do this in both of the queries that load PRs.

Kobzol:
Actually, there's a way simpler way to do this, nice!

```rust
let record = sqlx::query_as!(
        PullRequestModel,
        r#"
SELECT
    pr.id,
    pr.repository as "repository: GithubRepoName",
    pr.number as "number!: i64",
    (
     pr.approved_by,
     pr.approved_sha
    ) AS "approval_status: ApprovalStatus",
    pr.priority,
    pr.rollup as "rollup: RollupMode",
    pr.delegated,
    pr.created_at as "created_at: DateTime<Utc>",
    build AS "try_build: BuildModel"
FROM pull_request as pr
LEFT JOIN build ON pr.build_id = build.id
WHERE pr.repository = $1 AND
      pr.number = $2
"#,
        repo.to_string(),
        pr_number.0 as i64
    )
    .fetch_optional(executor)
    .await?;

    Ok(record)
```
plus derive `sqlx::Type` for `ApprovalStatus`.

I will use the same approach for the rollup mode.

src/database/mod.rs
```diff
@@ -133,8 +155,11 @@ impl<'r> sqlx::Decode<'r, sqlx::Postgres> for ApprovalStatus {
             <(Option<String>, Option<String>) as sqlx::Decode<sqlx::Postgres>>::decode(value)?;
 
         match (approver, sha) {
-            (Some(approver), Some(sha)) => Ok(ApprovalStatus { approver, sha }),
-            _ => Err("Cannot deserialize ApprovalStatus, both approve and SHA is required".into()),
+            (Some(approver), Some(sha)) => {
+                Ok(ApprovalStatus::Approved(ApprovalInfo { approver, sha }))
+            }
+            (None, None) => Ok(ApprovalStatus::NotApproved),
+            _ => Err("Incosistent DB approval status".into()),
```
Kobzol, line 162: `Incosistent` -> `Inconsistent` (yeah I know it's my typo xD I meant it as a temporary string, but it's enough I guess)

src/bors/handlers/info.rs
```diff
@@ -97,6 +98,7 @@ mod tests {
             .await;
     }
 
+    #[tracing_test::traced_test]
```
Kobzol, line 101:
```suggestion

```

src/database/operations.rs
```diff
@@ -147,39 +144,32 @@ pub(crate) async fn find_pr_by_build(
...
-        )
-    END AS "try_build: BuildModel",
-    pr.created_at as "created_at: DateTime<Utc>"
+    pr.priority,
+    pr.rollup as "rollup: RollupMode",
+    pr.created_at as "created_at: DateTime<Utc>",
+    build AS "try_build: BuildModel"
 FROM pull_request as pr
-    LEFT JOIN build ON pr.build_id = build.id
+    JOIN build ON pr.build_id = build.id
```
Kobzol, line 164: Why this change? I guess that in this situation an INNER JOIN will also work, I was just wondering if you made the change intentionally.

src/tests/mocks/pull_request.rs
```diff
@@ -38,7 +38,15 @@ pub async fn mock_pull_requests(
                 if pull_request_error {
                     ResponseTemplate::new(500)
                 } else {
-                    ResponseTemplate::new(200).set_body_json(GitHubPullRequest::new(pr_number))
+                    let mut pr = GitHubPullRequest::new(pr_number);
+
+                    if let Some(repo_pr) = repo_clone.lock().pull_requests.get(&pr_number) {
+                        if !repo_pr.head_sha.is_empty() {
```
Kobzol, line 44: Let's remove this condition and set a default value for the PR `head_sha`, something obvious like `<initial-pr-sha>`.

Kobzol (commented):
Thanks, I also like it more. Left a few final nits, after them you can squash the changes and we can merge this.

src/tests/mocks/pull_request.rs
Sakib25800: Tiny bit confused about this one. We already have a default SHA from `GitHubPullRequest::new()` in the form of `"pr-{number}-sha`?

src/database/operations.rs
Sakib25800: I accidentally removed this when I was doing multi cursor edits and formatting the query a little nicer.

src/tests/mocks/pull_request.rs
Kobzol: This is actually `PullRequest`, not `GitHubPullRequest`, right? The problem is that `PullRequest` implements `Default`, and thus now the default has an empty `head_sha`, which is why you used the condition, I assume. I meant to implement `Default` for this struct manually and prefill some reasonable SHA, rather than an empty string.

We should probably get rid of `new` altogether and implement `From<PullRequest>` instead, otherwise we're sending slightly inconsistent PR data in the tests. But that's a bigger change, we can do that later.

Sakib25800: I see what you mean. 

Fixed (at least for now).

Sakib25800:
Squashed!

src/tests/mocks/pull_request.rs
```diff
@@ -38,7 +38,15 @@ pub async fn mock_pull_requests(
                 if pull_request_error {
                     ResponseTemplate::new(500)
                 } else {
-                    ResponseTemplate::new(200).set_body_json(GitHubPullRequest::new(pr_number))
+                    let mut pr = GitHubPullRequest::new(pr_number);
+
+                    if let Some(repo_pr) = repo_clone.lock().pull_requests.get(&pr_number) {
+                        // if !repo_pr.head_sha.is_empty() {
```
Kobzol, line 44: Please remove the commented code, otherwise I have no other remarks for this PR.

Kobzol (approved):
Thanks! Another great PR.

PR #230, opened 2025-03-13
Track base branch

Sakib25800:
Prerequisite to #219


src/bors/handlers/pr_events.rs
```diff
@@ -214,4 +218,24 @@ PR will need to be re-approved."#,
         })
         .await;
     }
+
+    #[sqlx::test]
+    async fn update_base_branch_on_pr_edited(pool: sqlx::PgPool) {
+        run_test(pool, |mut tester| async {
+            tester
+                .edit_pull_request_with_pr(
```
Kobzol, line 269: What is the default base branch? The test should first assert (or explicitly set) the initial state, then do an edit that  changes it to something explicit and then check that it was changed. From this test it's not obvious what is checked and what was the original state.

src/database/operations.rs
```diff
@@ -78,6 +78,21 @@ VALUES ($1, $2) ON CONFLICT DO NOTHING
     Ok(())
 }
 
+pub(crate) async fn update_pr_base(
+    executor: impl PgExecutor<'_>,
+    pr_id: i32,
```
Kobzol, line 83: This should also take the repository into account :sweat_smile:

src/tests/mocks/bors.rs
```diff
@@ -293,6 +295,25 @@ impl BorsTester {
...
+        &mut self,
+        pr_number: u64,
+        pr: GitHubPullRequest,
+    ) -> anyhow::Result<()> {
+        self.send_webhook(
+            "pull_request",
+            GitHubPullRequestEventPayload::new(
+                pr_number,
+                "edited".to_string(),
+                Some(PullRequestChangeEvent {
```
Kobzol, line 324: We should expose the new base branch in `PullRequestChangeEvent`, and then set it explicitly in this function.

src/bors/event.rs
```diff
@@ -13,6 +13,8 @@ pub enum BorsRepositoryEvent {
     PullRequestEdited(PullRequestEdited),
     /// When a pull request is opened.
     PullRequestOpened(PullRequestOpened),
+    /// When a pull request is reopened.
+    PullRequestReopened(PullRequestReopened),
```
Kobzol, line 17: Do we really need this? I don't think that you can atomically reopen a pull request and change the base branch. So I would expect that it is enough to handle `PullRequestEdited`, or not? Or does GH not send `PullRequestEdited` when a base branch of a closed PR is modified? It would be great to check this :)

Kobzol (commented):
Thank you! The core of the implementation looks solid, but I have a couple of notes/questions.

src/bors/handlers/pr_events.rs
```diff
@@ -52,6 +52,19 @@ pub(super) async fn handle_push_to_pull_request(
     notify_of_pushed_pr(&repo_state, pr_number, pr.head.sha.clone()).await
 }
 
+pub(super) async fn handle_pull_request_opened(
+    repo_state: Arc<RepositoryState>,
+    db: Arc<PgDbClient>,
+    payload: PullRequestOpened,
+) -> anyhow::Result<()> {
+    db.update_pr_base(
```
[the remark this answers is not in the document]
Sakib25800: I _think_ I understood what you meant, but not totally sure. This now uses `create_pull_request_with_base_branch` instead.

src/bors/event.rs
Sakib25800: You're right, you cannot change the base branch on closed PRs in GitHub and events still get sent. 

I wonder why Homu does this then 🤔 ([source](https://github.com/rust-lang/homu/blob/d9ed287e059b34661cb14e4e893e305812c42524/homu/server.py#L453)). I guess they just update it when they can.

Kobzol: To be honest, homu probably does a lot of obsolete stuff that we don't necessarily need to mirror. Ideally, we should implement the minimum functionality necessary to keep bors working.

Kobzol: That being said, we will probably need to track the reopened webhook, because bors should also know the state of each PR (open/closed/merged). But it shouldn't be needed for tracking the base branch, so let's keep that for another PR.

Kobzol: Opened https://github.com/rust-lang/bors/issues/231 to track this.

src/bors/handlers/pr_events.rs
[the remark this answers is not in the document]
Kobzol: It can just be named `create_pull_request` :) Later there might be more parameters passed to the function, no need to explicitly talk about the base branch.

src/bors/handlers/pr_events.rs
```diff
@@ -14,11 +14,11 @@ pub(super) async fn handle_pull_request_edited(
 ) -> anyhow::Result<()> {
     let pr = &payload.pull_request;
     let pr_number = pr.number;
-    db.update_pr_base_branch(repo_state.repository(), pr_number, &pr.base.name)
-        .await?;
 
     // If the base branch has changed, unapprove the PR
     let Some(_) = payload.from_base_sha else {
+        db.update_pr_base_branch(repo_state.repository(), pr_number, &pr.base.name)
```
Kobzol: It's the opposite :sweat_smile: The PR has changed when `from_base_sha` is `Some`, not `None`.

src/bors/handlers/pr_events.rs
```diff
@@ -290,6 +290,32 @@ PR will need to be re-approved."#,
         .await;
     }
 
+    #[sqlx::test]
+    async fn preserve_base_branch_on_pr_edited_when_base_not_edited(pool: sqlx::PgPool) {
+        run_test(pool, |mut tester| async {
+            let pr = tester.get_default_pr().await?;
+            assert_eq!(pr.base_branch, None);
```
Kobzol, line 282: The base branch should never be NULL :thinking: In fact, it should be non-nullable in the DB.

Sakib25800: Ahh, shoot. I'll fix that. 

I'll update the existing `create_pull_request` method to take in `base_branch`, so all `get_or_create_pull_request` methods will now take in `base_branch`.

src/bors/handlers/pr_events.rs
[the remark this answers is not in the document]
Sakib25800: I extended the existing `create_pull_request`.

src/bors/handlers/pr_events.rs
Sakib25800: Fixed.

Kobzol (commented):
Passing the base branch to the PR is quite annoying. I still think that it is worth it to have the base branch non-nullable though, because a PR without a base branch doesn't really make any sense.

Could you please modify the DB methods that update PRs (e.g. approve/unapprove) to receive a `PullRequestModel` as a parameter, instead of PR number, base branch etc.? This will both ensure that a PR that we're trying to edit exists in the DB, and also it will avoid us having to pass N parameters to all these methods once we start adding more non-NULL attributes to PRs.

Kobzol:
Btw before rebasing I would advise you to squash the changes, yes :) It will make the rebase much more pleasant.

Sakib25800:
Fixed.

Looks much cleaner (using `PullRequestModel` as a param).

Kobzol (approved):
Thanks!

PR #235, opened 2025-03-16
Add measurements of DB and network operations

### LINKED ISSUE #233: Trace DB queries and network requests
For debugging and profiling, it could be useful to know how long do individual DB queries and network requests take.

We should wrap each DB query and network request in some helper function call that will measure the duration and output it using `tracing::trace!`. We could use the newly stabilized async closures for implementing the helper function.

shivam37483:
# Add timing measurements for DB and network operations

## Changes made
- Implemented utility functions to measure and trace operation durations
- Added `measure_db_query()` to track database query execution times
- Added `measure_network_request()` to track network request durations
- Used `tracing::trace!` for logging with operation names and durations

## Why
This implements #233 to provide better debugging and profiling capabilities. Each DB query and network request is now logged with its duration in milliseconds using tracing's trace level.

## Testing
- Verified timing measurements are logged correctly
- Confirmed trace output includes operation names and durations

Fixes #233


Sakib25800:
I think you forgot to remove `traced_test` for your tests ;)

shivam37483:
I've updated the PR based on the feedback:

1. Removed `traced_test` from timing tests as it's no longer needed since we're using proper `tracing::trace!`

2. Verified that timing logs work correctly for:
   - Real DB queries in `database/operations.rs`
   - Real network requests in `bors/handlers/trybuild.rs`
   - Not shown for mocked tests (as expected)

The implementation now fully satisfies the requirements:
- Measures duration of DB queries and network requests
- Uses `tracing::trace!` for logging
- Uses async closures
- Shows accurate timing measurements during debugging/profiling

src/bors/handlers/trybuild.rs
```diff
@@ -61,13 +62,20 @@ pub(super) async fn command_try_build(
 
     let base_sha = match get_base_sha(&pr_model, parent) {
         Ok(Some(base_sha)) => base_sha,
-        Ok(None) => repo
-            .client
-            .get_branch_sha(&pr.base.name)
-            .await
-            .context(format!("Cannot get SHA for branch {}", pr.base.name))?,
+        Ok(None) => {
+            measure_network_request("get_branch_sha", || async {
```
Kobzol, line 66: This approach is too disruptive, we need to add `measure_network_request` to many places in the code. Let's put this into the lowest level API possible, which in this case is `GithubRepositoryClient`, to avoid polluting the codebase too much.

src/tests/util.rs
```diff
@@ -1,6 +1,15 @@
 use std::sync::atomic::{AtomicUsize, Ordering};
 use thread_local::ThreadLocal;
 use tokio::sync;
+use tracing_subscriber::{fmt, EnvFilter};
+
+// Initialize tracing for all tests
+pub fn init_tracing() {
```
Kobzol, line 7: This shouldn't be needed, as we can just use the `#[traced_test]` macro.

src/utils/timing.rs
```diff
@@ -0,0 +1,118 @@
...
+pub async fn measure_operation<T, F, Fut>(operation_name: &str, f: F) -> T
+where
+    F: FnOnce() -> Fut,
+    Fut: std::future::Future<Output = T>,
+{
+    let start = Instant::now();
+    let result = f().await;
+    let duration = start.elapsed();
+
+    trace!(
```
Kobzol, line 14: Sometimes async/network/DB operations can last for a very long time (and get stuck if there's no timeout). In thatcase it is very valuable to also see the mmoent they have started. So I would prefer that we first log the start of the operation, and after it finishes, we log its duration.

src/utils/timing.rs
```diff
@@ -0,0 +1,118 @@
...
+    result
+}
+
+// Measures the duration of a database query and logs it using tracing.
+pub async fn measure_db_query<T, F, Fut>(query_name: &str, f: F) -> T
+where
+    F: FnOnce() -> Fut,
+    Fut: std::future::Future<Output = T>,
+{
+    measure_operation(&format!("db_query:{}", query_name), f).await
```
Kobzol, line 29:
```suggestion
    measure_operation(&format!("db_query:{query_name}"), f).await
```

src/utils/timing.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// Measures the duration of a network request and logs it using tracing.
+pub async fn measure_network_request<T, F, Fut>(request_name: &str, f: F) -> T
+where
+    F: FnOnce() -> Fut,
+    Fut: std::future::Future<Output = T>,
+{
+    measure_operation(&format!("network_request:{}", request_name), f).await
```
Kobzol, line 38:
```suggestion
    measure_operation(&format!("network_request:{request_name}"), f).await
```

src/utils/timing.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+#[cfg(test)]
```
Kobzol, line 41: I like to test many things in bors, but I think that this is overkill :) Let's not add unnecessary code, this isn't worth it, IMO.

Kobzol (commented):
Hi, thanks for the PR, it looks good, I left some comments.

Aside: I feel like I'm talking to an AI, not a human being :)

shivam37483:
Thank you for your feedback! I’ve made the necessary changes. Please let me know if you’d like to suggest anything else.

Kobzol (commented):
Thanks! Left one nit, otherwise LGTM.

shivam37483:
Great! Formatted duration to 2 decimal places.

docker-compose.yaml
```diff
@@ -1,4 +1,4 @@
-version: '3.7'
+﻿version: '3.7'
```
Kobzol, line 1: What is this diff? :thinking: It looks identical. Could you please revert the changes to this file? Thanks.

Kobzol (approved):

Kobzol (approved):
Thanks!

Kobzol:
Ah, this will need a rebase.

Kobzol:
Thanks, please squash the commits now and we can merge it :)

shivam37483:
Hi @Kobzol, I’ve squashed the commits into a single commit as requested. Ready for merge! Thanks! :)

Kobzol (approved):
Thanks!

PR #241, opened 2025-03-21
 Add init_db.sh script for database initialization and migration

Abdulrahman-Elfeky:
How it works?
- Runs the existing init_db function
- Applies database migrations using sqlx-cli
- Uses environment variables for configuration [POSTGRES_USER, ..... ,SKIP_DOCKER=1]
- Ensures safe execution 
- Requires psql and sqlx-cli to be installed
- Make sure to run chmod +x scripts/init_db.sh before executing the script

### DIFF
--- a/scripts/init_db.sh
+++ b/scripts/init_db.sh
@@ -1,0 +1,58 @@
+set -x
+
+set -eo pipefail
+
+if ! [ -x "$(command -v psql)" ]; then
+  echo >&2 "Error: psql is not installed."
+  exit 1
+fi
+
+if ! [ -x "$(command -v sqlx)" ]; then
+  echo >&2 "Error: sqlx is not installed."
+  echo >&2 "Use:"
+  echo >&2 "     cargo install sqlx-cli \
+--no-default-features --features rustls,postgres"
+  echo >&2 "to install it."
+  exit 1
+fi
+
+DB_USER="${POSTGRES_USER:=bors}"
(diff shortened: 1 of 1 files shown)


Kobzol:
Hi! Could you please provide a bit more info, what's the motivation for this? Do you think that the instructions in https://github.com/rust-lang/bors/blob/main/docs/development.md are too complicated and we should have a script for initing the DB? It's kind of hard to predict what environment people have, such scripts will necessarily not work for some people, so I usually rather provide more atomic commands that people can mix & match.

The linker development guide could probably be rewritten to be clearer on what to do if you want to develop on bors though. Maybe that would be better than the script? :)

Abdulrahman-Elfeky:
U broke my heart 💔😅
I was just setting the environment to start contributing, and yeah, I found that development docs can be simpler.

Kobzol:
I think that a better approach would be to use something like `just` (https://github.com/casey/just), or a simple `xtask` (https://github.com/matklad/cargo-xtask/) script, to provide more targeted commands, such as "Reset the .sqlx directory" (I do this relatively often and currently it requires copy-pasting three Bash commands out of the `docs/development.md` file). I think that would be more useful than a single bash script that does everything in one go.

So I'm going to close this, but if you'd like to experiment with the `justfile` or something similar, I'd be happy to review that. Thank you for the PR anyway!

PR #244, opened 2025-03-22
docs: fix description for `p=<priority>` command alias

defiberrys:
I've corrected the description for the `p=<priority>` command.
The previous text mentioned it as an alias for `priority=`, but it should refer to `r+ p=<priority>`, since `p=<priority>` is a shortcut for that command.


Kobzol:
Hi! Thanks for the PR. The description is actually correct, though. `@bors p=` is indeed an alias for `@bors priority=`. These commands just modify the priority without approving the PR.

PR #245, opened 2025-03-24
Extend delegation command for specific permissions

### LINKED ISSUE #242: Allow delegating try builds only
For a bit of extra safety, we should allow delegating try builds only, e.g. with `@bors delegate try`. We could store in the delegated column something like NULL, "try", "review", and then have a function that returns the delegated permissions of a PR.

Sakib25800:
Fixes #242 

- Adds `@bors delegate=<try|review>` 
- `@bors delegate+` is still a valid command (kept so that it matches Homu)

Be sure to run `sqlx database reset` as I have modified the delegate migration in place.


Kobzol:
> Be sure to run sqlx database reset as I have modified the delegate migration in place.

Even though it might be tempting, please don't do that :) Bors is already deployed on a test repository (https://github.com/kobzol/bors-kindergarten), and overwriting migrations will break it.

src/bors/handlers/help.rs
```diff
@@ -94,15 +97,15 @@ mod tests {
         run_test(pool, |mut tester| async {
             tester.post_comment("@bors help").await?;
             insta::assert_snapshot!(tester.get_comment().await?, @r"
-            - `r+ [p=<priority>] [rollup=<never/iffy/maybe/always>]`: Approve this PR. Optionally, you can specify `<priority>`, `<rollup>`.
+            - `r+ [p=<priority>] [rollup=<never|iffy|maybe|always>]`: Approve this PR. Optionally, you can specify `<priority>`, `<rollup>`.
             - `r=<user> [p=<priority>]`: Approve this PR on behalf of `<user>`. Optionally, you can specify a `<priority>`.
             - `r-`: Unapprove this PR
             - `p=<priority>`: Set the priority of this PR
-            - `delegate+`: Delegate approval authority to the PR author
+            - `delegate=<try|review>`: Delegate permissions to the PR author
```
Kobzol, line 104: Let's also mention `delegate+` here.

src/bors/handlers/review.rs
```diff
@@ -292,18 +298,21 @@ async fn notify_of_delegation(
     repo: &RepositoryState,
     pr: &PullRequest,
     delegatee: &str,
+    delegated_permission: DelegatedPermission,
 ) -> anyhow::Result<()> {
+    let message = match delegated_permission {
+        DelegatedPermission::Try => format!("@{} can now test this pull request", delegatee),
```
Kobzol, line 304:
```suggestion
        DelegatedPermission::Try => format!("@{} can now perform try builds this pull request", delegatee),
```

diff.txt
Kobzol, line 1: Extra file committed? :)

migrations/20250324151905_rename_delegated_to_delegated_permission.up.sql
```diff
@@ -0,0 +1,23 @@
+-- Add up migration script here
+ALTER TABLE pull_request
```
Kobzol, line 2: Maybe it would be easier to create the final column (`delegated_permission TEXT`), then fill it with data and then remove the previous column? That would avoid the renaming.

src/bors/command/parser.rs
```diff
@@ -244,12 +246,35 @@ fn parser_try_cancel<'a>(command: &CommandPart<'a>, parts: &[CommandPart<'a>]) -
...
 }
 
-/// Parses "@bors delegate+" and "@bors delegate-".
-fn parser_delegation<'a>(command: &CommandPart<'a>, _parts: &[CommandPart<'a>]) -> ParseResult<'a> {
-    match command {
-        CommandPart::Bare("delegate+") => Some(Ok(BorsCommand::Delegate)),
-        CommandPart::Bare("delegate-") => Some(Ok(BorsCommand::Undelegate)),
-        _ => None,
+/// Parses the first occurrence of `delegate=<try|review>` in `parts`.
+fn parse_delegation<'a>(parts: &[CommandPart<'a>]) -> ParseResult<'a, DelegatedPermission> {
```
Kobzol, line 250: Why did you add this function? I would expect that we still only find the delegate command in `command`, not `parts`, as before. Just now we will also look for the `KeyValue` option.

src/bors/handlers/review.rs
```diff
@@ -132,8 +138,8 @@ pub(super) async fn command_delegate(
         )
         .await?;
 
-    db.delegate(&pr_model).await?;
-    notify_of_delegation(&repo_state, pr, &pr.author.username).await
+    db.delegate(&pr_model, delegated_permission.clone()).await?;
```
Kobzol, line 141: nit: `DelegatedPermission` could probably just implement `Copy`.

src/bors/handlers/review.rs
```diff
@@ -1034,4 +1052,66 @@ mod tests {
...
+        BorsBuilder::new(pool)
+            .github(GitHubState::unauthorized_pr_author())
+            .run_test(|mut tester| async {
+                tester
+                    .post_comment(review_comment("@bors delegate=try"))
+                    .await?;
+                tester.expect_comments(1).await;
+
+                tester.post_comment("@bors try").await?;
+                tester.expect_comments(1).await;
```
Kobzol, line 1088: It would be nice to check that the comment here isn't something like "you are not allowed to perform a try build" xD Better check the comment text with a snapshot.

src/bors/handlers/review.rs
```diff
@@ -1034,4 +1052,66 @@ mod tests {
...
+
+                Ok(tester)
+            })
+            .await;
+    }
+
+    #[sqlx::test]
+    async fn delegated_try_can_not_approve(pool: sqlx::PgPool) {
```
Kobzol, line 1096: The name of the test seems off here? We should test that with a try delegatedpermission you cannot approve.

Kobzol (commented):
Thank you. Left a few minor comments, but otherwise looks great.

src/bors/command/parser.rs
Sakib25800: We did not. I've removed the `parse_delegation` function.

diff.txt
Sakib25800: I generated a `diff.txt` to check my work. 

Removed.

migrations/20250324151905_rename_delegated_to_delegated_permission.up.sql
Sakib25800: I'm like 90% sure I did this approach first, but everything broke then but now it works. 

Fixed.

src/bors/handlers/review.rs
Sakib25800: Added snapshot check.

src/bors/handlers/review.rs
Sakib25800: Fixed.

src/bors/handlers/review.rs
```diff
@@ -1098,17 +1098,17 @@ mod tests {
...
                     .await?;
                 tester.expect_comments(1).await;
                 tester
                     .default_pr()
                     .await
-                    .expect_delegated(DelegatedPermission::Review);
+                    .expect_delegated(DelegatedPermission::Try);
 
                 tester.post_comment("@bors r+").await?;
                 tester.expect_comments(1).await;
```
Kobzol, line 1110: Let's also check the error message here with a snapshot.

Kobzol (commented):
Left one nit, otherwise this looks ready :)

Sakib25800:
Squashed.

src/bors/handlers/.review.rs.pending-snap
```diff
@@ -0,0 +1,2 @@
+{"run_id":"1743092074-668724000","line":1110,"new":{"module_name":"bors__bors__handlers__review__tests","snapshot_name":"delegated_try_can_not_approve","metadata":{"source":"src/bors/handlers/review.rs","assertion_line":1110,"expression":"tester.get_comment().await?"},"snapshot":"@default-user: :key: Insufficient privileges: not in review users"},"old":{"module_name":"bors__bors__handlers__review__tests","metadata":{},"snapshot":"@unprivileged-user: :key: Insufficient privileges: not in review users"}}
```
Kobzol, line 1: Committed by accident :)

Sakib25800: Sorry! Removed.

Kobzol (approved):
Thanks! Great job.

PR #246, opened 2025-03-24
Add PR repo base branch database index 

Sakib25800:
Addresses FIXME:
```rust
// FIXME:
// 1) Add a database index on (repository, base_branch)
// 2) Filter PRs by state (only update open PRs, once we have PR state tracking)
pub(crate) async fn update_mergeable_states_by_base_branch(...) { ... }
```


src/database/operations.rs
```diff
@@ -175,7 +172,9 @@ pub(crate) async fn update_mergeable_states_by_base_branch(
             r#"
             UPDATE pull_request
             SET mergeable_state = $1
-            WHERE repository = $2 AND base_branch = $3
+            WHERE repository = $2
+            AND base_branch = $3
+            AND status != 'closed'
```
Kobzol, line 177: We also want to filter out merged PRs.

Kobzol (commented):
Actually, thinking about this a bit more, an index on the base branch column might be quite useless, because this column will be **very** low cardinality, like 99.9% of PRs will have the same value (`master`). What we want to filter quickly are only the PRs that are still open. So maybe it would make more sense to have an index on the status column instead?

Could you please try to run `EXPLAIN` on the updated `update_mergeable_states_by_base_branch` query with and without an index on the status column, to examine what would be the difference in the executed plan? Thanks! We might want to switch to using equality (`= open | draft`) rather than inequality, as I'm not sure if Postgres is able to leverage the index for the inequality check.

Sakib25800:
**Index on repository and base_branch columns**:

Without index:
```bash
bors=# EXPLAIN ANALYZE
UPDATE pull_request
SET mergeable_state = 'unknown'
WHERE repository = 'some-repo-name'
AND base_branch = 'master'
AND status NOT IN ('closed', 'merged');
                                                                       QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------
 Update on pull_request  (cost=0.15..8.17 rows=0 width=0) (actual time=0.003..0.004 rows=0 loops=1)
   ->  Index Scan using pull_request_repository_number_idx on pull_request  (cost=0.15..8.17 rows=1 width=38) (actual time=0.002..0.003 rows=0 loops=1)
         Index Cond: (repository = 'some-repo-name'::text)
         Filter: ((status <> ALL ('{closed,merged}'::text[])) AND (base_branch = 'master'::text))
 Planning Time: 1.374 ms
 Execution Time: 0.031 ms
(6 rows)
```

With index:
```bash
bors=# EXPLAIN ANALYZE
UPDATE pull_request
SET mergeable_state = 'unknown'
WHERE repository = 'some-repo-name'
AND base_branch = 'master'
AND status NOT IN ('closed', 'merged');
                                                                         QUERY PLAN                                                                     
-------------------------------------------------------------------------------------------------------------------------------------------------------------
 Update on pull_request  (cost=0.15..8.17 rows=0 width=0) (actual time=0.007..0.008 rows=0 loops=1)
   ->  Index Scan using pull_request_repository_base_branch_idx on pull_request  (cost=0.15..8.17 rows=1 width=38) (actual time=0.005..0.006 rows=0 loops=1)
         Index Cond: ((repository = 'some-repo-name'::text) AND (base_branch = 'master'::text))
         Filter: (status <> ALL ('{closed,merged}'::text[]))
 Planning Time: 1.867 ms
 Execution Time: 0.068 ms
(6 rows)
```

It got slower...?
---

**Index on status column**:

Without index:
```bash
bors=# EXPLAIN ANALYZE
UPDATE pull_request
SET mergeable_state = 'unknown'
WHERE repository = 'some-repo-name'
AND base_branch = 'master'
AND status NOT IN ('closed', 'merged');
                                                                       QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------
 Update on pull_request  (cost=0.15..8.17 rows=0 width=0) (actual time=0.006..0.007 rows=0 loops=1)
   ->  Index Scan using pull_request_repository_number_idx on pull_request  (cost=0.15..8.17 rows=1 width=38) (actual time=0.005..0.005 rows=0 loops=1)
         Index Cond: (repository = 'some-repo-name'::text)
         Filter: ((status <> ALL ('{closed,merged}'::text[])) AND (base_branch = 'master'::text))
 Planning Time: 2.032 ms
 Execution Time: 0.070 ms
(6 rows)
```

With index:
```bash
bors=# EXPLAIN ANALYZE
UPDATE pull_request
SET mergeable_state = 'unknown'
WHERE repository = 'some-repo-name'
AND base_branch = 'master'
AND status NOT IN ('closed', 'merged');
                                                                       QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------
 Update on pull_request  (cost=0.15..8.17 rows=0 width=0) (actual time=0.006..0.007 rows=0 loops=1)
   ->  Index Scan using pull_request_repository_number_idx on pull_request  (cost=0.15..8.17 rows=1 width=38) (actual time=0.005..0.005 rows=0 loops=1)
         Index Cond: (repository = 'some-repo-name'::text)
         Filter: ((status <> ALL ('{closed,merged}'::text[])) AND (base_branch = 'master'::text))
 Planning Time: 1.737 ms
 Execution Time: 0.066 ms
(6 rows)
```

Sakib25800:
Actually, I need to rerun the above queries since you said to use equality.

Sakib25800:
Using equality and not inequality.

Query: 
```bash
bors=# EXPLAIN ANALYZE
UPDATE pull_request
SET mergeable_state = 'unknown'
WHERE repository = 'some-repo-name'
AND base_branch = 'master'
AND status IN ('open', 'draft');
```

**Without index**: 
```bash
 Update on pull_request  (cost=0.15..8.17 rows=0 width=0) (actual time=0.006..0.007 rows=0 loops=1)
   ->  Index Scan using pull_request_repository_number_idx on pull_request  (cost=0.15..8.17 rows=1 width=38) (actual time=0.004..0.005 rows=0 loops=1)
         Index Cond: (repository = 'some-repo-name'::text)
         Filter: ((status = ANY ('{open,draft}'::text[])) AND (base_branch = 'master'::text))
 Planning Time: 1.666 ms
 Execution Time: 0.084 ms
(6 rows)
```

**Index on repository and base_branch columns**:
```bash
 Update on pull_request  (cost=0.15..8.17 rows=0 width=0) (actual time=0.005..0.005 rows=0 loops=1)
   ->  Index Scan using pull_request_repository_base_branch_idx on pull_request  (cost=0.15..8.17 rows=1 width=38) (actual time=0.004..0.004 rows=0 loops=1)
         Index Cond: ((repository = 'some-repo-name'::text) AND (base_branch = 'master'::text))
         Filter: (status = ANY ('{open,draft}'::text[]))
 Planning Time: 1.110 ms
 Execution Time: 0.047 ms
(6 rows)
```

**Index on status column**:
```bash
 Update on pull_request  (cost=0.15..8.17 rows=0 width=0) (actual time=0.006..0.007 rows=0 loops=1)
   ->  Index Scan using pull_request_repository_number_idx on pull_request  (cost=0.15..8.17 rows=1 width=38) (actual time=0.005..0.006 rows=0 loops=1)
         Index Cond: (repository = 'some-repo-name'::text)
         Filter: ((status = ANY ('{open,draft}'::text[])) AND (base_branch = 'master'::text))
 Planning Time: 1.169 ms
 Execution Time: 0.062 ms
(6 rows)
```

Sakib25800:
Looks like you were completely right - using equality is better.

Kobzol:
Thanks for such thorough investigation! I should have clarified that I was interested in the actual query plan, rather than estimated durations or costs. The estimated durations are pretty much worthless IMO, because you would need to test this on the actual DB machine and on production data. I'm pretty sure that the ms differences that you're seeing is pretty much just noise.

What I wanted to get rid of is this: `Filter: (status = ANY ('{open,draft}'::text[]))`, because this means that we iterate over all PRs of a repository and check their status. Ideally, I would like to move the status check to `Index cond`, to make that faster.

It seems that with `IN`, the index isn't applied though. Could you please try adding an index to the status column and using something like `WHERE status = 'open' OR status = 'draft'`? If that still doesn't work, then just `status = 'open'` (in theory we could ignore draft PRs).

I'm not sure if Postgres will even combine the two indices in this way though, because it already uses an index on the `(repository, PR number)` fields.

It's quite possible that all this is just premature optimization though. To do an actual benchmark, you could fill the DB with ~130k PRs (what is currently on `rust-lang/rust`), and then mark ~700 of them as being open or draft, and see how long it takes to execute the query. If the time is minimal, then I guess that we can just avoid the index altogether.

Sakib25800:
I'm guessing you wanted me to run these queries.

It looks like replacing the IN/OR condition with a single equality pushes the status check into the index condition.

```bash
bors=#  EXPLAIN ANALYZE
  UPDATE pull_request
  SET mergeable_state = 'unknown'
  WHERE repository = 'some-repo-name'
    AND base_branch = 'master'
    AND (status = 'open' OR status = 'draft');
                                                                       QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------
 Update on pull_request  (cost=0.15..8.17 rows=0 width=0) (actual time=0.005..0.006 rows=0 loops=1)
   ->  Index Scan using pull_request_repository_number_idx on pull_request  (cost=0.15..8.17 rows=1 width=38) (actual time=0.004..0.004 rows=0 loops=1)
         Index Cond: (repository = 'some-repo-name'::text)
         Filter: ((base_branch = 'master'::text) AND ((status = 'open'::text) OR (status = 'draft'::text)))
 Planning Time: 1.159 ms
 Execution Time: 0.056 ms
(6 rows)

bors=#
```

```bash
bors=# EXPLAIN ANALYZE
UPDATE pull_request
SET mergeable_state = 'unknown'
WHERE repository = 'some-repo-name'
AND base_branch = 'master'
AND status = 'open';
                                                                 QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------------------------
 Update on pull_request  (cost=0.15..8.17 rows=0 width=0) (actual time=0.017..0.018 rows=0 loops=1)
   ->  Index Scan using pull_request_status_idx on pull_request  (cost=0.15..8.17 rows=1 width=38) (actual time=0.016..0.016 rows=0 loops=1)
         Index Cond: (status = 'open'::text)
         Filter: ((repository = 'some-repo-name'::text) AND (base_branch = 'master'::text))
 Planning Time: 0.256 ms
 Execution Time: 0.081 ms
(6 rows)

bors=#
```

Kobzol:
Hmm, that's a bit sad, to be honest =D Without having the full status check in the index I don't think that this is worth it. 

We have query timings in the logs now, so in production (once bors actually runs on rust-lang/rust), we can observe the durations, and tune the query/DB accordingly if we saw that it's too slow. I would still be interested in how long it takes Postgres to scan through ~130k PRs, if you want to dig a bit deeper.

Anyway, let's postpone the index for now. You can modify the comment in code to add our conclusion from this attempt if you want. Thank you for doing the investigation though, without it we wouldn't know how Postgres behaves.

Sakib25800:
I've actually just finished my Database Systems module at uni, so this is good SQL practice and I've already learnt a lot in this PR. TBH, it only went over relational algebra, so this (practical SQL) is actually kinda new to me. 

I wrote a [quick script](https://gist.github.com/Sakib25800/a306d7d70ad030f8a00932bb3c5bbf3a) to simulate the rust repo. 

| Approach | Execution Time | Query Plan | Note |
|----------|----------------|------------|-------|
| No index | 38.653 ms | Sequential scan | Full table scan, filtering 129,307 rows |
| Repository + base_branch index | 30.704 ms | Sequential scan | Does full table scan even with index |
| Status index | 11.578 ms | Bitmap index scan | Uses index |
| Status index (OR instead of IN) | 16.058 ms | Bitmap OR + index scan | Slightly worse than IN clause |  
| Status index (just 'open') | 9.864 ms | Bitmap index scan | Best performance |

Kobzol:
Thank you, that is really great investigation! It seems to me that the results from the table do not fully correspond to the previously posted query plans though :thinking: Perhaps Postgres chooses things differently when the table really contains some data, vs when it is almost empty. Could you please also post the query plans of the individual queries on the 130k PR table here?

> Does full table scan even with index

Maybe because it knows that the cardinality of the base branch column is very low, and `master` is set for almost all PRs anyway.

Sakib25800:
Postgres does [behave differently](https://www.pgmustard.com/blog/why-isnt-postgres-using-my-index#:~:text=There%20are%20a%20few%20(normally%20good)%20reasons%20for%20Postgres%20choosing%20a%20sequential%20scan%20even%20when%20it%20could%20use%20an%20index%20scan) when the table contains data vs. when it does not. 

Yeah, it does do a full table scan despite the index due to the low cardinality. 

I reran the results on an empty DB vs. simulated data DB, just to make sure I didn't do anything wrong the first time. Here are the [raw results](https://gist.github.com/Sakib25800/7f47c02fd20562d3a0822fec24b99f78).

Based on the data, it seems like going with a status index and checking for just `open` PRs is the most optimal approach (since we can ignore draft PRs)? 

| Approach | Empty Table Plan | 130k PRs Plan |
|----------|------------------|---------------|
| No index | Index Scan on repository_number_idx (0.059 ms) | Sequential Scan checking 130,000 rows, filtering out 129,307 (32.342 ms) |
| Repository + base_branch index | Index Scan on repository_base_branch_idx (0.030 ms) | Sequential Scan checking 130,000 rows, filtering out 129,307 (34.734 ms) |
| Status index (`IN`) | Index Scan on repository_number_idx (0.033 ms) | Bitmap Heap Scan with Bitmap Index Scan on status_idx, examining 700 candidate rows (11.389 ms) |
| Status index (`OR`) | Index Scan on repository_number_idx (0.055 ms) | Bitmap Heap Scan with BitmapOr of two Bitmap Index Scans on status_idx, examining 1,350 candidate rows (14.261 ms) |
| Status index (just `open`) | Index Scan on status_idx (0.036 ms) | Bitmap Heap Scan with Bitmap Index Scan on status_idx, examining 1,283 candidate rows (13.104 ms) |

Kobzol:
I wonder how do these numbers look like on our RDS server that has like 1 virtual CPU and 2 GiB of RAM or something xD Adding an index also has the added cost of more expensive updates, although hopefully the status column of PRs won't be updated *that often*.

I would probably go with an index on the status column then. It doesn't seem like the queries are all that different, using `IN + open/draft` seems reasonable, and we'll also handle draft PRs as a bonus (not sure if homu handles draft PRs though - could you check please? :) ).

Sakib25800:
Homu updates the `mergeable` column to `NULL` for all PRs ([source](https://github.com/rust-lang/homu/blob/d9ed287e059b34661cb14e4e893e305812c42524/homu/server.py#L569)). 

Homu does not track draft PRs at all, which I guess checks out since draft PRs were only introduced in 2019. 

I have gone with IN and open/draft.

PR #247, opened 2025-03-24
Implement refresh mechanism for unknown mergeable PRs 

### LINKED ISSUE #218: Detect unmergeable status of PRs

Sakib25800:
Fixes #218 

This PR:
- Adds mergeable queue processing on periodic refresh

### DIFF
--- a/src/bors/handlers/refresh.rs
+++ b/src/bors/handlers/refresh.rs
@@ -249,12 +249,12 @@ timeout = 3600
             tester
                 .wait_for_default_pr(|pr| pr.mergeable_state == MergeableState::Unknown)
                 .await?;
-            tester.refresh().await;
             tester
                 .default_repo()
                 .lock()
                 .get_pr_mut(default_pr_number())
                 .mergeable_state = OctocrabMergeableState::Dirty;
+            tester.refresh().await;
             tester
                 .wait_for_default_pr(|pr| pr.mergeable_state == MergeableState::HasConflicts)
                 .await?;
--- a/src/bors/handlers/refresh.rs
+++ b/src/bors/handlers/refresh.rs
@@ -20,7 +20,6 @@ pub async fn refresh_repository(
     mergeable_queue_tx: MergeableQueueSender,
 ) -> anyhow::Result<()> {
     let repo = repo.as_ref();
-
     let results = join_all([
         cancel_timed_out_builds(repo, db.as_ref()).boxed(),
         reload_permission(repo, team_api_client).boxed(),
--- a/src/bors/handlers/refresh.rs
+++ b/src/bors/handlers/refresh.rs
@@ -3,10 +3,13 @@ use std::time::Duration;
 
 use anyhow::Context;
 use chrono::{DateTime, Utc};
+use futures::FutureExt;
+use futures::future::join_all;
 
 use crate::bors::Comment;
 use crate::bors::RepositoryState;
 use crate::bors::handlers::trybuild::cancel_build_workflows;
+use crate::bors::mergeable_queue::MergeableQueueSender;
 use crate::database::BuildStatus;
 use crate::{PgDbClient, TeamApiClient};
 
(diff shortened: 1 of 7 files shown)


src/bors/handlers/refresh.rs
```diff
@@ -16,10 +16,11 @@ pub async fn refresh_repository(
     team_api_client: &TeamApiClient,
 ) -> anyhow::Result<()> {
     let repo = repo.as_ref();
-    if let (Ok(_), _, Ok(_)) = tokio::join!(
+    if let (Ok(_), _, Ok(_), Ok(_)) = tokio::join!(
```
Kobzol, line 19: It might be time to switch to `join_all` here :)

src/database/operations.rs
```diff
@@ -189,6 +189,45 @@ pub(crate) async fn update_mergeable_states_by_base_branch(
...
 
+pub(crate) async fn get_prs_with_unknown_mergeable_state(
+    executor: impl PgExecutor<'_>,
+    repo: &GithubRepoName,
+) -> anyhow::Result<Vec<PullRequestModel>> {
+    measure_db_query("get_prs_with_unknown_mergeable_state", || async {
+        let prs = sqlx::query_as!(
+            PullRequestModel,
+            r#"
+SELECT
```
Kobzol, line 200: Uhh, it is annoying that we have to repeat this complicated fragment all the time :thinking: I asked on the sqlx Discord if this can be improved somehow.

Sakib25800:
@Kobzol could you take a quick look to make sure I'm on the right track on how I've implemented the background processing for the mergeable queue?

You can ignore the logic and what actually happens right now since it's very much a WIP, but I want to make sure I'm going about the actual queue task implementation correctly.

src/database/operations.rs
Sakib25800: Seems we'll have to go for a string literal concat approach for now.

Kobzol: I scanned the file again and realized that we only do this query twice after this PR, I thought it's much more than that :D So it's fine for now. In the future we could do a macro with a literal concatenation indeed.

src/bors/handlers/refresh.rs
```diff
@@ -207,6 +238,51 @@ timeout = 3600
...
+        run_test(pool, |mut tester| async {
+            tester
+                .edit_pr(default_repo_name(), default_pr_number(), |pr| {
+                    pr.mergeable_state = OctocrabMergeableState::Unknown
+                })
+                .await?;
+            tester
+                .wait_for_default_pr(|pr| pr.mergeable_state == MergeableState::Unknown)
+                .await?;
+            tester.refresh().await;
```
Kobzol, line 252: Shouldn't we set `Dirty` *before* doing the refresh? In theory it could race and the refresh could run sooner than `Dirty` is set.

Sakib25800: Yeah, it could. Fixed.

src/bors/handlers/refresh.rs
```diff
@@ -3,24 +3,32 @@ use std::time::Duration;
...
-    ) {
+    let results = join_all([
+        cancel_timed_out_builds(repo, db.as_ref()).boxed(),
+        reload_permission(repo, team_api_client).boxed(),
+        reload_config(repo).boxed(),
+        reload_unknown_mergeable_prs(repo, db.as_ref(), mergeable_queue_tx).boxed(),
+    ])
+    .await;
+
+    if results.iter().all(|result| result.is_ok()) {
```
Kobzol, line 31: We're swallowing errors here, we should log them instead. It was pre-existing though, so let's fix it in a separate PR.

Kobzol (approved):
Thanks

PR #260, opened 2025-03-28
Prevent spam from GitHub mentions in merge commits

### LINKED ISSUE #258: Implement mention suppression
Bors should remove mentions (`@<username>`) from PR descriptions, which are turned into merge commit bodies, to avoid spamming people on GitHub.

See https://github.com/rust-lang/homu/pull/100 and https://github.com/rust-lang/homu/pull/230 for more details.

xosnrdev:
Resolves #258

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -1,2 +1,3 @@
 pub mod logging;
+pub mod text;
 pub mod timing;
--- a/src/utils/text.rs
+++ b/src/utils/text.rs
@@ -1,0 +1,29 @@
+use regex::{Captures, Regex};
+
+/// Replaces github @mentions with backticks to prevent accidental pings
+pub fn suppress_github_mentions(text: &str) -> String {
+    if !text.contains('@') {
+        return text.to_string();
+    }
+
+    let pattern = r"\B(@\S+)";
+
+    let re = Regex::new(pattern).unwrap();
+    re.replace_all(text, |caps: &Captures| format!("`{}`", &caps[1]))
+        .to_string()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_suppress_github_mentions() {
+        assert_eq!(suppress_github_mentions("r? @matklad\n"), "r? `@matklad`\n");
+        assert_eq!(suppress_github_mentions("@bors r+\n"), "`@bors` r+\n");
+        assert_eq!(
+            suppress_github_mentions("mail@example.com"),
+            "mail@example.com"
+        )
+    }
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,9 @@ chrono = "0.4"
 
 itertools = "0.14"
 
+# Text processing
+regex = "1"
+
 [dev-dependencies]
 insta = "1.26"
 wiremock = "0.6"
--- a/src/bors/handlers/trybuild.rs
+++ b/src/bors/handlers/trybuild.rs
@@ -20,6 +20,7 @@ use crate::github::{
     CommitSha, GithubUser, LabelTrigger, MergeError, PullRequest, PullRequestNumber,
 };
 use crate::permissions::PermissionType;
+use crate::utils::text::suppress_github_mentions;
 
 use super::deny_request;
 use super::has_permission;
@@ -286,7 +287,7 @@ fn auto_merge_commit_message(
 {pr_message}"#,
         pr_label = pr.head_label,
         pr_title = pr.title,
-        pr_message = pr.message,
+        pr_message = suppress_github_mentions(&pr.message),
         repo_owner = name.owner(),
         repo_name = name.name()
     );


Kobzol:
Hi, thanks for the PR, I'll try to take a look in the upcoming days.

src/utils/mod.rs
```diff
@@ -1,2 +1,99 @@
 pub mod logging;
 pub mod timing;
+
+/// Converts GitHub @mentions to markdown-backticked text to prevent notifications.
```
Kobzol, line 4: Let's put this into a `text` module.

src/utils/mod.rs
```diff
@@ -1,2 +1,99 @@
...
+    re.replace_all(text, |caps: &regex::Captures| {
+        let mention = &caps[0];
+        let position = caps.get(0).unwrap().start();
+
+        if !is_github_mention(text, mention, position) {
+            return mention.to_string();
+        }
+
+        let name = &mention[1..]; // Drop the @ symbol
+        format!("`{}`", name)
```
Kobzol, line 34:
```suggestion
        format!("`{name}`")
```

src/utils/mod.rs
```diff
@@ -1,2 +1,99 @@
...
+        return text.to_string();
+    }
+
+    let segment = r"[A-Za-z0-9][A-Za-z0-9\-]{0,38}";
+    let pattern = format!(r"@{0}(?:/{0})*", segment);
+
+    let re = regex::Regex::new(&pattern).unwrap();
+    re.replace_all(text, |caps: &regex::Captures| {
+        let mention = &caps[0];
+        let position = caps.get(0).unwrap().start();
```
Kobzol, line 27:
```suggestion
        let entry = caps.get(0).unwrap();
        let mention = entry.as_str();
        let position = entry.start();
```

src/utils/mod.rs
```diff
@@ -1,2 +1,99 @@
...
+        format!("`{}`", name)
+    })
+    .to_string()
+}
+
+// Determines if a potential mention would actually trigger a notification
+fn is_github_mention(text: &str, mention: &str, pos: usize) -> bool {
+    // Not a valid mention if preceded by alphanumeric or underscore (email)
+    if pos > 0 {
+        let c = text.chars().nth(pos - 1).unwrap();
```
Kobzol, line 43: This is not correct, I think. `.start()` returns a byte offset, but here you are accessing the `nth` Unicode char, that does not necessarily correspond to each other. For example, in `čauš@gmail.com`, the variable `c` corresponds to the character `g`!

Since you always need to check the preceding and following character, I would modify the regex to always capture one character before and after the mention match, and then check that explicitly, rather than messing with positions, that's always quite tricky in Unicode text.

src/utils/mod.rs
```diff
@@ -1,2 +1,99 @@
...
+
+    true
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_suppress_github_mentions() {
```
Kobzol, line 66: Please split the test into multiple tests that check the different properties (essentially the different commented sections). And also add some Unicode text to make sure that we have coverage of it.

Kobzol (commented):
Thank you for the PR! The implementation looks like a good start, but there are a few things that should be modified, I left a bunch of comments.

xosnrdev:
Thanks for the feedback. I didn't put unicode text into consideration 😞.

Will resolve shortly.

xosnrdev:
Would fuzz test be ideal or overkill ?

Kobzol (commented):
Fuzz testing would be way overkill I think :) To clarify, I didn't mean that the GitHub username could include Unicode, as far as I know, it cannot contain it. So for matching the usernames, doing what you had before (`a-zA-Z0-9`) was fine. Just the way you searched for the character before/after the match wasn't correct.

The code is now more complicated than before. What I meant was to solve the preceding/following word in the regex itself.

What do you think about this regex? `(^|\s)@([a-zA-Z0-9][a-zA-Z0-9]*)($|\s)`

Kobzol:
I know that `regex` doesn't support lookaround, but we don't need it, right? We can just match the before/after part of the regex, and then ignore it, it doesn't need to be lookbehind/lookahead. Of course the regex would have to be extended for teams, it was just an example.

Kobzol:
Nice, this made the code much shorter and easier to understand! :) I think that we should use `\s` for the before/after chars though, otherwise e.g. `abč@user.com` would count as a mention.

xosnrdev:
I'll add test for this `abč@user.com`.

using `/s` will fail this test `Hello,@user! How are you?` as it matches whitespace characters like spaces, tabs, and newlines

Kobzol:
> using /s will fail this test Hello,@user! How are you? as it matches whitespace characters like spaces, tabs, and newlines

Yes, that was the idea :D I wonder if GitHub does indeed ping in these situations. It doesn't seem like so? I think that whitespace should be enough.

Hello,@xosrndev, are you pinged?

Kobzol:
Hello, @xosrndev, are you pinged?

Hmm, it doesn't ping even above, unless I manually select the user. I guess that it shouldn't be tested in PR/issue comments, but rather in the commit messages directly. We should first test on some test repo in which exact situations does GitHub indeed ping when the mention is in the commit message.

xosnrdev:
Well it did ping on a test repo.

Screenshot 2025-04-02 at 11 01 53 AM

Screenshot 2025-04-02 at 11 06 37 AM

Kobzol:
Interesting, thanks for testing that out! Ok, in that case let's just use what homu does for now, which is `\B`, so "not a word boundary". It tries to detect a situation where there is a word on the left, but not on the right (and on the right we always have `@`). I'm not fully sure how https://github.com/rust-lang/homu/pull/230 works yet, so let's start with just copying homu, and we can improve the regex in the future, once we have the baseline functionality merged.

xosnrdev:
bet.

src/utils/text.rs
```diff
@@ -0,0 +1,29 @@
+use regex::{Captures, Regex};
+
+/// Replaces valid GitHub @mentions with backticks to prevent accidental pings
+pub fn suppress_github_mentions(text: &str) -> String {
+    if text.is_empty() || !text.contains('@') {
```
Kobzol, line 5:
```suggestion
    if !text.contains('@') {
```
It it unnecessary.

Kobzol (commented):
Thanks! Copying homu is usually not a bad default, we can always make it smarter in the future. Left one last nit, otherwise LGTM.

PR #262, opened 2025-03-28
Escape `|` to avoid interpreting it as column seperator on Github

dennisschagt:
Before (see [docs/commands.md](https://github.com/rust-lang/bors/blob/fb973017022e2195769eb92e236724a9e6bbffe4/docs/commands.md)):
image

After:
image


Kobzol (approved):
Good catch, thanks!

PR #265, opened 2025-04-02
Implement mergeable queue loop

Sakib25800:
Prerequisite to #247.

This PR implements the mergeable queue processing loop.

- Pushes to the queue after pushes to the base branch
- Pushes to the queue after PR edit

### DIFF
--- a/src/bin/bors.rs
+++ b/src/bin/bors.rs
@@ -6,8 +6,9 @@ use std::time::Duration;
 
 use anyhow::Context;
 use bors::{
-    BorsContext, BorsGlobalEvent, CommandParser, PgDbClient, ServerState, TeamApiClient,
-    WebhookSecret, create_app, create_bors_process, create_github_client, load_repositories,
+    BorsContext, BorsGlobalEvent, BorsProcess, CommandParser, PgDbClient, ServerState,
+    TeamApiClient, WebhookSecret, create_app, create_bors_process, create_github_client,
+    load_repositories,
 };
 use clap::Parser;
 use sqlx::postgres::PgConnectOptions;
@@ -109,7 +110,12 @@ fn try_main(opts: Opts) -> anyhow::Result<()> {
     }
 
     let ctx = BorsContext::new(CommandParser::new(opts.cmd_prefix), Arc::new(db), repos);
-    let (repository_tx, global_tx, bors_process) = create_bors_process(ctx, client, team_api);
+    let BorsProcess {
+        repository_tx,
+        global_tx,
+        bors_process,
+        ..
+    } = create_bors_process(ctx, client, team_api);
 
     let refresh_tx = global_tx.clone();
     let refresh_process = async move {
--- a/src/bors/handlers/mod.rs
+++ b/src/bors/handlers/mod.rs
@@ -31,6 +31,8 @@ use tracing::Instrument;
 #[cfg(test)]
 use crate::tests::util::TestSyncMarker;
 
+use super::mergeable_queue::MergeableQueueSender;
+
 mod help;
 mod info;
 mod labels;
@@ -48,6 +50,7 @@ pub static WAIT_FOR_WORKFLOW_STARTED: TestSyncMarker = TestSyncMarker::new();
 pub async fn handle_bors_repository_event(
     event: BorsRepositoryEvent,
     ctx: Arc<BorsContext>,
+    mergeable_queue_tx: MergeableQueueSender,
 ) -> anyhow::Result<()> {
     let db = Arc::clone(&ctx.db);
     let Some(repo) = ctx
@@ -129,7 +132,7 @@ pub async fn handle_bors_repository_event(
             let span =
                 tracing::info_span!("Pull request edited", repo = payload.repository.to_string());
 
-            handle_pull_request_edited(repo, db, payload)
+            handle_pull_request_edited(repo, db, mergeable_queue_tx, payload)
                 .instrument(span.clone())
                 .await?;
         }
@@ -199,7 +202,7 @@ pub async fn handle_bors_repository_event(
             let span =
                 tracing::info_span!("Pushed to branch", repo = payload.repository.to_string());
 
-            handle_push_to_branch(repo, db, payload)
+            handle_push_to_branch(repo, db, mergeable_queue_tx, payload)
                 .instrument(span.clone())
                 .await?;
         }
--- a/src/bors/handlers/pr_events.rs
+++ b/src/bors/handlers/pr_events.rs
@@ -5,6 +5,7 @@ use crate::bors::event::{
     PushToBranch,
 };
 use crate::bors::handlers::labels::handle_label_trigger;
+use crate::bors::mergeable_queue::MergeableQueueSender;
 use crate::bors::{Comment, PullRequestStatus, RepositoryState};
 use crate::database::MergeableState;
 use crate::github::{CommitSha, LabelTrigger, PullRequestNumber};
@@ -13,6 +14,7 @@ use std::sync::Arc;
 pub(super) async fn handle_pull_request_edited(
     repo_state: Arc<RepositoryState>,
     db: Arc<PgDbClient>,
+    mergeable_queue: MergeableQueueSender,
     payload: PullRequestEdited,
 ) -> anyhow::Result<()> {
     let pr = &payload.pull_request;
@@ -31,6 +33,8 @@ pub(super) async fn handle_pull_request_edited(
     let Some(_) = payload.from_base_sha else {
         return Ok(());
     };
+
+    mergeable_queue.enqueue(pr_model.repository.clone(), pr_number);
 
     if !pr_model.is_approved() {
         return Ok(());
@@ -154,6 +158,7 @@ pub(super) async fn handle_pull_request_ready_for_review(
 pub(super) async fn handle_push_to_branch(
     repo_state: Arc<RepositoryState>,
     db: Arc<PgDbClient>,
+    mergeable_queue: MergeableQueueSender,
     payload: PushToBranch,
 ) -> anyhow::Result<()> {
     let rows = db
@@ -163,8 +168,18 @@ pub(super) async fn handle_push_to_branch(
             MergeableState::Unknown,
         )
         .await?;
+    let affected_prs = db
+        .get_nonclosed_pull_requests_by_base_branch(repo_state.repository(), &payload.branch)
+        .await?;
 
-    tracing::info!("Updated mergeable_state to `unknown` for {} PR(s)", rows);
+    tracing::info!(
+        "Adding {} PR(s) to the mergeable queue due to base branch change",
+        rows
+    );
+
+    for pr in affected_prs {
+        mergeable_queue.enqueue(repo_state.repository().clone(), pr.number);
+    }
 
     Ok(())
 }
@@ -207,7 +222,7 @@ mod tests {
     use crate::bors::PullRequestStatus;
     use crate::tests::mocks::default_pr_number;
     use crate::{
-        database::MergeableState,
+        database::{MergeableState, OctocrabMergeableState},
         tests::mocks::{User, default_branch_name, default_repo_name, run_test},
     };
 
@@ -344,7 +359,7 @@ mod tests {
         run_test(pool.clone(), |mut tester| async {
             tester
                 .edit_pr(default_repo_name(), default_pr_number(), |pr| {
-                    pr.mergeable_state = octocrab::models::pulls::MergeableState::Dirty;
+                    pr.mergeable_state = OctocrabMergeableState::Dirty;
                 })
                 .await?;
             tester
@@ -425,6 +440,7 @@ mod tests {
         .await;
     }
 
+    #[tracing_test::traced_test]
     #[sqlx::test]
     async fn open_and_merge_pr(pool: sqlx::PgPool) {
         run_test(pool, |mut tester| async {
@@ -439,6 +455,53 @@ mod tests {
                 .wait_for_pr(default_repo_name(), pr.number.0, |pr| {
                     pr.pr_status == PullRequestStatus::Merged
                 })
+                .await?;
+            Ok(tester)
+        })
+        .await;
+    }
+
+    #[sqlx::test]
+    async fn mergeable_queue_processes_pr_base_change(pool: sqlx::PgPool) {
+        run_test(pool, |mut tester| async {
+            let branch = tester.create_branch("beta").clone();
+            tester
+                .edit_pr(default_repo_name(), default_pr_number(), |pr| {
+                    pr.base_branch = branch;
+                    pr.mergeable_state = OctocrabMergeableState::Unknown;
+                })
+                .await?;
+            tester
+                .wait_for_default_pr(|pr| pr.mergeable_state == MergeableState::Unknown)
+                .await?;
+            tester
+                .default_repo()
+                .lock()
+                .get_pr_mut(default_pr_number())
+                .mergeable_state = OctocrabMergeableState::Dirty;
+            tester
+                .wait_for_default_pr(|pr| pr.mergeable_state == MergeableState::HasConflicts)
+                .await?;
+            Ok(tester)
+        })
+        .await;
+    }
+
+    #[sqlx::test]
+    async fn enqueue_prs_on_push_to_branch(pool: sqlx::PgPool) {
+        run_test(pool, |mut tester| async {
+            tester.open_pr(default_repo_name(), false).await?;
+            tester.push_to_branch(default_branch_name()).await?;
+            tester
+                .wait_for_default_pr(|pr| pr.mergeable_state == MergeableState::Unknown)
+                .await?;
+            tester
+                .default_repo()
+                .lock()
+                .get_pr_mut(default_pr_number())
+                .mergeable_state = OctocrabMergeableState::Dirty;
+            tester
+                .wait_for_default_pr(|pr| pr.mergeable_state == MergeableState::HasConflicts)
                 .await?;
             Ok(tester)
         })
--- a/src/bors/mergeable_queue.rs
+++ b/src/bors/mergeable_queue.rs
@@ -1,0 +1,320 @@
+use crate::database::OctocrabMergeableState;
+use crate::github::{GithubRepoName, PullRequestNumber};
+use std::cmp::Reverse;
+use std::collections::BinaryHeap;
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, Instant};
+use tokio::sync::Notify;
+use tokio::time::timeout;
+
+use super::BorsContext;
+
+/// Delay before processing a mergeable queue item for the first time.
+const BASE_DELAY: Duration = Duration::from_millis(500);
+/// Exponential backoff delay multiplier.
+const BACKOFF_MULTIPLIER: f64 = 2.0;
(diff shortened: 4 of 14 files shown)


src/bors/handlers/pr_events.rs
```diff
@@ -444,4 +454,33 @@ mod tests {
         })
         .await;
     }
+
+    #[sqlx::test]
+    async fn mergeable_queue_processes_pr_base_change(pool: sqlx::PgPool) {
```
Sakib25800, line 459: In this test, I chose to not explicitly check that the edited PR has been queued. Instead, I verify that its side effects (the PR’s mergeable state being updated in the DB) are successful.

@Kobzol is this fine, or do I need to extend BorsTester to “spy” on the mergeable queue?

We don't _really_ need to check the mergeable queue in tests since it's not like some other method will also update the `mergeable_state` I guess.

src/bors/handlers/pr_events.rs
```diff
@@ -444,4 +454,33 @@ mod tests {
...
+        run_test(pool, |mut tester| async {
+            tester
+                .edit_pr(default_repo_name(), default_pr_number(), |pr| {
+                    pr.mergeable_state = OctocrabMergeableState::Clean;
```
Kobzol, line 463: Btw, I'm not so sure if `Clean` actually means completely without conflicts? I recently saw a PR that had the "Merge conflicts" marker added by GH, but it also said "the PR has merge conflicts and can be cleanly merged", or something like that.

src/bors/handlers/pr_events.rs
Kobzol: This is fine. We shouldn't test implementation details, but rather the "externally visible behavior", which is either comments, GitHub state or DB state (if there is no other way).

We will need to somehow simulate in the test that the PR state has changed, through our GH API mocks. We can e.g. wait a second and then set the new PR state, and check that it is reflected in the DB. To speed the test up, we can either make refresh in tests be super fast or (which I would prefer) just manually trigger refresh, which we already do in other tests.

src/bors/mergeable_queue.rs
```diff
@@ -0,0 +1,41 @@
...
+pub struct MergeableQueue {
+    tx: Option<mpsc::Sender<MergeableQueueItem>>,
+}
+
+impl MergeableQueue {
+    pub fn new() -> Self {
+        Self { tx: None }
+    }
+
+    pub fn with_sender(tx: mpsc::Sender<MergeableQueueItem>) -> Self {
```
Kobzol, line 20: Why does it need to be `Option`?

src/bors/handlers/mod.rs
```diff
@@ -454,6 +457,36 @@ async fn reload_repos(
...
+        .unwrap()
+        .get(&mergeable_queue_item.repository)
+        .cloned()
+        .unwrap();
+    let fetched_pr = repo_state
+        .client
+        .get_pull_request(mergeable_queue_item.pr_number)
+        .await?;
+
+    ctx.db
```
Kobzol, line 484: Right now, the queue isn't super useful, because instead of doing "fetch PR state immediately", we do "push PR to queue and then fetch its state immediately", and then we forget the PR.

We don't know beforehand when will the mergeable state be updated, and it might take e.g. a few seconds. We thus need to figure out how to build the queue in a way where we can say something like "do this after N seconds", and then when the state is still unknown, we should requeue the PR again.

A naive solution would be to use `tokio::spawn` with `tokio::timeout`, which would put an item into the queue after a specified duration. I hope that we can use something more structured though.

Maybe https://docs.rs/tokio-util/latest/tokio_util/time/delay_queue/struct.DelayQueue.html?

src/bors/handlers/pr_events.rs
Sakib25800: You're right it doesn't mean without conflict (it says CAN be merged in the API description), but I guess it's technically `Mergeable` which means `Clean -> Mergeable`, which is what we have currently so it isn't technically wrong. 

But it's also possible to be `Mergeable` and `HasConflicts` at the same time.

src/bors/handlers/mod.rs
Sakib25800: Bors got away with doing that somehow, but I thought the same. 

I've used `DelayQueue`.

src/bors/mergeable_queue.rs
Sakib25800: I've reworked this quite a bit.

Sakib25800:
I've not actually written much async Rust code, so sorry for the dumb mistakes that this PR probably has.

src/bors/handlers/pr_events.rs
Kobzol: > I'm pretty sure

I hope it does, but we should check (later) :laughing:

src/database/client.rs
```diff
@@ -68,6 +77,14 @@ impl PgDbClient {
         update_mergeable_states_by_base_branch(&self.pool, repo, base_branch, mergeable_state).await
     }
 
+    pub async fn get_pull_requests_by_base_branch(
```
Kobzol, line 80:
```suggestion
    pub async fn get_nonclosed_pull_requests_by_base_branch(
```
Otherwise the name of the function kind of lies (it doesn't return all PRs by base branch).

src/github/server.rs
```diff
@@ -90,11 +91,14 @@ pub fn create_bors_process(
...
 ) {
+    let ctx = Arc::new(ctx);
+
     let (repository_tx, repository_rx) = mpsc::channel::<BorsRepositoryEvent>(1024);
     let (global_tx, global_rx) = mpsc::channel::<BorsGlobalEvent>(1024);
+    let mut mergeable_queue = MergeableQueue::new(ctx.clone());
 
     let service = async move {
-        let ctx = Arc::new(ctx);
+        mergeable_queue.start();
```
Kobzol, line 101: Any particular reason why you use `tokio::spawn` vs just returning a `Future` that we could poll here? I prefer that, because it makes it simpler to track the execution of the future, and also make sure that it stops together with the `service` future.

src/bors/mergeable_queue.rs
```diff
@@ -0,0 +1,273 @@
...
+    github::{GithubRepoName, PullRequestNumber},
+};
+
+use super::BorsContext;
+
+type ScheduledPullRequests = Arc<Mutex<DelayQueue<MergeableCheckEntry>>>;
+type TrackedPullRequests = Arc<Mutex<HashMap<MergeableCheckKey, Instant>>>;
+
+/// How long to wait before checking mergeable state (for first time).
+const INITIAL_DELAY: Duration = Duration::from_secs(2);
```
Kobzol, line 24: We might want to override this in tests to make it shorter. There are also better approaches of how to deterministically simulate time in tests, even with tokio (https://docs.rs/tokio/latest/tokio/time/fn.pause.html), but that's out of scope for this PR :)

src/bors/mergeable_queue.rs
```diff
@@ -0,0 +1,273 @@
...
+            ctx,
+        }
+    }
+
+    pub fn start(&mut self) {
+        let scheduled_prs = Arc::clone(&self.scheduled_prs);
+        let tracked_prs = Arc::clone(&self.tracked_prs);
+        let ctx = Arc::clone(&self.ctx);
+
+        let (shutdown_tx, mut shutdown_rx) = mpsc::channel::<()>(1);
```
Kobzol, line 84: Isn't this just `tokio::sync::oneshot`? :thinking:

Kobzol (commented):
This is a complex one! I left some comments.

src/bors/mergeable_queue.rs
```diff
@@ -0,0 +1,273 @@
...
+}
+
+/// A queue system for updating PR mergeable states.
+///
+/// This queue allows for PRs to be scheduled for a mergeable_state recheck after a delay.
+/// - PRs are initially checked after a short delay
+/// - If still unknown, check will be retried with delays
+pub struct MergeableQueue {
+    scheduled_prs: ScheduledPullRequests,
+    tracked_prs: TrackedPullRequests,
```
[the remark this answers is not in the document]
Sakib25800: The key comes from the `.insert` method, so I'm not sure this would work as it would have had to be inserted first  then `try_remove`, but we would have just inserted it so it will always be `Some`.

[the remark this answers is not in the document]
Kobzol: I see, sorry, I was confused by the fact that you also insert `MergeableCheck*Key*` into the queue, but in terms of the queue's naming, it's actually a value. I somehow assumed that the values in the queue have to be unique, but it looks like the queue allows the same value to be stored multiple times.

In that case there is a small edge case in this:
```rust
if let Some(expire_time) = tracked_prs.get(&key) {
    if *expire_time > now {
        tracing::debug!("Ignoring PR {key} as already in mergeable queue");
        return;
    }
}

let expire_time = now + INITIAL_DELAY;
tracked_prs.insert(key.clone(), expire_time);
```
If `expire_time` is `<= now`, then the item will be inserted into the queue two times, which is not ideal. When `expire_time > now && expire_time < now + INITIAL_DELAY`, then we postpone an already "scheduled" PR. In theory, if `enqueue` was called with the right interval (e.g. every second), then the PR would never be checked, because it would always get reset in the queue.

But both situations are rare edge cases, so I'm fine with ignoring them.

Sakib25800:
Due to this PR being reworked a few times 😓, most of these comments should be resolved as they no longer apply.

src/bors/mergeable_queue.rs
Sakib25800: I was going to just make test specific constants, but this is a better approach. 

I'll make a PR for this after this one (or you can make an issue if there's more that can be done to speed up tests).

src/tests/mocks/bors.rs
```diff
@@ -616,6 +626,7 @@ impl BorsTester {
     async fn finish(self, bors: JoinHandle<()>) -> GitHubState {
         // Make sure that the event channel senders are closed
         drop(self.app);
+        self.mergeable_queue_tx.close();
```
Sakib25800, line 629: I wanted mergeable_queue to behave like an `mpsc` channel, returning `None` when all senders are dropped. 

But `consume_mergeable_queue` holds onto a sender so it can re‑enqueue items, so it keeps one sender reference alive indefinitely.

So I ended up opting for this explicit `.close()` approach.

src/bin/bors.rs
```diff
@@ -109,7 +109,8 @@ fn try_main(opts: Opts) -> anyhow::Result<()> {
     }
 
     let ctx = BorsContext::new(CommandParser::new(opts.cmd_prefix), Arc::new(db), repos);
-    let (repository_tx, global_tx, bors_process) = create_bors_process(ctx, client, team_api);
+    let (repository_tx, global_tx, _mergeable_queue_tx, bors_process) =
```
Kobzol, line 112: I think it's high time to return a struct with named fields here, a four tuple is becoming too messy, especially since some of these variables have the same type and could be swapped inadvertedly.

src/bors/handlers/pr_events.rs
```diff
@@ -444,4 +459,58 @@ mod tests {
...
+                .wait_for_default_pr(|pr| pr.mergeable_state == MergeableState::Mergeable)
+                .await?;
+
+            let branch = tester.create_branch("beta").clone();
+            tester
+                .edit_pr(default_repo_name(), default_pr_number(), |pr| {
+                    pr.base_branch = branch;
+                })
+                .await?;
+            tester
```
Kobzol, line 481: We should wait for the PR having the `Unknown` state here, to make sure that the bot can deal with `Dirty` being set sometime in the future.

src/bors/handlers/pr_events.rs
```diff
@@ -444,4 +459,58 @@ mod tests {
         })
         .await;
     }
+
+    #[sqlx::test]
+    async fn mergeable_queue_processes_pr_base_change(pool: sqlx::PgPool) {
+        run_test(pool, |mut tester| async {
```
Kobzol, line 465: This prelude with `edit_pr` and `wait_for_default_pr` seems like it's not needed for this test?

src/tests/mocks/bors.rs
Kobzol: If we used a normal tokio queue, we could use https://docs.rs/tokio/1.44.2/tokio/sync/mpsc/struct.Sender.html#method.downgrade. Since you already store `Arc` in the sender, I guess that you could emulate it with a "Weaksender" variant that only stores a `Weak` pointer.

src/bors/mergeable_queue.rs
```diff
@@ -0,0 +1,316 @@
...
+        MergeableQueueSender {
+            inner: shared.clone(),
+        },
+        MergeableQueueReceiver { inner: shared },
+    )
+}
+
+impl MergeableQueueSender {
+    pub fn close(&self) {
+        self.inner.closed.store(true, Ordering::Relaxed);
```
Kobzol, line 100: Not sure if relaxed is enough, the compiler or the CPU could theoretically reorder this to be execute after the notify below. In theory we should use `acquire + release` semantics, but I don't want to think about that, so let's just use `SeqCst` everywhere.

src/bors/mergeable_queue.rs
```diff
@@ -0,0 +1,316 @@
...
+            MergeableQueueItem {
+                pull_request: QueuedPullRequest { pr_number, repo },
+                attempt: 1,
+                immediate: false,
+            },
+            expiration,
+        );
+    }
+
+    #[allow(dead_code)]
```
Kobzol, line 117: Oh, well if we don't actually use `enqueue_now` anywhere, then just delete it =D

src/bors/mergeable_queue.rs
```diff
@@ -0,0 +1,316 @@
...
+            MergeableQueueItem {
+                pull_request: QueuedPullRequest { pr_number, repo },
+                attempt: 1,
+                immediate: true,
+            },
+            None,
+        );
+    }
+
+    pub fn enqueue_retry(&self, queue_item: MergeableQueueItem) {
```
Kobzol, line 129:
```suggestion
    fn enqueue_retry(&self, queue_item: MergeableQueueItem) {
```
This should be private (at least until the processing logic is still in this file).

src/bors/mergeable_queue.rs
```diff
@@ -0,0 +1,316 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}/#{}", self.repo, self.pr_number)
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct MergeableQueueItem {
+    pub pull_request: QueuedPullRequest,
+    pub attempt: u32,
+    pub immediate: bool,
```
Kobzol, line 36:
```suggestion

```
This seems to be unused, it's only set and propagated, but never read.

src/bors/mergeable_queue.rs
```diff
@@ -0,0 +1,316 @@
...
+}
+
+impl Ord for Item {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.expiration.cmp(&other.expiration)
+    }
+}
+
+struct SharedInner {
+    queue: Mutex<BinaryHeap<Item>>,
```
Kobzol, line 68: Let's use `parking_lot::Mutex` here, since we already have it in the crate.

src/bors/mergeable_queue.rs
```diff
@@ -0,0 +1,316 @@
...
+        // Notify when:
+        // 1. The current item expires sooner than the head of the queue
+        let has_earlier_expiration =
+            queue
+                .peek()
+                .is_some_and(|head| match (expiration, head.expiration) {
+                    (Some(new_exp), Reverse(Some(head_exp))) => new_exp < head_exp,
+                    _ => false,
+                });
+        // 2. The queue was empty before insertion (reader is waiting)
```
Kobzol, line 156:
```suggestion
        // 2. The queue was empty before insertion (reader might be waiting)
```
There might not be any reader.

src/bors/mergeable_queue.rs
```diff
@@ -0,0 +1,316 @@
...
+    notify: Notify,
+    closed: AtomicBool,
+}
+
+#[derive(Clone)]
+pub struct MergeableQueueSender {
+    inner: Arc<SharedInner>,
+}
+
+#[derive(Clone)]
```
Kobzol, line 78:
```suggestion

```
The receiver should not be cloneable, this implementation doesn't really assume multiple concurrent readers, it would complicate it unnecessarily.

src/tests/mocks/bors.rs
Kobzol: Or, an alternative design could be that `dequeue` returns you both the item and a sender, and then before calling `dequeue` you drop the sender, so there should not be any outstanding senders.

Kobzol (commented):
The design with the closing boolean isn't super nice, and it's tricky to figure out all the possible interleavings for the `dequeue` method. An alternative could be to use weak senders, or create an explicit Stop message that can be sent to the queue to stop it. But even with the boolean I think that the implementation is fine, and it will run forever in the live service anyway, so it's not a real issue outside of tests.

src/tests/mocks/bors.rs
Sakib25800: I'll go for the Weaksender approach. I realised that the approach of checking `if Arc::strong_count(&self.inner) == 2` is a little brittle.

src/bors/handlers/pr_events.rs
Sakib25800: Removed it. 

It's not needed. I thought it would be good to make it clear that the state changes.

src/bors/mergeable_queue.rs
Sakib25800: `parking_lot` is a dev dep, so I figured no other `src` code uses it so its not worth promoting to an actual dep.

src/tests/mocks/bors.rs
Kobzol: I would go with one of the following:
- Store a weak sender Rc in the receiver. When a message is received, try to upgrade the weak to strong and return it together with the message. If it can't be upgraded, stop the reading loop. When a sender is dropped, it has to notify the reader, otherwise it might not ever realized that the last sender was removed.
- Just add a message type to the reader that tells it to quit.

Honestly, I would just go with option 2. Dealing with the notify, RC counters, weak/strong references etc. seems like something that is too low-level for bors. We don't need high performance here, we just need to be sure that it is rock solid. With an explicit message type, that should be the easiest. We will need to make sure that the retry logic can deal with a failing send (in case the receiver has been stopped in the meantime), but that shouldn't be hard.

Sakib25800: Option 2 sounds like what I had initially but with an explicit message type for shut down. I've now gone for this. 

Option 1 looks quite complicated and not worth it considering shutdowns are only for tests.

Kobzol:
Thanks! We don't actually need the `closed` boolean anymore if we consider the shutdown to be more of an "interrupt" that closes the queue only once (interrupts the reader who uses `dequeue` as an iterator), as that is how the queue is supposed to be used, but we can keep it, it doesn't really matter. If cou yould please squash the commits, and we can finally merge this :) Great work!

Sakib25800:
Squashed. 

Also removed the `closed` boolean.

Kobzol (approved):
Thank you for putting up with all my review comments :laughing: Really awesome job.

Issue #255, opened 2025-03-25
Handle GitHub API timeouts and retries


Kobzol:
Currently, we send requests to the GitHub API without dealing with timeouts and retries in any way. We should make this more robust by adding some automatic retry/timeout mechanism:

- Implement timeouts for the requests. Say, 10s at most, before the request is marked as timeouted.
- Implement retry mechanism for GH API requests. If the request fails due to API rate limiting or a timeout, retry it again, ideally with some exponential backoff. See https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#checking-the-status-of-your-rate-limit and https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit.
- I don't think that this will be needed at first, but in the future we might consider using some queue for requests (https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2022-11-28) to avoid too many concurrent requests. But I would only implement this if we find some issues in production.

geetanshjuneja:
Hi, I would like to work on this.

geetanshjuneja:
Hi, for the timeout part, I just need to check if the api call takes more than 10 sec then it needs to cancelled which can be done using `tokio::time::timeout`.

Kobzol:
`tokio::time::timeout` is a bit low-level. We could use it if needed, but I suppose that `reqwest` or `octocrab` should have more fine-grained options for handling timeouts, e.g. read timeouts, connection timeouts, request timeouts, etc.

geetanshjuneja:
`octocrabbuilder::set_read_timeout` would be fine for this.

Kobzol:
Yeah, something like that. Ideally, we should be able to find out from the error if it was actually a timeout though, so we could retry. I remember that octocrab had a bit opaque errors, so if it's not possible to figure it out from the error easily, then we might want to converge to the tokio timeout anyway.

geetanshjuneja:
[...]

PR #267, opened 2025-04-03
Handle Github API timeouts

geetanshjuneja:
Links to #255

### DIFF
--- a/src/github/api/client.rs
+++ b/src/github/api/client.rs
@@ -1,6 +1,6 @@
 use anyhow::Context;
+use octocrab::Octocrab;
 use octocrab::models::{App, Repository};
-use octocrab::{Error, Octocrab};
 use tracing::log;
 
 use crate::bors::event::PullRequestComment;
@@ -10,7 +10,7 @@ use crate::database::RunId;
 use crate::github::api::base_github_html_url;
 use crate::github::api::operations::{MergeError, merge_branches, set_branch_to_commit};
 use crate::github::{CommitSha, GithubRepoName, PullRequest, PullRequestNumber};
-use crate::utils::timing::measure_network_request;
+use crate::utils::timing::{measure_network_request, perform_network_request_with_retry};
 
 /// Provides access to a single app installation (repository) using the GitHub API.
 pub struct GithubRepositoryClient {
@@ -55,7 +55,7 @@ impl GithubRepositoryClient {
     /// Loads repository configuration from a file located at `[CONFIG_FILE_PATH]` in the main
     /// branch.
     pub async fn load_config(&self) -> anyhow::Result<RepositoryConfig> {
-        measure_network_request("load_config", || async {
+        perform_network_request_with_retry("load_config", || async {
             let mut response = self
                 .client
                 .repos(&self.repo_name.owner, &self.repo_name.name)
@@ -83,12 +83,12 @@ impl GithubRepositoryClient {
                     Ok(config)
                 })
         })
-        .await
+        .await?
     }
 
     /// Return the current SHA of the given branch.
     pub async fn get_branch_sha(&self, name: &str) -> anyhow::Result<CommitSha> {
-        measure_network_request("get_branch_sha", || async {
+        perform_network_request_with_retry("get_branch_sha", || async {
             // https://docs.github.com/en/rest/branches/branches?apiVersion=2022-11-28#get-a-branch
             let branch: octocrab::models::repos::Branch = self
                 .client
@@ -100,12 +100,12 @@ impl GithubRepositoryClient {
                 .context("Cannot deserialize branch")?;
             Ok(CommitSha(branch.commit.sha))
         })
-        .await
+        .await?
     }
 
     /// Resolve a pull request from this repository by it's number.
     pub async fn get_pull_request(&self, pr: PullRequestNumber) -> anyhow::Result<PullRequest> {
-        measure_network_request("get_pull_request", || async {
+        perform_network_request_with_retry("get_pull_request", || async {
             let pr = self
                 .client
                 .pulls(self.repository().owner(), self.repository().name())
@@ -116,7 +116,7 @@ impl GithubRepositoryClient {
                 })?;
             Ok(pr.into())
         })
-        .await
+        .await?
     }
 
     /// Post a comment to the pull request with the given number.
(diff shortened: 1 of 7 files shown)


geetanshjuneja:
I've made a change in one place for now to confirm that this is the approach we want to take for handling timeouts.

src/github/api/mod.rs
```diff
@@ -17,6 +18,9 @@ use crate::permissions::TeamApiClient;
 pub mod client;
 pub(crate) mod operations;
 
+// timeout in seconds
+const TIMEOUT: u64 = 10;
```
Kobzol, line 22:
```suggestion
const HTTP_TIMEOUT: Duration = Duration::from_secs(10);
```
And then we don't even need the comment :)

src/github/api/client.rs
```diff
@@ -12,6 +15,31 @@ use crate::github::api::operations::{MergeError, merge_branches, set_branch_to_c
...
+pub async fn auto_retry_github_request<T, F, Fut>(
+    request_name: &str,
+    f: F,
+) -> Result<T, octocrab::Error>
+where
+    F: Fn() -> Fut + Copy,
+    Fut: std::future::Future<Output = Result<T, octocrab::Error>>,
+{
+    let mut res = measure_network_request(request_name, f).await;
+    if let Err(octocrab::Error::Service { source, .. }) = &res {
```
Kobzol, line 27: Hmm, I think that this is too ugly in the end. But mainly, I realized that if octocrab changes its internal representation of the errors, then this will silently stop working, and that's not great.

Let's just use the tokio timeout then (but still keep the octocrab timeout setting, that's good practice IMO), at least we will have more control over it.

I would generalize `measure_network_request`, rename it to something like `perform_network_request`, and pass it a timeout and a number of retries that it should perform when a timeout is encountered. By doing the timeouts in that function, we will be able to provide better log messages about what happens, because currently the log would just contain something like "operation started", and the "operation finished" part would be missing.

geetanshjuneja:
@Kobzol I am not sure how can I check rate limit with each gh request as octocrab doesn't expose response headers. We can check rate limit through its octocrab [rate limit handler](https://docs.rs/octocrab/latest/octocrab/struct.Octocrab.html#method.ratelimit).

Kobzol:
Let's deal only with timeouts for now, then. We probably can't do anything about the rate limit anyway.. But hopefully it shouldn't (ever :crossed_fingers:) be triggered, at least not the per hour one (which should be 5k requests, IIRC).

geetanshjuneja:
I am done with this, you can review whenever you have a moment.

src/utils/timing.rs
```diff
@@ -33,12 +91,23 @@ where
 }
 
 // Measures the duration of a network request and logs it using tracing.
-pub async fn measure_network_request<T, F, Fut>(request_name: &str, f: F) -> T
+pub async fn perform_network_request<T, F, Fut>(
+    request_name: &str,
+    f: F,
+    timeout: Duration,
+    num_retry: u32,
```
Kobzol, line 98:
```suggestion
    retry_count: u32,
```

src/github/api/mod.rs
```diff
@@ -17,6 +18,10 @@ use crate::permissions::TeamApiClient;
 pub mod client;
 pub(crate) mod operations;
 
+pub const TIMEOUT: Duration = Duration::from_secs(10);
+
+pub const NUM_RETRY: u32 = 5;
```
Kobzol, line 23:
```suggestion
pub const DEFAULT_RETRY_COUNT: u32 = 5;
```

src/github/api/mod.rs
```diff
@@ -17,6 +18,10 @@ use crate::permissions::TeamApiClient;
 pub mod client;
 pub(crate) mod operations;
 
+pub const TIMEOUT: Duration = Duration::from_secs(10);
```
Kobzol, line 21:
```suggestion
pub const DEFAULT_REQUEST_TIMEOUT: Duration = Duration::from_secs(10);
```

src/utils/timing.rs
```diff
@@ -23,6 +25,62 @@ where
...
+where
+    F: Fn() -> Fut,
+    Fut: std::future::Future<Output = T>,
+{
+    trace!(operation = operation_name, "Starting operation");
+
+    for attempt in 0..=num_retries {
+        let start = Instant::now();
+
+        match time::timeout(timeout, f()).await {
```
Kobzol, line 44: We might want to have some small back-off here, at least sleep for a second or two.

Kobzol (commented):
The approach looks fine, thanks!

src/github/api/client.rs
```diff
@@ -164,101 +200,116 @@ impl GithubRepositoryClient {
...
-        measure_network_request("get_check_suites_for_commit", || async {
-            #[derive(serde::Deserialize, Debug)]
-            struct CheckSuitePayload {
-                conclusion: Option<String>,
-                head_branch: String,
-            }
+        perform_network_request(
+            "get_check_suites_for_commit",
+            || async {
+                #[derive(serde::Deserialize, Debug)]
```
Kobzol, line 206: Let's move these struct definitions outside the closure, to avoid the unnecessary indentation.

Kobzol (commented):
Thanks! Now that you added the timeouts everywhere, it is a bit noisy having the timeout and retry count at all GitHub call-sites. Since they all use the same values, let's just hardcode `DEFAULT_REQUEST_TIMEOUT` and `DEFAULT_RETRY_COUNT` within `perform_network_request`. We can always add a parametrized version in the future if the need arises.

Kobzol (approved):
Nice, even the indentation and diff got better. Thanks!

Kobzol:
Oh, the test failure showed something interesting. In `GitHubRepositoryClient::cancel_workflows`, we shouldn't retry the whole function, because it actually does several network calls. Let's not retry at all in that function, because it is best-effort anyway.

Seems like retrying is not always the best choice. Please also remove the function to `perform_network_request_with_retry`, or something like that, to make it super obvious that a retry mechanism will be used inside.

geetanshjuneja:
> Please also remove the function to perform_network_request_with_retry, or something like that, to make it super obvious that a retry mechanism will be used inside.

Ok so

geetanshjuneja:
> Oh, the test failure showed something interesting. In `GitHubRepositoryClient::cancel_workflows`, we shouldn't retry the whole function, because it actually does several network calls. Let's not retry at all in that function, because it is best-effort anyway.

Since all the workflows are cancelled in parallel, shouldn't we a have retry on each on them?

Kobzol:
If it was a critical operation, then yes, but cancelling is just a best-effort thing, it doesn't really matter if it fails or not, so we shouldn't block the rest of the bot waiting for it if it takes too long.

Kobzol (approved):
Thanks! Let's try again.

PR #269, opened 2025-04-06
Add database migrations population tests

### LINKED ISSUE #257: Implement a test for checking DB migrations
Recently we had some problems with migrations breaking the deployment of bors (https://kobzol.github.io/rust/2025/03/25/just-write-a-test-for-it.html). I created a basic smoke test against one situation that can cause these issues, but a more general solution is to devise a way to test that a given migration will work even when applied on a *populated* database.

We need to figure out how to get a populated database in tests though. I suppose that we could either commit a sample dataset for each migration, or automatically fill the DB with all possible values based on the column data types.

Sakib25800:
Fixes #257 

This PR adds tests that incrementally populate the database with test data for every migration.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -118,6 +118,10 @@ Make sure to also run `cargo sqlx migrate run` to apply the migrations to the da
     ```console
     $ cargo sqlx migrate run
     ```
+4) Add a test data file to `tests/data/migrations/<timestamp>-<new-migration>.sql`.
+    - The file should contain SQL that inserts some reasonable data into a test database after the migration is applied.
+    The goal is to check that we have a test database with production-like data, so that we can test that applying migrations will not produce errors on a non-empty database.
+    - If it doesn't make sense to add any data to the migration (e.g. if the migration only adds an index), put `-- Empty to satisfy migration tests` into the file.
 
 ### Generate `.sqlx` directory
 ```console
--- a/tests/migrations.rs
+++ b/tests/migrations.rs
@@ -5,10 +5,11 @@ use sqlparser::ast::{
 };
 use sqlparser::dialect::PostgreSqlDialect;
 use sqlparser::parser::Parser;
+use sqlx::{Executor, PgPool};
 use std::collections::{BTreeSet, HashSet};
 use std::ffi::OsStr;
 use std::ops::ControlFlow;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 struct CheckNotNullWithoutDefault {
     error: Option<String>,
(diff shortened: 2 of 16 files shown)


tests/migrations.rs
```diff
@@ -114,3 +115,69 @@ fn check_non_null_column_without_default() {
...
+
+fn get_up_migrations() -> Vec<PathBuf> {
+    let root = env!("CARGO_MANIFEST_DIR");
+    let migrations_dir = PathBuf::from(root).join("migrations");
+
+    let mut migrations = Vec::new();
+    for entry in std::fs::read_dir(migrations_dir).unwrap() {
+        let path = entry.unwrap().path();
+
+        if path.is_file() && path.to_string_lossy().contains(".up.sql") {
```
Kobzol, line 127:
```suggestion
        if path.is_file() && path.ends_with(".up.sql") {
```

tests/migrations.rs
```diff
@@ -114,3 +115,69 @@ fn check_non_null_column_without_default() {
...
+        }
+    }
+
+    // Sort by timestamp to enforce migration order
+    migrations.sort_by(|a, b| a.file_name().unwrap().cmp(b.file_name().unwrap()));
+    migrations
+}
+
+fn get_test_data_path(migration_path: &Path) -> PathBuf {
+    let filename = migration_path.file_name().unwrap().to_str().unwrap();
```
Kobzol, line 138: https://doc.rust-lang.org/std/path/struct.Path.html#method.file_prefix still unstable :(

tests/migrations.rs
```diff
@@ -114,3 +115,69 @@ fn check_non_null_column_without_default() {
...
+        assert!(
+            test_data_path.exists(),
+            "Migration {:?} does not have an associated test data file at {:?}",
+            migration_path,
+            test_data_path
+        );
+    }
+}
+
+#[sqlx::test(migrations = false)]
```
Kobzol, line 162: Please add a brief comment above this function that explains what the function does.

tests/migrations.rs
```diff
@@ -114,3 +115,69 @@ fn check_non_null_column_without_default() {
...
+}
+
+#[test]
+fn check_migrations_have_sample_data() {
+    for migration_path in get_up_migrations() {
+        let test_data_path = get_test_data_path(&migration_path);
+
+        assert!(
+            test_data_path.exists(),
+            "Migration {:?} does not have an associated test data file at {:?}",
```
Kobzol, line 155:
```suggestion
            "Migration {:?} does not have an associated test data file at {:?}. Add a test data file there that fills some test data into the database after that migration is applied.",
```

docs/development.md
```diff
@@ -118,6 +118,7 @@ Make sure to also run `cargo sqlx migrate run` to apply the migrations to the da
     ```console
     $ cargo sqlx migrate run
     ```
+4) Add sample data to `tests/data/migrations/<timestamp>-<new-migration>.sql`.
```
Kobzol, line 121:
```suggestion
4) Add a test data file to `tests/data/migrations/<timestamp>-<new-migration>.sql`.
  - The file should contain SQL that inserts some reasonable data into a test database after the migration is applied. The goal is to check that we have a test database with production-like data, so that we can test that applying migrations will not produce errors on a non-empty database.
  - If it doesn't make sense to add any data to the migration (e.g. if the migration only adds an index), put `-- Empty to satisfy migration tests` into the file.
```

tests/migrations.rs
```diff
@@ -114,3 +115,69 @@ fn check_non_null_column_without_default() {
...
+    for migration_path in migrations {
+        let migration_sql = std::fs::read_to_string(&migration_path)
+            .unwrap_or_else(|_| panic!("Failed to read migration file: {:?}", migration_path));
+
+        pool.execute(&*migration_sql)
+            .await
+            .unwrap_or_else(|e| panic!("Failed to apply migration {:?}: {}", migration_path, e));
+
+        let test_data_path = get_test_data_path(&migration_path);
+        if let Ok(test_data) = std::fs::read_to_string(&test_data_path) {
```
Kobzol, line 175: I know that we have the separate test for checking that this file exists, but we should still panic here if it doesn't. Otherwise the test could e.g. skip one migration and then report an unrelated error, which could be confusing. So let's just `expect()` here.

tests/migrations.rs
```diff
@@ -114,3 +115,69 @@ fn check_non_null_column_without_default() {
...
+            pool.execute(&*test_data).await.unwrap_or_else(|e| {
+                panic!("Failed to apply test data {:?}: {}", test_data_path, e)
```
Kobzol, line 177:
```suggestion
                panic!("Failed to apply migration test data {:?}: {}", test_data_path, e)
```

Kobzol (commented):
This looks awesome :heart: And isn't even so complicated. I first wanted to apply the data *before* the migration, but I agree that it makes more sense to do it after the migration has been applied.

Sakib25800:
Fixed all issues. Nice and easy PR unlike the other one 😮‍💨.

tests/migrations.rs
Kobzol: Sorry, this was a bad advice, I didn't realize that `ends_with` only checks the last component, so this doesn't work. But at least it helped me find a loophole in the tests - right now all tests are green even if you remove all test data files, but the tests don't actually find any migrations :sweat_smile: Let's add this:

```rust
#[test]
fn check_migrations_have_sample_data() {
    let migrations = get_up_migrations();
    assert!(!migrations.is_empty());
    for migration_path in migrations {
```

Kobzol:
The `ends_path` check was wrong, sorry (left a comment). After you fix it, and check that the tests actually fail if files are missing :laughing: Please squash the commits and I'll merge it.

Kobzol:
Looks like too much stuff was pushed :)

Sakib25800:
Whoops, my branch wasn't up to date. 

Squashed.

Kobzol (approved):
Thank you! :)

PR #276, opened 2025-04-27
Synchorize pr state periodically

### LINKED ISSUE #266: Synchronize state after (re)start and periodically
One important piece of state that bors should have up-to-date is the queue of (open) PRs. However, when a PR is closed/merged/converted to draft while the bors is offline or being redeployed, or a PR status change webhook is lost, the state in the bors DB can become outdated.

We do update the PR state after every comment on that PR that contains a known bors command, but it would be nice to also update the PRs proactively.

When the bot starts, or in the periodic refresh handler, we could update all opened PRs in our DB, and check if they are still opened. However, doing this one-by-one would be very slow and create many GitHub requests. What we could do instead is batch load PRs with a specific state (opened) from GH, which allows loading 100 PRs at once, and also batch load opened PRs from our DB. Then, when we find a PR where the opened state on GH does not match the state in DB, we update the DB state. See [here](https://github.com/rust-lang/triagebot/pull/1918/commits/89153babdc48ec873f526f03b6760267127ab32c) for an example of how to load that data.

geetanshjuneja:
Closes #266.

### DIFF
--- a/src/bin/bors.rs
+++ b/src/bin/bors.rs
@@ -29,6 +29,9 @@ const CANCEL_TIMED_OUT_BUILDS_INTERVAL: Duration = Duration::from_secs(60 * 5);
 /// How often should the bot reload the mergeability status of PRs?
 const MERGEABILITY_STATUS_INTERVAL: Duration = Duration::from_secs(60 * 10);
 
+/// How often should the bot synchronize PR state.
+const PR_STATE_PERIODIC_REFRESH: Duration = Duration::from_secs(60 * 10);
+
 #[derive(clap::Parser)]
 struct Opts {
     /// Github App ID.
@@ -141,6 +144,7 @@ fn try_main(opts: Opts) -> anyhow::Result<()> {
         let mut permissions_refresh = make_interval(PERMISSIONS_REFRESH_INTERVAL);
         let mut cancel_builds_refresh = make_interval(CANCEL_TIMED_OUT_BUILDS_INTERVAL);
         let mut mergeability_status_refresh = make_interval(MERGEABILITY_STATUS_INTERVAL);
+        let mut prs_interval = make_interval(PR_STATE_PERIODIC_REFRESH);
         loop {
             tokio::select! {
                 _ = config_refresh.tick() => {
@@ -154,6 +158,9 @@ fn try_main(opts: Opts) -> anyhow::Result<()> {
                 }
                 _ = mergeability_status_refresh.tick() => {
                     refresh_tx.send(BorsGlobalEvent::RefreshPullRequestMergeability).await?;
+                }
+                _ = prs_interval.tick() => {
+                    refresh_tx.send(BorsGlobalEvent::RefreshPullRequestState).await?;
                 }
             }
         }
--- a/src/bors/handlers/refresh.rs
+++ b/src/bors/handlers/refresh.rs
@@ -3,8 +3,10 @@ use std::time::Duration;
 
 use anyhow::Context;
 use chrono::{DateTime, Utc};
+use std::collections::BTreeMap;
 
 use crate::bors::Comment;
+use crate::bors::PullRequestStatus;
 use crate::bors::RepositoryState;
 use crate::bors::handlers::trybuild::cancel_build_workflows;
 use crate::bors::mergeable_queue::MergeableQueueSender;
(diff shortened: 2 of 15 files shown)


src/github/api/client.rs
```diff
@@ -319,6 +320,27 @@ impl GithubRepositoryClient {
         run_ids.map(|workflow_id| self.get_workflow_url(workflow_id))
     }
 
+    pub async fn fetch_prs(&self) -> anyhow::Result<Vec<PullRequest>> {
```
Kobzol, line 323: This will download all PRs in the repo, which is more than 80 thousand PRs on rust-lang/rust. We should only download the open/draft (if it's possible to filter for both in one API call) PRs.

Then when we find a PR that is not amongst the open ones, we assume that it is closed (in theory it could also be merged, but bors does the merging, so it should not miss that event).

src/bin/bors.rs
```diff
@@ -18,6 +18,8 @@ use tracing_subscriber::filter::EnvFilter;
 
 /// How often should the bot check DB state, e.g. for handling timeouts.
 const PERIODIC_REFRESH: Duration = Duration::from_secs(120);
+/// How often should the bot synchronize PR state.
+const PR_STATE_PERIODIC_REFRESH: Duration = Duration::from_secs(60 * 60);
```
Kobzol, line 22: I'm not sure if we need to separate this from the normal periodic refresh, because PRs are also *per repository*. So it's weird to have a separate event for it.

That being said, it should have a different update frequency than the refresh of configs and permissions. We could have a single outside trigger for the refresh logic, that would "tick" e.g. every minute, and the refresh code could then remember the most recent timestamps for the individual refresh logic handlers. But this would be a bit annoying in the tests, as we'd need to override the period between refreshes, rather than just triggering the refresh with a mocked event. Or maybe we could do something like `Event::RefreshPrs { force: bool }` to enable us to force the refresh in tests.

I will split the refresh handlers into individual events in a separate PR, and then we'll rebase this one on top of that.

src/tests/mocks/bors.rs
```diff
@@ -447,22 +472,25 @@ impl BorsTester {
         &mut self,
         repo_name: GithubRepoName,
         pr_number: u64,
+        send_webhook: bool,
```
Kobzol, line 475: Why should this ever be `false`?

src/bors/handlers/refresh.rs
```diff
@@ -28,6 +28,36 @@ pub async fn refresh_repository(
...
 
+pub async fn sync_prs_state(repo: Arc<RepositoryState>, db: Arc<PgDbClient>) -> anyhow::Result<()> {
+    let repo = repo.as_ref();
+    let db = db.as_ref();
+    let repo_name = repo.repository();
+    // load prs from github
+    let prs = repo.client.fetch_prs().await?;
+    let pr_to_status = db.get_all_pull_request_status(repo_name).await?;
+
+    for pr in prs {
```
Kobzol, line 39: As noted in my other comment, and in the original issue, we cannot simply download all PRs from GitHub, there's too many. We only have to fetch the open ones and synchronize smartly based on that.

Kobzol (commented):
Thank you! Left a few comments. I'll take a look at refactoring the refresh handlers, to see if it makes sense.

src/tests/mocks/bors.rs
geetanshjuneja: Will fix test `refresh_pr_with_status_closed` in which a pr's status is closed in gh but not in db.

src/github/api/client.rs
geetanshjuneja: It could also happen that open and close webhook of a pr is lost and therefore this pr has no trace of it in db. Also while refreshing pr state we will download open/draft prs, so still there is no trace of it in the db.

src/tests/mocks/bors.rs
Kobzol: Then let's just modify the mock GH state in the test directly, to avoid having to pass true/false everywhere.

src/github/api/client.rs
Kobzol: That's fine. If we miss some PR that was opened and then closed while bors wasn't running, it doesn't really matter, we don't care about closed PRs. We are only interested in open PRs waiting to be merged.
These situations should be quite rare anyway.

geetanshjuneja: I need some help in implementation of this logic. So we have two vectors `nonclosed_gh_pull_request` and `nonclosed_db_pull_requests`.
For prs that are in `nonclosed_gh_pull_request` and not in `nonclosed_db_pull_requests`. I have to upsert prs in the db.
For prs that are in `nonclosed_db_pull_request` and not in `nonclosed_gh_pull_requests`. I have to mark their state close in db.

I was trying to implement this using `BTreeset` and use set difference. Is this fine or anything better than this is there.

src/tests/mocks/bors.rs
geetanshjuneja: Can you elaborate more on this? Which specific struct in test you are talking about?

src/bors/handlers/refresh.rs
```diff
@@ -97,6 +99,71 @@ pub async fn reload_repository_config(repo: Arc<RepositoryState>) -> anyhow::Res
...
+                &gh_pr.base.name,
+                gh_pr.mergeable_state.clone().into(),
+                &gh_pr.status,
+            )
+            .await?;
+        }
+    }
+    // PRs that are closed in GitHub but not in the DB. In theory PR could also be merged
+    // but bors does the merging so it should not happen.
+    for pr_num in nonclosed_db_prs_num.keys() {
```
Kobzol, line 152: There is a potential race condition here, where if the PR *just* got merged on GH in between the GH API and DB calls, we would mark it as closed even though it was merged. But let's not deal with that now.

src/bors/handlers/refresh.rs
```diff
@@ -139,6 +208,15 @@ mod tests {
         .await;
     }
 
+    #[sqlx::test]
+    async fn refresh_without_out_of_sync_prs(pool: sqlx::PgPool) {
```
Kobzol, line 211: `refresh_pr_state`

src/tests/mocks/bors.rs
Kobzol: I mean to either copy-paste the logic for creating a PR in the mock GH database into the test, or creating a separate function that will only create the PR on GH, but not send the webhook. It's quite hard to read the `false, false` etc. parameters to `open_pr` in tests.

Kobzol: Actually, I think that we could solve this in a more general way, by creating something like this:
```rust
tester.block_webhooks(async move {
    // While this future runs, no webhooks will be sent
});
```

src/github/api/client.rs
Kobzol: You implemented the logic well, thank you! :)

src/tests/mocks/bors.rs
geetanshjuneja: But sometimes in tests I need only open webhook to work.

Kobzol:
```rust
let pr = tester.block_webhooks(async move {
    // While this future runs, no webhooks will be sent
   tester.open_pr(default_repo_name(), false).await
}).await?;
tester
    .close_pr(default_repo_name(), pr.number.0)
    .await?;
```

geetanshjuneja: But this require the `BorsTester` to be cloned.

Kobzol: Oh, right, this was a bit tricky until recently. Luckily, we now finally have async closures! :) I pushed an implementation of `with_blocked_webhooks`.

geetanshjuneja: Thanks for the help. I didn't knew about async closures and now will about it.

src/bors/handlers/refresh.rs
geetanshjuneja: If the PR just got merged b/w GH API and DB calls, then it would send a webhook which will be processed after it changing the status again to merged.

src/bors/handlers/pr_events.rs
```diff
@@ -466,7 +466,9 @@ mod tests {
                     pr.pr_status == PullRequestStatus::Open
                 })
                 .await?;
-            tester.merge_pr(default_repo_name(), pr.number.0).await?;
+            tester
+                .merge_pr(default_repo_name(), pr.number.0, true)
```
Kobzol, line 470: I missed this function, let's also remove `send_webhook: bool,` from its parameters, now that we have a better way of blocking webhooks.

src/bors/handlers/refresh.rs
Kobzol: I was thinking something like this:
Bors merges a PR. Before setting its state to merged, this refresh handler runs and loads the PR as open from GH. Then bors sets the local state as merged, but it gets overwritten by this refresh handler. This is possible because the refresh handler can run concurrently with command handlers.

That being said, this is incredibly unlikely to happen, and even if we did, the state should be correctly synchronized during the next refresh, or during the next comment sent on the PR, so it should be fine.

Kobzol (commented):
Thanks! Left one nit, once you remove the `send_webhook` parameter, please squash the commits and we can merge this.

Kobzol (approved):
Thank you! :)

PR #281, opened 2025-05-19
Bootstrap website

Sakib25800:
This PR adds the website's home page, where it displays the repos. It also displays "[TREECLOSED]" is the tree is closed for the corresponding repo. 

The HTML is essentially the same as it was 11 years ago with slight differences:
- Homu -> Bors
- The section on "Customizing the Queue's Contents" has not been added as that will be added when the /queue is added and I don't know if all of the mentioned features are going to be supported
- The commands shown in the home page do not completely match as either some are not yet handled by bors just yet (e.g. `force` command) or some have been removed (e.g. `r=<name>`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,10 +35,12 @@ arc-swap = "1"
 hyper = "1"
 axum = "0.8"
 tower = { version = "0.5", features = ["limit"] }
+tower-http = { version = "0.6.4", features = ["catch-panic"] }
 jsonwebtoken = "9"
 url = "2"
 reqwest = { version = "0.12", features = ["json", "rustls-tls"], default-features = false }
 http = "1"
+askama = "0.14.0"
 
 # Cryptography
 sha2 = "0.10"
@@ -52,6 +54,7 @@ sqlx = { version = "0.8", features = ["runtime-tokio", "tls-rustls", "postgres",
 # Time
 chrono = "0.4"
 
+# Utilities
 itertools = "0.14"
 
 # Text processing
--- a/src/github/server.rs
+++ b/src/github/server.rs
@@ -3,30 +3,40 @@ use crate::bors::mergeable_queue::{
     MergeableQueueReceiver, MergeableQueueSender, create_mergeable_queue,
     handle_mergeable_queue_item,
 };
-use crate::bors::{BorsContext, handle_bors_global_event, handle_bors_repository_event};
+use crate::bors::{
+    BorsContext, RepositoryState, handle_bors_global_event, handle_bors_repository_event,
+};
 use crate::github::webhook::GitHubWebhook;
 use crate::github::webhook::WebhookSecret;
-use crate::{BorsGlobalEvent, BorsRepositoryEvent, TeamApiClient};
+use crate::templates::{HtmlTemplate, IndexTemplate, RepositoryView};
+use crate::{BorsGlobalEvent, BorsRepositoryEvent, PgDbClient, TeamApiClient};
 
 use anyhow::Error;
 use axum::Router;
 use axum::extract::State;
 use axum::http::StatusCode;
-use axum::response::IntoResponse;
+use axum::response::{IntoResponse, Response};
 use axum::routing::{get, post};
 use octocrab::Octocrab;
+use std::any::Any;
+use std::collections::HashMap;
 use std::future::Future;
 use std::pin::Pin;
 use std::sync::Arc;
 use tokio::sync::mpsc;
 use tower::limit::ConcurrencyLimitLayer;
+use tower_http::catch_panic::CatchPanicLayer;
 use tracing::{Instrument, Span};
+
+use super::GithubRepoName;
 
 /// Shared server state for all axum handlers.
 pub struct ServerState {
     repository_event_queue: mpsc::Sender<BorsRepositoryEvent>,
     global_event_queue: mpsc::Sender<BorsGlobalEvent>,
     webhook_secret: WebhookSecret,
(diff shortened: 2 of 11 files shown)


src/templates.rs
```diff
@@ -1,4 +1,26 @@
...
+use axum::response::{Html, Response, IntoResponse};
+use http::StatusCode;
+
+pub struct HtmlTemplate<T>(pub T);
+
+impl<T> IntoResponse for HtmlTemplate<T>
+where
+    T: Template
+{
+    fn into_response(self) -> Response {
```
Kobzol, line 11: Now that we have a web frontend, which can potentially fail in all sorts of interesting ways, it would be nice to add https://docs.rs/tower-http/latest/tower_http/catch_panic/index.html to the server, so that we can return a proper 500 error on panics, and also log the error.

src/github/server.rs
```diff
@@ -73,26 +70,13 @@ async fn health_handler() -> impl IntoResponse {
...
-            Err(error) => {
-                tracing::error!("Failed to load repository {repo} from database: {error:?}");
-                continue;
-            }
-        };
-
-        repos.push(RepositoryInfo { name: repo.name.clone(), treeclosed: repo_db.is_treeclosed() });
-    }
-
     HtmlTemplate(IndexTemplate {
```
Kobzol, line 75: I would rather load the ground truth from the DB rather than to duplicate the state in memory, where it can potentially diverge. Any specific reason for this commit?

src/templates.rs
```diff
@@ -12,10 +12,11 @@ where
         match self.0.render() {
             Ok(html) => Html(html).into_response(),
             Err(error) => {
-                tracing::error!("Failed to render template: {error:?}");
+                let message = format!("Failed to render template: {error:?}");
+                tracing::error!(message);
```
Kobzol, line 16: `tracing::error!("{message}")`, otherwise this would be unnecessarily rendered as `message=<message>`.

src/bors/handlers/refresh.rs
```diff
@@ -349,6 +349,9 @@ timeout = 3600
     async fn refresh_pr_with_status_closed(pool: sqlx::PgPool) {
         run_test(pool, |mut tester| async move {
             let pr = tester.open_pr(default_repo_name(), false).await?;
+            tester
```
Kobzol, line 352: Could you please move this change to a separate PR? Thanks!

Kobzol (commented):
I was shocked by how similar the page looked to the homu homepage :laughing: Good job! This PR is already quite big, so let's implement further improvements in follow-up PRs. Left a bunch of comments.

Cargo.toml
```diff
@@ -39,6 +39,8 @@ jsonwebtoken = "9"
 url = "2"
 reqwest = { version = "0.12", features = ["json", "rustls-tls"], default-features = false }
 http = "1"
+askama = { git = "https://github.com/djc/askama", features = ["with-axum"] }
```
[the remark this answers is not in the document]
Sakib25800: Fixed. 

Accidentally left it over.

Cargo.toml
```diff
@@ -39,6 +39,8 @@ jsonwebtoken = "9"
...
+askama_axum = { git = "https://github.com/djc/askama" }
```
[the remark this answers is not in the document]
Sakib25800: Fixed.

src/github/server.rs
```diff
@@ -71,13 +74,28 @@ async fn health_handler() -> impl IntoResponse {
...
-        repos: state.repositories.iter().map(|(repo, _)| {
-            RepositoryInfo {
-                name: repo.name.to_string()
+    let mut repos = Vec::with_capacity(state.repositories.len());
+
+    for (repo, _) in &state.repositories {
+        let repo_db = match state.db.repo_db(&repo).await {
+            Ok(Some(repo))  => repo,
+            Ok(None) => {
+                tracing::error!("Repository {repo} not found in database");
```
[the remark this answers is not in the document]
Sakib25800: Fixed.

src/templates.rs
Sakib25800: Added.

src/github/server.rs
Sakib25800: Homu used internal state and synced the in memory repos like I did. 

I chose to do it because I thought that it would save on a database query every page load. But I'm sure the database can handle a few more queries.

src/bors/handlers/refresh.rs
Sakib25800: I removed the line I added and it now works, but failed before? So I guess this test is just a little flaky.

Sakib25800:
TBH this page (and the others) could probably do with a super tiny design change so that it matches what the other rust sites look like. But of course, that's for later.

src/bors/handlers/refresh.rs
Kobzol: In that case we should still fix that (in a separate PR), if it is flaky.

src/github/server.rs
Kobzol: Yeah I think that should be fine. If the DB can't handle to do one repo query per page load, then we have bigger problems :laughing:

Kobzol:
Looks great! Could you please squash some of the changes into a smaller number of commits?

src/bors/handlers/refresh.rs
Sakib25800: Looks like the flaky test has failed here 😞.

Kobzol (approved):
Thank you, great work :heart:

PR #283, opened 2025-05-21
Fix flaky test `refresh_pr_with_status_closed`

Sakib25800:
This test is flaky.


Kobzol (approved):
Thanks! I wonder if perhaps we should do this check automatically in `open_pr`, but this is good enough for now.

Sakib25800:
I don't think we need to because even though most of the tests utilise `.wait_for_pr(...)` right after `.open_pr(...)`, they all do slightly different checks which make it clear what the test is actually focusing on e.g. a test on `open_and_merge_pr` checks for `pr_status` after opening. 

It makes it clear what state changes we want to see in the tests.

PR #286, opened 2025-05-29
update postgres in tests to 16.9

marcoieni:
align with what we use in prod


Kobzol (approved):
Thanks! That's quite nice indeed.

Issue #289, opened 2025-05-30
suggest how to cancel job


marcoieni:
after launching the try build, sometimes people realize they made a mistake and they want to cancel the job.

What if we add to the current comment body:

> to cancel the try build, run the command`try cancel`

![Image](https://github.com/user-attachments/assets/22cbcea5-5e94-421f-9ea4-7fe9fd49121e)

PR #290, opened 2025-05-30
suggest how to cancel try job

marcoieni:
close https://github.com/rust-lang/bors/issues/289

### DIFF
--- a/src/bors/handlers/trybuild.rs
+++ b/src/bors/handlers/trybuild.rs
@@ -46,6 +46,7 @@ pub(super) async fn command_try_build(
     author: &GithubUser,
     parent: Option<Parent>,
     jobs: Vec<String>,
+    command_prefix: &str,
 ) -> anyhow::Result<()> {
     let repo = repo.as_ref();
     if !has_permission(repo, author, pr, &db, PermissionType::Try).await? {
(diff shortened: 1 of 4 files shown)


src/bors/handlers/trybuild.rs
```diff
@@ -301,7 +301,9 @@ fn auto_merge_commit_message(
 
 fn trying_build_comment(head_sha: &CommitSha, merge_sha: &CommitSha) -> Comment {
     Comment::new(format!(
-        ":hourglass: Trying commit {head_sha} with merge {merge_sha}…"
+        ":hourglass: Trying commit {head_sha} with merge {merge_sha}…
+
+To cancel the try build, run the command `try cancel`."
```
Kobzol, line 306: The cancel command example should contain the name of the bot. It's not easily available now though, we would have to store the bot name into `BorsContext`, and then propagate the context all the way to the function that calls `trying_build_comment`, and `trying_build_comment` would then receive the bot name.

Kobzol (approved):
Bors doesn't yet know how to skip commands in backticks, so let's hope that the "I don't run commands contained in comments sent by myself" logic works :joy: Thanks!

Kobzol:
At that time we definitely have to ignore backticks properly, I'll work on it this week.

PR #291, opened 2025-05-31
Add queue page

Sakib25800:
This PR does the following:
- Adds `/queue/<repo_name>` page
  - Displays PR statistics 
  - Displays queue in table with columns: number, status, mergeable, approved by, priority, rollup
  - The following columns have not been added (will make PRs for each):
    - Title (as we do not track this)
    - Head ref (same reason as above) 
    - Assignee (same reason as above) 
- Redirects homepage to `/queue/rust`
- Moves original homepage to new `/help` page


templates/index.html
```diff
@@ -63,8 +64,9 @@ <h3>Commands</h3>
         posted as comments on the PR they refer to. Comments may include
         multiple commands. Bors will only listen to official reviewers that it
         is configured to listen to. A comment must mention the GitHub account
-        bors is configured to use. (e.g. for the Rust project this is <code>@bors</code>)
-        Note that bors will only recognize comments in <i>open</i> PRs.
+        bors is configured to use. (e.g. for the Rust project this is
+        <code>@bors</code>) Note that bors will only recognize comments in
```
Kobzol: This is not currently the case with bors, right? I think it will happily run commands even in closed/merged PRs right now.

src/database/operations.rs
```diff
@@ -789,6 +789,40 @@ pub(crate) async fn get_repository(
...
 
+/// Retrieves a repository by its name (without owner) from the database.
+/// This searches for repositories where the name field ends with "/{repo_name}".
+/// Returns the first match found.
+pub(crate) async fn get_repository_by_name(
+    executor: impl PgExecutor<'_>,
+    repo_name: &str,
+) -> anyhow::Result<Option<RepoModel>> {
+    measure_db_query("get_repository_by_name", || async {
+        let search_pattern = format!("%/{}", repo_name);
```
Kobzol: nit: `format!("%/{repo_name}")`

templates/queue.html
```diff
@@ -0,0 +1,11 @@
+<!doctype html>
```
Kobzol: Could you please create a [base template](https://askama.readthedocs.io/en/stable/template_syntax.html#base-template) to share the boilerplate HTML meta/head fields? We only need to modify title and body per page, I think.

src/github/server.rs
```diff
@@ -126,11 +126,13 @@ async fn queue_handler(
         }
     };
     let stats = state.db.get_pr_statistics(&repo.name).await?;
+    let prs = state.db.get_nonclosed_pull_requests(&repo.name).await?;
```
Kobzol, line 134: You made a nice SQL query for the PR stats :grin: But since we fetch all the nonclosed PRs from the DB anyway, I think that it would be simpler to avoid that query and just compute the statistics in Rust, from `prs` directly.

src/database/mod.rs
```diff
@@ -354,6 +354,20 @@ impl TreeState {
...
     }
+
+    pub fn priority(&self) -> Option<u32> {
+        match self {
+            TreeState::Closed { priority, .. } => Some(*priority),
+            TreeState::Open => None,
+        }
+    }
+
+    pub fn src(&self) -> Option<&str> {
```
Kobzol: Maybe `comment_source`? To make it a bit more obvious.

templates/queue.html
```diff
@@ -3,10 +3,15 @@
...
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Bors queue - {{ repo_name }}</title>
+    <title>Bors queue - {{ repo_name }} {% if tree_state.is_closed() %} [TREECLOSED] {% endif %}</title>
   </head>
   <body>
-    <h1>Bors queue - <a href="{{ repo_url }}" target="_blank">{{ repo_name }}</a></h1>
+    <h1>
+        Bors queue - <a href="{{ repo_url }}" target="_blank">{{ repo_name }}</a>
+        {% if tree_state.is_closed() %}
```
Kobzol: You could use if let (https://askama.readthedocs.io/en/stable/template_syntax.html#if-let) to extract the priority, and display `TREECLOSED` only if the priority is present, to avoid the `unwrap()`.

src/database/mod.rs
```diff
@@ -251,6 +251,18 @@ pub enum BuildStatus {
     Timeouted,
 }
 
+impl Display for BuildStatus {
```
Kobzol, line 254: This is fine for now (keep it like it is), but I'd like to point out that the presentation of the build status should ideally be implemented in each presentation layer (currently we only have one - the website). It's possible that we might want to render it in different ways at multiple places, and probably the logic for deciding the visualization shouldn't live in the database module.

It can be done by creating a function in the website module that renders the status, or creating some helper presentation wrapper (e.g. `BuildStatusTableDisplay`) that would wrap the status and implement `Display` itself.

Implementing `Display` on the core "domain" types is a bit of an anti-pattern, even though it's quite tempting to do, and we already did it for other structs (but there we kind of abuse `Display` to serialize the structs form/into the DB :laughing:).

src/github/server.rs
```diff
@@ -106,6 +110,24 @@ async fn index_handler(State(state): State<ServerStateRef>) -> impl IntoResponse
     HtmlTemplate(IndexTemplate { repos })
 }
 
+async fn queue_handler(
+    Path(repo_name): Path<String>,
+    State(state): State<ServerStateRef>,
+) -> Result<impl IntoResponse, AppError> {
+    let repo = match state.db.repo_by_name(&repo_name).await? {
```
Kobzol, line 123: This should fetch the repo from the app state (repositories where the GitHub app is installed), not from the DB. Because right now when you open bors with an empty DB, and it is installed on repository X, it will display "repository not found", because that repo doesn't yet have any entry in the DB (it is currently only created when the tree is closed, or some other similar operation).

Or, alternatively (and I think that this would be even better, because then we wouldn't need to consider the edge case where a known repo is not in the DB), when the bot starts and fetches the repositories where it is installed, it should eagerly upsert all the found repos into the DB. Could you please do that in a separate PR? Thank you!

src/github/server.rs
```diff
@@ -103,7 +112,31 @@ async fn index_handler(State(state): State<ServerStateRef>) -> impl IntoResponse
...
+}
+
+async fn queue_handler(
+    Path(repo_name): Path<String>,
+    State(state): State<ServerStateRef>,
+) -> Result<impl IntoResponse, AppError> {
+    let repo = match state.db.repo_by_name(&repo_name).await? {
+        Some(repo) => repo,
+        None => {
+            return Ok((StatusCode::NOT_FOUND, "Repository {repo_name} not found").into_response());
```
Kobzol:
```suggestion
            return Ok((StatusCode::NOT_FOUND, format!("Repository {repo_name} not found")).into_response());
```
Otherwise the name of the repo isn't interpolated.

templates/index.html
Sakib25800: Ahh yeah, right now it will. 

Homu deletes PRs from in-memory state and the database once merged and then skips webhook calls for these since they no longer exist to it. 

I'm assuming we should do something similar? In which case that statement _will_ become true but right now it's not.

src/github/server.rs
Sakib25800: Working on it.

templates/index.html
Kobzol: I'm not sure if we really need to delete PRs from the database, but in the long term we might implement something like that to reduce the DB storage, yeah (although I suspect it will be so small anyway that it doesn't really matter).

src/github/server.rs
```diff
@@ -125,15 +126,41 @@ async fn queue_handler(
...
-    let stats = state.db.get_pr_statistics(&repo.name).await?;
 
     let prs = state.db.get_nonclosed_pull_requests(&repo.name).await?;
 
+    let (approved_count, rolled_up_count, failed_count) =
+        prs.iter()
+            .fold((0, 0, 0), |(approved, rolled_up, failed), pr| {
+                let is_approved = if pr.is_approved() { 1 } else { 0 };
+                let is_rolled_up = if pr.rollup.is_some() { 1 } else { 0 };
+                let is_failed = if pr
```
Kobzol: This seems a bit weird, to calculate the failed state from the try build. How does homu do it? I think that this should be "PRs that are approved, but their auto (merge attempt) failed". Which means that we can't yet compute the failed count, as we don't have merge results.

src/github/server.rs
```diff
@@ -125,15 +126,41 @@ async fn queue_handler(
...
     };
-    let stats = state.db.get_pr_statistics(&repo.name).await?;
 
     let prs = state.db.get_nonclosed_pull_requests(&repo.name).await?;
 
+    let (approved_count, rolled_up_count, failed_count) =
+        prs.iter()
+            .fold((0, 0, 0), |(approved, rolled_up, failed), pr| {
+                let is_approved = if pr.is_approved() { 1 } else { 0 };
+                let is_rolled_up = if pr.rollup.is_some() { 1 } else { 0 };
```
Kobzol: We should probably not count PRs that have `pr.rollup` equal to `Some(RollupMode::Never)` - I assume that's what homu does?

src/github/server.rs
Sakib25800: Yeah, we can't compute it just yet. I got confused with the two types of builds. 

I've removed it for now and added a `TODO`.

src/github/server.rs
Sakib25800: Fixed.

We should only count `RollupMode::Always` PRs - Homu excludes `Never`/`Maybe`/`Iffy` from the rollup statistic.

Sakib25800:
Will merge that PR first then come back to this.

src/github/server.rs
Sakib25800: I guess this can be resolved.

Sakib25800:
Just realised that they're two completely separate PRs now. 

I was going to fetch repos from state rather than the database but realised that we need `TreeState` anyway, so we have to fetch from the database regardless. 

Squashed.

PR #298, opened 2025-06-03
Upsert repos into database on load

Sakib25800:
See [#291](https://github.com/rust-lang/bors/pull/291#discussion_r2120396579) for context.

### DIFF
--- a/src/bin/bors.rs
+++ b/src/bin/bors.rs
@@ -123,8 +123,6 @@ fn try_main(opts: Opts) -> anyhow::Result<()> {
         runtime.block_on(async {
             if let Err(error) = db.insert_repo_if_not_exists(&name, TreeState::Open).await {
                 tracing::warn!("Failed to insert repository {name}: {error:?}");
-            } else {
-                tracing::debug!("Repository {name} inserted into database");
             }
         });
 


src/bin/bors.rs
```diff
@@ -119,6 +119,15 @@ fn try_main(opts: Opts) -> anyhow::Result<()> {
                 ));
             }
         };
+
+        runtime.block_on(async {
+            if let Err(error) = db.upsert_repository(&name, TreeState::Open).await {
```
Kobzol: Hmm, this doesn't really work if the repo was already in the DB, in that case its tree could have been closed, and we should not overwrite it when the bot restarts. So probably we should create a new function that inserts the repository only if it did not already exist in the DB before.

Sakib25800: Sorry, yeah it should be that.

src/bin/bors.rs
```diff
@@ -119,6 +119,15 @@ fn try_main(opts: Opts) -> anyhow::Result<()> {
                 ));
             }
         };
+
+        runtime.block_on(async {
+            if let Err(error) = db.insert_repo_if_not_exists(&name, TreeState::Open).await {
+                tracing::warn!("Failed to insert repository {name}: {error:?}");
+            } else {
```
Kobzol: I think that we can drop the `else` branch, if the repo existed before, it wasn't actually inserted into the DB, so the log is a bit misleading.

src/bin/bors.rs
Kobzol: No need to apologize :) Thanks!

Kobzol (approved):
Thanks!

Issue #295, opened 2025-06-01
New bors looks within markdown code blocks


jieyouxu:
It's very useful to be able to *not* trigger bors (old or new) when you want to show an bors invocation for whatever reason. In old bors, I believe there's special logic to prevent old bors to interpret `@bors try` or similar invocations within inline code blocks or code blocks. However, this doesn't seem to be carried forward to new bors.

### Reproduction steps

- User writes
  > Reminder (for myself): if try job fails, use `@bors2 try` to try out new bors
- Old bors: nothing
- New bors: tries to interpret as ``@bors2 try` to try out new bors`` and fails because ``try``` is not a valid command. See https://github.com/rust-lang/rust/pull/141856#issuecomment-2927257411.

### Expected behavior

Ignored, because user is most likely trying to demonstrate the invocation and not actually run it.

### Actual behavior (new bors)

New bors tries to run it.

Urgau:
We have module specifically dedicated to figuring out which part of the markdown we should ignore. bors should follow it.

https://github.com/rust-lang/triagebot/blob/7044e5449aaf29eea939bc7db81c63b7c872b9b8/parser/src/ignore_block.rs#L10-L61

jieyouxu:
Maybe worth pulling that out into some internal lib

Kobzol:
I would love to have a shared monorepo for all our bots 😆 But that's probably not going to happen anytime soon. This is definitely a bug that should be fixed, but I wonder if we need such a heavyweight solution as triagebot uses.

I would be personally fine with saying that all bors commands have to start at the beginning of a line (modulo whitespace), otherwise they won't work. That would solve this issue on its own.

What do you think?

Urgau:
Many people (including me) uses: `r? Urgau @bors r+`, would be good to not break it.

PR #300, opened 2025-06-04
Parse PR comments as Markdown and ignore code and links

Kobzol:
This PR changes command parsing in bors to ignore commands in Markdown code blocks, backticks, links, images, etc. Because `pulldown_cmark` uses `CowStr` for the text, I had to get rid of the "zero-copy" errors in the parser, but that way way overkill anyway :laughing:

Fixes: https://github.com/rust-lang/bors/issues/295

### DIFF
--- a/src/bors/command/parser.rs
+++ b/src/bors/command/parser.rs
@@ -69,7 +69,7 @@ fn extract_text_segments(text: &str) -> Vec<CowStr> {
 
     for event in md_parser.into_iter() {
         match event {
-            Event::Text(text) => {
+            Event::Text(text) | Event::Html(text) => {
                 // Only consider commands in raw text outside of wrapping elements
                 if stack.is_empty() {
                     segments.push(text);
@@ -79,7 +79,6 @@ fn extract_text_segments(text: &str) -> Vec<CowStr> {
                 // Ignore content in the following wrapping elements
                 Tag::BlockQuote(_)
                 | Tag::CodeBlock(_)
-                | Tag::HtmlBlock
                 | Tag::Link { .. }
                 | Tag::Heading { .. }
                 | Tag::Image { .. } => {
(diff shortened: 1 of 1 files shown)


Kobzol:
(Just looking for a sanity check that the list of ignored Markdown elements makes sense).

src/bors/command/parser.rs
```diff
@@ -1161,6 +1208,36 @@ line two
...
+@bors try
+```
+"#,
+        );
+        assert_eq!(cmds.len(), 0);
+    }
+
+    #[test]
+    fn ignore_markdown_backtikcs() {
+        let cmds = parse_commands("Don't do `@bors try`!");
```
jieyouxu, line 1243: Question (marginally adjacent to this PR): I'm not familiar with new bors, do you need the `@` to precede a new bors command? I.e. would `@/bors2 try` trigger new bors? It wouldn't trigger old bors, tho old bors is an actual github user (albeit a bot).

FWIW, I would expect that to be a "yes", that a preceding `@` is required (mostly to delineate it from other text).

jieyouxu (commented):
Apparently I lost this notification. Anyway, the ignored elements makes sense to me.

Though note that I haven't looked at this vs triagebot's ignore behavior, and it probably would make sense if both matched.

src/bors/command/parser.rs
```diff
@@ -61,10 +61,57 @@ impl CommandParser {
...
+            Event::Start(tag) => match tag {
+                // Ignore content in the following wrapping elements
+                Tag::BlockQuote(_)
+                | Tag::CodeBlock(_)
+                | Tag::HtmlBlock
+                | Tag::Link { .. }
+                | Tag::Heading { .. }
+                | Tag::Image { .. } => {
+                    stack.push(tag);
+                }
```
jieyouxu: Question: not super sure, does the following trigger new bors?

```
<!--
I am markdown HTML comment
@bors r+ rollup
-->
```

Old bors doesn't consider HTML comment as a valid "escape", so the above would trigger old bors. (It's useful if you want to be funny but not necessarily mission-critical so.)

Kobzol: Heh, I was just thinking about the same thing :laughing: Added support for this, along with a test :)

src/bors/command/parser.rs
Kobzol: Bors just looks for whatever prefix you configure it with when you start it. We currently configure it with `@bors2` on rust-lang/rust.

PR #301, opened 2025-06-05
Track PR title

Sakib25800:
This PR adds a new column for title to the pull request table in the database.


src/bors/handlers/review.rs
```diff
@@ -158,6 +162,7 @@ pub(super) async fn command_undelegate(
         .upsert_pull_request(
```
Kobzol, line 162: Hmm, I see now that these upsert calls are very repetitive, and we pass a bunch of arguments to them. I'll take a look if we could do them one level above.

Kobzol (approved):
Looks straightforward enough, thanks!

PR #303, opened 2025-06-05
Track PR author

Sakib25800:
This PR tracks PR author. This will be needed for the queue page.

### DIFF
--- a/src/database/client.rs
+++ b/src/database/client.rs
@@ -99,12 +99,7 @@ impl PgDbClient {
         repo: &GithubRepoName,
         params: UpsertPullRequestParams,
     ) -> anyhow::Result<PullRequestModel> {
-        let pr = upsert_pull_request(
-            &self.pool,
-            repo,
-            &params,
-        )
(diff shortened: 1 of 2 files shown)


src/database/client.rs
```diff
@@ -94,11 +94,13 @@ impl PgDbClient {
         get_pull_request(&self.pool, repo, pr_number).await
     }
 
+    #[allow(clippy::too_many_arguments)]
     pub async fn upsert_pull_request(
```
Sakib25800, line 97: Considering I need to add another field (`assignee`), it may be best to pull the params out into a struct? Or do we just clippy allow here?

Kobzol: Yeah, I was thinking the same thing. Creating some helper struct would indeed be useful. We could implement `From<PullRequest>` for it (if possible) to make the conversion simple. Or just some function that converts `PullRequest` to the struct.

Kobzol (approved):
Great, thanks!

PR #304, opened 2025-06-08
Track PR assignees

Sakib25800:
This PR tracks assignees for pull requests. We need this for the queue page.

Assignees are updated on the following events:
- PR open
- PR assigned
- PR unassigned

`assignees` is stored as a text array in the database.

Note that Homu stores (a single) `assignee` in the database, but that's now a legacy field in the GH API.


migrations/20250607174500_add_assignees_to_pr.up.sql
```diff
@@ -0,0 +1,2 @@
+-- Add up migration script here
+ALTER TABLE pull_request ADD COLUMN assignees TEXT[] NOT NULL DEFAULT ARRAY[]::TEXT[];
```
Kobzol: I'm personally not a big fan of using arrays in SQL, and I don't think it's really needed here. For us, this field is completely opaque, we only use it for rendering, so I would just store the assignees as a comma-separated string. In 99.9% of cases it will be a single username anyway.

src/bors/handlers/pr_events.rs
```diff
@@ -445,6 +466,28 @@ mod tests {
...
+                .wait_for_pr(default_repo_name(), pr.number.0, |pr| {
+                    pr.assignees.is_empty()
+                })
+                .await?;
+            tester
+                .assign_pr(default_repo_name(), pr.number.0, User::reviewer())
+                .await?;
+            tester
+                .wait_for_pr(default_repo_name(), pr.number.0, |pr| {
+                    pr.assignees == vec!["reviewer"]
```
Kobzol: `== vec![User::reviewer().name]`

Kobzol (commented):
Thans!

migrations/20250607174500_add_assignees_to_pr.up.sql
Sakib25800: After looking into it, I'm also not a fan of SQL arrays.

src/bors/handlers/pr_events.rs
Kobzol: You actually fixed it in another test (where it was also hardcoded :laughing:).

Kobzol (approved):
Thank you! :)

src/bors/handlers/pr_events.rs
Sakib25800: 😂

PR #305, opened 2025-06-08
Replace asserts with `.wait_for_pr`

Sakib25800:
There's some old code that utilises assertions where it could be updated to use `.wait_for_pr(...)`.

### DIFF
--- a/src/bors/handlers/refresh.rs
+++ b/src/bors/handlers/refresh.rs
@@ -325,11 +325,10 @@ timeout = 3600
                 })
                 .await?;
             tester.refresh_prs().await;
-            tester
-                .wait_for_pr(default_repo_name(), pr.number.0, |pr| {
-                    pr.pr_status == PullRequestStatus::Open
-                })
-                .await?;
+            assert_eq!(
+                tester.get_pr_from_db(pr.number.0).await.pr_status,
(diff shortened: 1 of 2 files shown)


src/bors/handlers/refresh.rs
```diff
@@ -325,15 +325,11 @@ timeout = 3600
...
-            assert_eq!(
-                tester
-                    .db()
-                    .get_pull_request(&default_repo_name(), pr.number)
-                    .await?
-                    .unwrap()
-                    .pr_status,
-                PullRequestStatus::Open
-            );
+            tester
```
Kobzol: Hmm, I'm not actually sure if these are 100% equivalent. We want to make sure here that the PR has the specific state already after the refresh is done, not sometime in the future. Of course, the `await` might cause some minor race conditions on its own, but I still think that we shouldn't need `wait_for_pr` in these situations.

I agree that the code is a bit verbose though, so I'm fine with adding a function to `tester`, something like `get_pr_from_db(pr_number)` that loads a PR with a given number from the default repo and unwraps the result.

Sakib25800: Yeah, you're right. I can see the issue. 

I added the tester method: `get_pr_from_db(pr_number)`.

PR #306, opened 2025-06-08
Fix flaky test `refresh_cancel_workflow_after_timeout`

Sakib25800:
This PR fixes the flaky test `refresh_cancel_workflow_after_timeout`.

From my understanding, there's a race condition:
1. Try command completes (build is created in DB)
2. Workflow event sent and processing beings
3. Test immediately cancels timed out builds
4. Test tries to get workflows for the build but finds none (this is where it fails)
5. All the while workflow processing is still going

Essentially the test is checking for workflows before the workflow event has actually finished creating the workflow in the DB.

Logs:
```sh
2025-06-08T14:02:00.872420Z ... get_workflows_for_build (test checking)
2025-06-08T14:02:00.986623Z ... create_workflow (workflow event still processing)
```

### DIFF
--- a/src/bors/handlers/refresh.rs
+++ b/src/bors/handlers/refresh.rs
@@ -280,6 +280,28 @@ timeout = 3600
                     .await?;
                 WAIT_FOR_WORKFLOW_STARTED.sync().await;
 
+                // Wait for workflow to be created before timing out builds
+                tester
+                    .wait_for(|| async {
+                        let builds = tester
+                            .db()
+                            .get_running_builds(&default_repo_name())
+                            .await
+                            .unwrap();
+                        if builds.is_empty() {
+                            return Ok(false);
+                        }
+
+                        let workflows = tester
+                            .db()
+                            .get_workflows_for_build(&builds[0])
+                            .await
+                            .unwrap();
+                        Ok(!workflows.is_empty())
+                    })
+                    .await
+                    .unwrap();
+
                 with_mocked_time(Duration::from_secs(4000), async {
                     tester.cancel_timed_out_builds().await;
                 })


src/bors/handlers/refresh.rs
```diff
@@ -280,6 +280,28 @@ timeout = 3600
                     .await?;
                 WAIT_FOR_WORKFLOW_STARTED.sync().await;
 
+                // Wait for workflow to be created before timing out builds
```
Kobzol: Hmm, something weird must be going on, because the logic that you have here should actually be performed by the `WAIT_FOR_WORKFLOW_STARTED.sync().await` line above.

Sakib25800:
I'm guessing changing `thread_local::ThreadLocal` for `std::sync::OnceLock` would solve the issue. 

Since `TestSyncMarker` currently uses `ThreadLocal`, which creates separate channel pairs for each thread. The test's workflow check happened just before the workflow was even created in the database even with  `WAIT_FOR_WORKFLOW_STARTED.sync().await` being called so the sync mechanism fails.

Using`OnceLock` would make sure all threads share the same sync channel.

Kobzol:
> Since TestSyncMarker currently uses ThreadLocal, which creates separate channel pairs for each thread. The test's workflow check happened just before the workflow was even created in the database even with WAIT_FOR_WORKFLOW_STARTED.sync().await being called so the sync mechanism fails.

> UsingOnceLock would make sure all threads share the same sync channel.

Hmm, but it uses a thread-local exactly for this reason. Each test runs on a single thread (otherwise a bunch of things would break), and we don't want to share the values of the markers between threads, because then they would conflict.

However, you actually gave me an idea of what might be wrong. We do not clear the value of the markers at the beginning of each test. So a previous test running on the same thread might have set the marker, which is a problem. I guess we will need to improve the mark/sync mechanism to make sure that this doesn't happen.

Sakib25800:
I've made it drain stale messages before waiting for a fresh mark. 

But if you're going to improve the mark/sync mechanisms then I can close this PR.

PR #310, opened 2025-06-10
Drain previously seen notifications in `TestSyncMarker`

Kobzol:
This should hopefully fix the spurious errors that we have been seeing with `refresh_cancel_workflow_after_timeout`.

CC @Sakib25800


Kobzol:
I'll do a few force pushes to check if CI is (repeatedly) green.

PR #312, opened 2025-06-13
Help page redesign

Sakib25800:
Redesigns the help page and adds some base styling. 

- Fira Sans is used since other Rust sites use it 
- Fira Mono for code blocks 
- #28607F is official Rust blue

### DIFF
--- a/templates/help.html
+++ b/templates/help.html
@@ -1,109 +1,149 @@
 {% extends "base.html" %}
 
-{% block title %}Bors{% endblock %}
+{% block title %}Help - Bors{% endblock %}
 
 {% block head %}
 <style>
-  div.wrapper {
-    max-width: 60em;
-    margin-left: auto;
-    margin-right: auto;
-  }
+    main {
+        max-width: 50rem;
+    }
+
+    table {
+        margin-block: var(--space-m);
+    }
+
+	tbody tr:last-child td {
+	    border-bottom: none;
+	}
+
+    th,
+    td {
+        padding: var(--space-xs) var(--space-s);
+        border-bottom: 1px solid var(--color-border-muted);
+        word-break: break-all;
+        overflow-wrap: break-word;
+    }
+
+    th {
+        background-color: var(--color-bg-higlight);
+    }
+
+    ul {
+        margin-block: var(--space-s);
+        padding-inline-start: var(--space-m);
+    }
+
+    li {
+        margin-block-end: var(--space-2xs);
+    }
 </style>
 {% endblock %}
 
 {% block body %}
-<div class="wrapper">
-  <h1>Bors</h1>
+<main>
+    <h1>Bors</h1>
 
-  <h2>Repositories</h2>
+    <h2>Repositories</h2>
+    <ul class="repos">
+      {% for repo in repos %}
+      <li>
+        <a href="queue/{{ repo.name }}">{{ repo.name }}</a>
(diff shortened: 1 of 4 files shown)


templates/help.html
```diff
@@ -1,135 +1,149 @@
...
-    </li>
-    <li>
-      <code>delegate+</code>: Delegate review permissions to the PR author.
-    </li>
-    <li>
-      <code>delegate-</code>: Remove any previously granted delegation.
-    </li>
-  </ul>
+    <h2>Commands</h2>
+    <p>Commands must be posted as comments on the PR and mention the bot account (e.g. <code>@bors</code>).</p>
```
Kobzol: Could you please put the actual bors prefix here? To show the actual bot name.

templates/help.html
```diff
@@ -1,135 +1,149 @@
...
+         <thead>
+             <tr>
+                 <th>Command</th>
+                 <th>Description</th>
+             </tr>
+         </thead>
+         <tbody>
+             <tr>
+                 <td><code>info</code></td>
+                 <td>Get information about the current PR</td>
```
Kobzol: We already have the `@bors help` command, which duplicates this information. I thought about making some data-based structure describing the commands and then rendering them both to HTML and the GH command, but in the end I just decided to hard-code it, as it allows us to configure the help to be better.

We should not forget to update both places when something changes though. Could you please add some note to the bors development MD file to update both this page and the `@bors help` command if some commands change? Ideally also add a short note as a comment on top of the `BorsCommand` struct. Doesn't have to be in this PR, could be a separate one. Thanks!

docs/commands.md
```diff
@@ -1,27 +0,0 @@
-# Commands
-Here is a list of commands currently supported by bors. All commands have to be prefixed by the *command prefix*,
```
Kobzol, line 2: Oh, I forgot we also have this :laughing: I guess it's better to check the help page, since this will bitrot, yeah. But we lost the information about which permissions are allowed. Would be nice to at least show that on the website.

docs/development.md
```diff
@@ -121,6 +121,12 @@ $ cargo sqlx prepare -- --all-targets
 
 After that, you should commit the changes to the `.sqlx` directory.
 
+## Updating commands
+When modifying commands, make sure to update both:
+
+1. The `BorsCommand` enum in `src/bors/command/mod.rs`
```
Kobzol: The main thing is to modify the output of the help command. Modifying `BorsCommand` is kind of implicit "when modifying commands" :laughing:

docs/commands.md
Sakib25800: - Added permissions column

Kobzol:
Thanks! I went through the commands on the help page and noticed that it doesn't fully correspond to the `help` command (e.g. `delegate=try` is missing). Could you please double check it and make sure that everything that is in `@bors help` is on the help page? Thanks!

Sakib25800:
I've made sure that the `help` command and help page now match and also added `rollup` and `rollup-` to the help command as it was missing.

Kobzol (approved):
Thanks! Reading the help table in HTML is quite difficult, eventually we should define the help commands in Rust (even with just a tuple of `(cmd, permissions, description)` and then just render it in the template with a for loop, to avoid repeating all the HTML table items. But that doesn't need to happen here.

PR #324, opened 2025-06-24
Add PR auto build column

Sakib25800:
This PR adds a new auto build column to PRs.

That's it.


migrations/20250624153928_add_auto_build_to_pr.up.sql
```diff
@@ -0,0 +1,2 @@
+-- Add up migration script here
+ALTER TABLE pull_request ADD COLUMN auto_build_id INTEGER REFERENCES build(id);
```
Kobzol: We might as well rename `build_id` to `try_build_id`, to make the distinction explicit (in another migration).

Kobzol (approved):
Thanks!

Issue #325, opened 2025-06-24
Merge queue tracking issue


Sakib25800:
- [x] Add auto builds to PRs (#324)
- [x] Delete attached auto build and cancel running workflows on PR push (#326)
- [x] Add merge queue skeleton - we introduce the merge queue and refresh event (#350)
- [x] Add auto build creation, merge to base branch (#330)
- [x] Add tests for PR#326 (#383)
- [x] Update auto build docs (#385)
- [x] Improve auto build error handling (#386)
  - [x] Reduce indentation and remove rightward drift (#405)
  - [x] Deal with merge conflicts / non-fast-forward (#406)
  - [x] Auto build startup failure (#410)
  - [x] Sanity checks (#413)
- Mini-DSL for testing API?  
    - https://github.com/rust-lang/bors/pull/330#discussion_r2223128697

Kobzol:
I think that you did most/all of what I described in https://github.com/rust-lang/bors/pull/386#issuecomment-3132390663. I think that we might want to have a few more tests, and maybe some DSL for that, but that's not urgent. I will take a look at the current implementation holistically, and try to test it on the kindergarten repo, and maybe add a few more tests.

In the meantime, if you want to continue forward, then I think you can either work on improving the queue page (https://github.com/rust-lang/bors/issues/371), implementing bors retry (https://github.com/rust-lang/bors/issues/319), or start thinking about rollups! Let me know on Zulip if you want to brainstorm how rollups could be handled.

PR #326, opened 2025-06-25
Delete auto build on PR and cancel running workflows

Sakib25800:
- Deletes a PR's auto build on PR push (if it exists)
- Cancels a running auto builds workflows on PR push
- Adds new migration to delete associated workflows when a build is deleted

I can't really write tests for this yet since the merge queue is what creates auto builds, so I'll have to write tests for this in another PR after the merge queue is added.

### DIFF
--- a/src/bors/handlers/pr_events.rs
+++ b/src/bors/handlers/pr_events.rs
@@ -1,3 +1,4 @@
+use crate::bors::handlers::trybuild::cancel_build_workflows;
 use crate::PgDbClient;
 use crate::bors::event::{
     PullRequestAssigned, PullRequestClosed, PullRequestConvertedToDraft, PullRequestEdited,
@@ -7,7 +8,7 @@ use crate::bors::event::{
 use crate::bors::handlers::labels::handle_label_trigger;
 use crate::bors::mergeable_queue::MergeableQueueSender;
 use crate::bors::{Comment, PullRequestStatus, RepositoryState};
-use crate::database::MergeableState;
+use crate::database::{BuildStatus, MergeableState};
 use crate::github::{CommitSha, LabelTrigger, PullRequestNumber};
 use std::sync::Arc;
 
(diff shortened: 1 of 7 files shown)


src/bors/handlers/pr_events.rs
```diff
@@ -53,6 +54,13 @@ pub(super) async fn handle_push_to_pull_request(
 
     mergeable_queue.enqueue(repo_state.repository().clone(), pr_number);
 
+    if let Some(auto_build) = &pr_model.auto_build {
+        if auto_build.status == BuildStatus::Pending {
+            tracing::info!("Cancelling auto build for PR {pr_number} due to push");
+            cancel_build_workflows(&repo_state.client, &db, auto_build).await?;
```
Kobzol: We should post a comment on the PR, otherwise it would be quite weird why was a running auto build cancelled out of nowhere.

src/database/operations.rs
```diff
@@ -941,3 +941,31 @@ pub(crate) async fn upsert_repository(
...
+
+/// Removes the auto build associated with a PR and deletes the build record (if one exists).
+pub(crate) async fn delete_auto_build(
+    executor: impl PgExecutor<'_>,
+    pr_id: i32,
+) -> anyhow::Result<()> {
+    measure_db_query("delete_auto_build", || async {
+        sqlx::query!(
+            r#"
+            -- Clear the PR's auto_build_id
```
Kobzol, line 953: Cool query! :) Would be indeed nice to add a few tests later.

Kobzol (commented):
Does homu also proactively delete builds and workflows in DB when a push to a PR happens?

Kobzol:
Thanks! It's funny, I don't remember a single case where users would push to a PR after it being approved, or more accurately while the auto build was running. It's good to be prepared though :)

src/database/operations.rs
Sakib25800: Added as the next task #325.

Kobzol (approved):
Thanks!

PR #328, opened 2025-06-27
Refactor seed script to use timestamp for file names 

Sakib25800:
This PR fixes an error that occurs when test files already exist on the default branch. When the script creates a new branch  and attempts to create a file with the same name, the GH API requires the SHA of the existing file to update it.

The `create_file(...)` method is trying to create a new file, but since it already exists on the branch, GitHub expects an update operation with the SHA.

```console
✗ Failed to create PR 1: Invalid request.
"sha" wasn't supplied.: 422 {"message": "Invalid request.\n\n\"sha\" wasn't supplied.", "documentation_url": "https://docs.github.com/rest/repos/contents#create-or-update-file-contents", "status": "422"}
```

We solve this by making the file name unique using the timestamp, ensuring no conflicts with existing files.

### DIFF
--- a/scripts/seed.py
+++ b/scripts/seed.py
@@ -18,7 +18,8 @@ from github import Github, Auth
 from github.GithubException import GithubException
 
 def create_pr(repo, index):
-    branch_name = f"test-pr-{index}-{int(time.time())}"
+    timestamp = int(time.time())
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Heh, I remember thinking to myself "I wonder what would happen if the files already exist" during the previous review, but then forgot about it 😆

Issue #322, opened 2025-06-23
Implement PR build status using check run API


Kobzol:
After a try or an auto build is started on a PR, we should use the GitHub check run API to mark the PR as being "CI pending" (yellow dot). Similarly as homu did, although it used commit statuses instead of check runs.

- [x] Try builds (https://github.com/rust-lang/bors/pull/338)
- [x] Auto builds (https://github.com/rust-lang/bors/pull/330)

Kobzol:
Awesome!

PR #338, opened 2025-07-03
Add try build check run

Sakib25800:
Fixes half of #322 

This PR adds the dots and a title: "Bors try build". No detail or text is added to the check run UI. 

The check run is added to the PR head and starts off in pending -> error / success. 

Success: https://github.com/Sakib25800/rust/pull/424
Fail: https://github.com/Sakib25800/rust/pull/425

Before merging the try build we could also set `CheckRunStatus` to `CheckRunStatus::Queued` but I think the window is so small it's not worth adding and no one is going to notice.

### DIFF
--- a/src/bors/handlers/refresh.rs
+++ b/src/bors/handlers/refresh.rs
@@ -3,6 +3,7 @@ use std::time::Duration;
 
 use anyhow::Context;
 use chrono::{DateTime, Utc};
+use octocrab::params::checks::{CheckRunConclusion, CheckRunStatus};
 use std::collections::BTreeMap;
 
 use crate::bors::Comment;
@@ -28,6 +29,22 @@ pub async fn cancel_timed_out_builds(
 
             db.update_build_status(&build, BuildStatus::Cancelled)
                 .await?;
+
+            if let Some(check_run_id) = build.check_run_id {
+                if let Err(error) = repo
+                    .client
+                    .update_check_run(
+                        check_run_id as u64,
+                        CheckRunStatus::Completed,
+                        Some(CheckRunConclusion::TimedOut),
+                        None,
+                    )
+                    .await
+                {
+                    tracing::error!("Could not update check run {check_run_id}: {error:?}");
+                }
+            }
+
             if let Some(pr) = db.find_pr_by_build(&build).await? {
                 if let Err(error) = cancel_build_workflows(&repo.client, db, &build).await {
                     tracing::error!(
@@ -181,11 +198,13 @@ fn elapsed_time(date: DateTime<Utc>) -> Duration {
 mod tests {
     use crate::bors::PullRequestStatus;
     use crate::bors::handlers::refresh::MOCK_TIME;
+    use crate::bors::handlers::trybuild::TRY_BUILD_CHECK_RUN_NAME;
     use crate::database::{MergeableState, OctocrabMergeableState};
     use crate::tests::mocks::{
         BorsBuilder, GitHubState, default_pr_number, default_repo_name, run_test,
     };
     use chrono::Utc;
+    use octocrab::params::checks::{CheckRunConclusion, CheckRunStatus};
     use std::future::Future;
     use std::time::Duration;
     use tokio::runtime::RuntimeFlavor;
@@ -262,6 +281,33 @@ timeout = 3600
                         .len(),
                     0
                 );
+                Ok(tester)
+            })
+            .await;
+    }
+
+    #[sqlx::test]
+    async fn refresh_cancel_build_updates_check_run(pool: sqlx::PgPool) {
+        BorsBuilder::new(pool)
+            .github(gh_state_with_long_timeout())
+            .run_test(|mut tester| async move {
+                tester.post_comment("@bors try").await?;
+                tester.expect_comments(1).await;
+
+                with_mocked_time(Duration::from_secs(4000), async {
+                    tester.cancel_timed_out_builds().await;
+                })
+                .await;
+                tester.expect_comments(1).await;
+
+                tester.expect_check_run(
+                    &tester.default_pr().await.get_gh_pr().head_sha,
+                    TRY_BUILD_CHECK_RUN_NAME,
+                    "Bors try build",
+                    CheckRunStatus::Completed,
+                    Some(CheckRunConclusion::TimedOut),
+                );
+
                 Ok(tester)
             })
             .await;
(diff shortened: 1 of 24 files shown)


src/database/mod.rs
```diff
@@ -310,6 +310,7 @@ pub struct BuildModel {
     pub status: BuildStatus,
     pub parent: String,
     pub created_at: DateTime<Utc>,
+    pub check_run_id: Option<i64>,
```
Kobzol, line 319: Can you please add a doc comment on top of this to explain what it is? We should really document all the database fields, we now have quite a lot of them :grin:

src/bors/handlers/trybuild.rs
```diff
@@ -91,6 +96,35 @@ pub(super) async fn command_try_build(
 
             handle_label_trigger(repo, pr.number(), LabelTrigger::TryBuildStarted).await?;
 
+            let build = db
```
Kobzol: Maybe we can just return the buil ID from `run_try_build`, and pass it to `create_check_run` and `update_build_check_run_id`? We don't need to load the whole Build from the DB.

src/bors/handlers/trybuild.rs
```diff
@@ -91,6 +96,35 @@ pub(super) async fn command_try_build(
...
+                .find_build(
+                    repo.repository(),
+                    TRY_BRANCH_NAME.to_string(),
+                    merge_sha.clone(),
+                )
+                .await?
+                .ok_or_else(|| anyhow!("Could not find build just created"))?;
+
+            let check_run = repo
```
Kobzol, line 103: Can you write some comment on top of this please, to explain what it does?

src/bors/handlers/trybuild.rs
```diff
@@ -861,4 +898,27 @@ try_failed = ["+foo", "+bar", "-baz"]
...
+
+            tester.expect_check_run(
+                &tester.default_pr().await.get_gh_pr().head_sha,
+                TRY_BUILD_CHECK_RUN_NAME,
+                CheckRunStatus::InProgress,
+                "Bors try build",
+            );
+
+            let check_run = tester.get_check_run().await?;
+            insta::assert_snapshot!(check_run.summary, @"");
```
Kobzol: Maybe we should assert the title instead?

src/tests/mocks/repository.rs
```diff
@@ -600,6 +619,102 @@ async fn mock_config(repo: Arc<Mutex<Repo>>, mock_server: &MockServer) {
...
+            struct CheckRunRequest {
+                name: String,
+                head_sha: String,
+                status: String,
+                output: CheckRunRequestOutput,
+                external_id: String,
+            }
+
+            #[derive(serde::Deserialize)]
+            struct CheckRunRequestOutput {
```
Kobzol: Maybe we can reuse some of these from octocrab (e.g. https://docs.rs/octocrab/latest/octocrab/models/checks/struct.CheckRunOutput.html?). But might not be worth it.

src/tests/mocks/repository.rs
```diff
@@ -205,6 +206,18 @@ impl Repo {
         self.check_runs.push(check_run);
     }
 
+    pub fn update_check_run(
+        &mut self,
+        check_run_id: u64,
+        status: String,
+        conclusion: Option<String>,
+    ) {
+        if let Some(check_run) = self.check_runs.get_mut(check_run_id as usize) {
```
Kobzol: If the check run is missing, that should be a failure in the test, right? So I'd unwrap here.

Kobzol (commented):
Thank you! We should also update the check run status in `cancel_timed_out_builds`.

src/database/mod.rs
Sakib25800: I had to go through the codebase to figure out what some of them meant tbh. 

I'll make a PR for it. Just creating an issue so I remember to: #339.

src/bors/handlers/trybuild.rs
Sakib25800: Do you mean assert the `summary` / `text`? I chose to snapshot test those because they are markdown compatible.

src/tests/mocks/repository.rs
Sakib25800: I tried to but a lot of the enums are marked as `non_exhaustive` and missing Serde traits.

src/bors/handlers/trybuild.rs
Kobzol: But they are also empty, no? :laughing: It's a bit weird asserting empty text that we don't even set. While the title is something that we do set on the check run.

src/bors/handlers/refresh.rs
```diff
@@ -28,6 +29,22 @@ pub async fn cancel_timed_out_builds(
...
+                    .client
+                    .update_check_run(
+                        check_run_id as u64,
+                        CheckRunStatus::Completed,
+                        Some(CheckRunConclusion::Cancelled),
+                        None,
+                    )
+                    .await
+                {
+                    tracing::error!("Could not update check run {}: {error:?}", check_run_id);
```
Kobzol: Inline `check_run_id` (we'd have a Clippy lint warning about it in a few Rust versions :) ).

src/bors/handlers/refresh.rs
```diff
@@ -28,6 +29,22 @@ pub async fn cancel_timed_out_builds(
...
             db.update_build_status(&build, BuildStatus::Cancelled)
                 .await?;
+
+            if let Some(check_run_id) = build.check_run_id {
+                if let Err(error) = repo
+                    .client
+                    .update_check_run(
+                        check_run_id as u64,
+                        CheckRunStatus::Completed,
+                        Some(CheckRunConclusion::Cancelled),
```
Kobzol: There is a specific `TimedOut` option, I think that would be more accurate.

src/bors/handlers/trybuild.rs
Sakib25800: It is weird but in the future we may populate it - I'll remove it. 

We do assert the title in `.expect_check_run(...)`?

Kobzol: Right, so I was confused at the difference, one thing was checked in `expect_check_run`, while some others were checked in the test after calling `get_check_run`. But now that function is gone, so nevermind.

Kobzol:
Thanks! I think this looks good, and we can test it out on real bors with try builds. Please just squash the commits a bit.

Kobzol (approved):
Thank you! Let's try it in the wild.

PR #349, opened 2025-07-09
Update check run on build cancelled

Sakib25800:
This PR updates the check run when a build is cancelled.

This includes both auto builds and try builds.

### DIFF
--- a/src/bors/handlers/workflow.rs
+++ b/src/bors/handlers/workflow.rs
@@ -203,9 +203,13 @@ async fn try_complete_build(
             (CheckRunStatus::Completed, Some(CheckRunConclusion::Success))
         };
 
-        repo.client
+        if let Err(error) = repo
+            .client
             .update_check_run(check_run_id as u64, status, conclusion, None)
-            .await?;
+            .await
+        {
+            tracing::error!("Could not update check run {check_run_id}: {error:?}");
+        }
     }
 
     let message = if !has_failure {


src/bors/handlers/trybuild.rs
```diff
@@ -307,6 +316,20 @@ pub async fn cancel_build_workflows(
     db.update_build_status(build, BuildStatus::Cancelled)
         .await?;
 
+    if let Some(check_run_id) = build.check_run_id {
```
Kobzol, line 319: I wonder if we should also reset `check_run_id` :thinking: Have you tried setting the conclusion for a check run twice? Does GitHub error out on it? Anyway, since we ignore the error here, it shouldn't matter that much. But in `try_complete_build` we do not ignore it, if that is executed multiple times, it could fail if it's not possible to "finish" a check run multiple times.

I wonder if we should represent the build status as something like
```rust
pub enum BuildStatus {
    /// The build is still waiting for results.
    Pending { check_run_id: u64 },
    /// The build has succeeded.
    Success,
    /// The build has failed.
    Failure,
    /// The build has been manually cancelled by a user.
    Cancelled,
    /// The build ran for too long and was timeouted by the bot.
    Timeouted,
}
```
but that would probably be overkill and complicate things.

Just leave it as it is for now, but please also ignore the check run error in `try_complete_build`.

Sakib25800: > Have you tried setting the conclusion for a check run twice? Does GitHub error out on it? 

Tested it. It doesn't error out.

Kobzol: Ok, that's good. Thank you!

PR #350, opened 2025-07-09
Merge queue skeleton

Sakib25800:
This PR adds the "skeleton" of the merge queue. No logic or processing but just the event itself. 

It adds the background process and the refresh event.

### DIFF
--- a/src/bors/handlers/mod.rs
+++ b/src/bors/handlers/mod.rs
@@ -318,7 +318,7 @@ pub async fn handle_bors_global_event(
             crate::bors::WAIT_FOR_PR_STATUS_REFRESH.mark();
         }
         BorsGlobalEvent::ProcessMergeQueue => {
-            merge_queue_tx.send(()).await?;
+            merge_queue_tx.trigger().await?;
         }
     }
     Ok(())
--- a/src/bors/merge_queue.rs
+++ b/src/bors/merge_queue.rs
@@ -19,7 +19,7 @@ pub struct MergeQueueSender {
 }
 
 impl MergeQueueSender {
-    pub async fn send(&self, _value: ()) -> Result<(), mpsc::error::SendError<()>> {
+    pub async fn trigger(&self) -> Result<(), mpsc::error::SendError<()>> {
         self.inner
             .send(MergeQueueEvent::Trigger)
             .await
--- a/src/bors/handlers/review.rs
+++ b/src/bors/handlers/review.rs
@@ -56,7 +56,7 @@ pub(super) async fn command_approve(
     db.approve(&pr.db, approval_info, priority, rollup).await?;
     handle_label_trigger(&repo_state, pr.number(), LabelTrigger::Approved).await?;
 
-    merge_queue_tx.send(()).await?;
+    merge_queue_tx.trigger().await?;
     notify_of_approval(&repo_state, pr, approver.as_str()).await
 }
 
@@ -176,7 +176,7 @@ pub(super) async fn command_close_tree(
     )
     .await?;
 
-    merge_queue_tx.send(()).await?;
+    merge_queue_tx.trigger().await?;
     notify_of_tree_closed(&repo_state, pr.number(), priority).await
 }
 
@@ -195,7 +195,7 @@ pub(super) async fn command_open_tree(
     db.upsert_repository(repo_state.repository(), TreeState::Open)
         .await?;
 
-    merge_queue_tx.send(()).await?;
+    merge_queue_tx.trigger().await?;
     notify_of_tree_open(&repo_state, pr.number()).await
 }
 
--- a/src/bors/handlers/workflow.rs
+++ b/src/bors/handlers/workflow.rs
@@ -224,7 +224,7 @@ async fn try_complete_build(
 
     // Trigger merge queue when an auto build completes
     if payload.branch == AUTO_BRANCH_NAME {
-        merge_queue_tx.send(()).await?;
+        merge_queue_tx.trigger().await?;
     }
 
     Ok(())


src/bors/merge_queue.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    let fut = async move {
+        while rx.recv().await.is_some() {
+            let span = tracing::info_span!("MergeQueue");
+            tracing::debug!("Processing merge queue");
+            if let Err(error) = merge_queue_tick(ctx.clone()).instrument(span.clone()).await {
+                // In tests, we want to panic on all errors.
+                #[cfg(test)]
+                {
+                    panic!("Handler failed: {error:?}");
```
Kobzol:
```suggestion
                    panic!("Merge queue handler failed: {error:?}");
```

src/bors/merge_queue.rs
```diff
@@ -0,0 +1,78 @@
...
+    crate::bors::WAIT_FOR_MERGE_QUEUE.mark();
+
+    Ok(())
+}
+
+pub fn start_merge_queue(
+    ctx: Arc<BorsContext>,
+) -> (
+    mpsc::Sender<MergeQueueEvent>,
+    impl Future<Output = anyhow::Result<()>>,
```
Kobzol:
```suggestion
    impl Future<Output = ()>,
```
I guess we don't need any error propagation here in the end, same as the other "endless" async loops.

That can also this:
```rust
result = merge_queue_fut => {
                    match result {
                        Ok(()) => tracing::error!("Merge queue handling process has ended"),
                        Err(e) => tracing::error!("Merge queue handling process has ended with error: {:?}", e)
                    }
                }
```

src/tests/mocks/bors.rs
```diff
@@ -108,6 +109,7 @@ pub struct BorsTester {
     github: GitHubState,
     db: Arc<PgDbClient>,
     mergeable_queue_tx: MergeableQueueSender,
+    merge_queue_tx: Sender<MergeQueueEvent>,
```
Kobzol: Maybe also introduce an alias, as we have for `MergeableQueueSender`? We should also somehow rename these, it's quite confusing that we have a "merge queue" and "mergeable" queue :laughing: I'll try to think of a better name, suggestions welcome.

src/database/operations.rs
```diff
@@ -991,3 +991,58 @@ pub(crate) async fn delete_auto_build(
...
+                auto_build AS "auto_build: BuildModel"
+            FROM pull_request as pr
+            LEFT JOIN build AS try_build ON pr.try_build_id = try_build.id
+            LEFT JOIN build AS auto_build ON pr.auto_build_id = auto_build.id
+            WHERE pr.repository = $1
+              AND pr.status = 'open'
+              AND pr.approved_by IS NOT NULL
+              AND pr.mergeable_state = 'mergeable'
+              -- Exclude PRs with failed auto builds
+              AND (auto_build.status IS NULL OR auto_build.status NOT IN ('failure', 'cancelled', 'timeouted'))
```
Kobzol: I'd prefer if we enumerated only the states that we want, so pending/success. A bit safer if we add some new (error) states in the future and forget to update this.

src/utils/sort_queue.rs
```diff
@@ -0,0 +1,66 @@
...
+            // 6. Compare PR numbers (older first)
+            .then_with(|| a.number.cmp(&b.number))
+    });
+    prs
+}
+
+fn get_status_priority(pr: &PullRequestModel) -> i32 {
+    match &pr.auto_build {
+        Some(build) => match build.status {
+            BuildStatus::Success => 1,
```
Kobzol, line 49: This is only relevant for the website, not for merging, but I always find the order of homu confusing, because if something is approved, but fails CI, it suddenly jumps to the bottom of the page. I would ideally like to see approved failed PRs just below the approved "waiting for build" PRs.

Maybe we could sort by approved first, and only then by status? Success/pending should always be approved, so it shouldn't change anything at the "front" of the queue. What do you think?

src/utils/sort_queue.rs
```diff
@@ -0,0 +1,66 @@
...
+
+fn get_mergeable_priority(pr: &PullRequestModel) -> i32 {
+    match pr.mergeable_state {
+        MergeableState::Mergeable => 0,
+        MergeableState::HasConflicts => 1,
+        MergeableState::Unknown => 1,
+    }
+}
+
+fn get_rollup_priority(rollup: Option<&RollupMode>) -> i32 {
```
Kobzol: I think that in homu never comes before iffy? Or at least it would make sense to me, to distinguish them, rather than using the same sort value for them. Also can we just start at `0`, same as for the other values? Let's use `u32` everywhere and avoid negative numbers.

docs/design.md
```diff
@@ -113,6 +113,90 @@ the branch to parent, and then again after merging the PR commit).
...
+   |                                   |                                     |       |         |
+   |                                   | fast-forward base branch            |       |         |
+   |                                   |------------------------------------>|       |         |
+   |                                   |                                     |       |         |
+   | comment: ":sunny: Test successful"|                                     |       |         |
+   |<----------------------------------|                                     |       |         |
+   |                                   |                                     |       |         |
+```
+
+The merge queue first merges the PR with the latest base branch commit in `automation/bors/auto-merge`, then pushes
```
Kobzol, line 190: Could you please also update the `Special branches` section in `README.md`? Thanks.

src/config.rs
```diff
@@ -21,6 +21,8 @@ pub struct RepositoryConfig {
     pub labels: HashMap<LabelTrigger, Vec<LabelModification>>,
     #[serde(default, deserialize_with = "deserialize_duration_from_secs_opt")]
     pub min_ci_time: Option<Duration>,
+    #[serde(default)]
```
Kobzol, line 33: Please also document it here with a doc comment. I should have done that with the other fields above too :man_facepalming:

Sakib25800: Decided to add doc comments to all of them.

src/utils/sort_queue.rs
Sakib25800: Homu doesn't distinguish between `Never` and `Iffy` and treats them the equivalent. I'm guessing because both, as individual PRs, are equivalent in a sense?

I've gone for `Never` to come before `Iffy` though and for `u32`.

src/utils/sort_queue.rs
Sakib25800: Yeah, the sorting (for the website) is odd. 

I've update the sorting logic to prioritize approval status. 

The new sorting order is:
1. Pending builds (must come first to properly block the merge queue)
2. Approval status (approved PRs grouped together)
3. Build status (within each approval group)
4. Mergeability, priority, rollup, age (existing tie-breakers)

Sakib25800:
Tests have gotten stuck 😩

src/bors/handlers/mod.rs
```diff
@@ -313,6 +317,9 @@ pub async fn handle_bors_global_event(
             #[cfg(test)]
             crate::bors::WAIT_FOR_PR_STATUS_REFRESH.mark();
         }
+        BorsGlobalEvent::ProcessMergeQueue => {
+            merge_queue_tx.send(()).await?;
```
Kobzol:
```suggestion
            merge_queue_tx.trigger().await?;
```

src/utils/sort_queue.rs
```diff
@@ -0,0 +1,80 @@
...
+        MergeableState::HasConflicts => 1,
+        MergeableState::Unknown => 1,
+    }
+}
+
+fn get_rollup_priority(rollup: Option<&RollupMode>) -> u32 {
+    match rollup {
+        Some(RollupMode::Always) => 3,
+        Some(RollupMode::Maybe) => 2,
+        None => 1, // Default case
```
Kobzol:
```suggestion
        None => 2, // Default case
```
The default cause (no rollup mode) corresponds to `Maybe`, not `Iffy`.

Kobzol (commented):
Thanks, this looks pretty solid! Left a few nits, and would be nice to rebase it.

src/bors/merge_queue.rs
```diff
@@ -19,7 +19,7 @@ pub struct MergeQueueSender {
 }
 
 impl MergeQueueSender {
-    pub async fn send(&self, _value: ()) -> Result<(), mpsc::error::SendError<()>> {
+    pub async fn trigger(&self, _value: ()) -> Result<(), mpsc::error::SendError<()>> {
```
Kobzol: We also don't need the argument now :sweat_smile:

Sakib25800: I misread it as just a name change. 

Fixed.

Sakib25800:
Rebased.

Issue #288, opened 2025-05-30
Enable sanity checks for PR approval


Kobzol:
Homu does some checks that prevent it to approve PRs that contain some messages in the title (e.g. `[do not merge]` etc.). Bors should do the same. Example [here](https://github.com/rust-lang/rust/pull/112049#issuecomment-2921562667).

- [x] Draft status (https://github.com/rust-lang/bors/pull/336)
- [x] WIP title (https://github.com/rust-lang/bors/pull/366)
- [ ] Blocking labels (https://github.com/rust-lang/bors/pull/367)

jieyouxu:
(Apparently homu doesn't check draft, cf. #333)

PR #367, opened 2025-07-19
Forbid approving PRs with labels blocking approval

Kobzol:
Closes: https://github.com/rust-lang/bors/issues/288


rust-bors.example.toml
```diff
@@ -19,3 +19,7 @@ approve = ["+approved"]
 try = ["+foo", "-bar"]
 try_succeed = ["+foobar", "+foo", "+baz"]
 try_failed = []
+
+# Labels that will block approval when present on a PR
+# (Optional)
+labels_blocking_approval = ["final-comment-period", "proposed-final-comment-period"]
```
RalfJung, line 25: Should we add "needs-fcp" here, or is that a label we expect to keep even after FCP passed?

Kobzol: Note that this is just an example config, this hasn't been enabled on rust-lang/rust yet :) I'll send a PR shortly, we can discuss it there.

Kobzol: Btw seems like `needs-fcp` indeed isn't removed after the FCP finishes (https://github.com/rust-lang/rust/pull/79278), so maybe we shouldn't mark it as blocking.

PR #374, opened 2025-07-21
Configure web URL for bors and show it in approved commit message

Kobzol:
This will need configuration in simpleinfra (https://github.com/rust-lang/simpleinfra/pull/752).


Kobzol:
Merging to test on staging.

PR #383, opened 2025-07-23
Add auto build cancel tests

### LINKED ISSUE #369: When a PR gets unapproved (`r-`) during an auto build, the auto build should be cancelled
When an approved PR is currently being tested in an auto build, and it gets unapproved, we should cancel the auto build. In theory, we could also do this if the PR is converted to a draft, gets an approval blocking label, etc., but I think that unapproval is enough.

Sakib25800:
- Adds PR #326 tests. 
- Closes #369 

Also slips in a change from cooked to raw string literal for the auto build success comment.

### DIFF
--- a/src/bors/handlers/pr_events.rs
+++ b/src/bors/handlers/pr_events.rs
@@ -68,14 +68,6 @@ pub(super) async fn handle_push_to_pull_request(
             )
             .await
             {
-                Err(error) => {
-                    tracing::error!(
-                        "Could not cancel workflows for SHA {}: {error:?}",
-                        auto_build.commit_sha
-                    );
-
-                    notify_of_unclean_auto_build_cancelled_comment(&repo_state, pr_number).await?
-                }
                 Ok(workflow_ids) => {
                     tracing::info!("Auto build cancelled");
 
@@ -83,6 +75,14 @@ pub(super) async fn handle_push_to_pull_request(
                         .client
                         .get_workflow_urls(workflow_ids.into_iter());
                     notify_of_cancelled_workflows(&repo_state, pr_number, workflow_urls).await?
+                }
+                Err(error) => {
+                    tracing::error!(
+                        "Could not cancel workflows for SHA {}: {error:?}",
+                        auto_build.commit_sha
+                    );
+
+                    notify_of_unclean_auto_build_cancelled_comment(&repo_state, pr_number).await?
                 }
             };
         }
--- a/src/bors/handlers/pr_events.rs
+++ b/src/bors/handlers/pr_events.rs
@@ -339,11 +339,19 @@ async fn notify_of_unclean_auto_build_cancelled_comment(
 #[cfg(test)]
 mod tests {
     use crate::bors::PullRequestStatus;
-    use crate::tests::mocks::default_pr_number;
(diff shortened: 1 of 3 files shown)


src/bors/handlers/pr_events.rs
```diff
@@ -753,4 +761,81 @@ mod tests {
...
+                tester.process_merge_queue().await;
+                tester.expect_comments(1).await;
+                tester
+                    .wait_for_default_pr(|pr| pr.auto_build.is_some())
+                    .await?;
+                tester.workflow_start(tester.auto_branch()).await?;
+                tester
+                    .push_to_pr(default_repo_name(), default_pr_number())
+                    .await?;
+                insta::assert_snapshot!(tester.get_comment().await?, @r"
```
Kobzol, line 805: This won't actually work for rollups, because the rollup PR won't be pushed to :thinking: But I haven't ever seen this occur in practice anyway.

Kobzol (commented):
While at it, could you also please implement the same auto build cancellation when `r-` is used on a PR that is currently being tested on `auto`? https://github.com/rust-lang/bors/issues/369

src/bors/handlers/review.rs
```diff
@@ -1423,4 +1426,35 @@ merge_queue_enabled = true
...
+                tester.post_comment("@bors r+").await?;
+                tester.expect_comments(1).await;
+                tester.process_merge_queue().await;
+                tester.expect_comments(1).await;
+                tester
+                    .wait_for_default_pr(|pr| pr.auto_build.is_some())
+                    .await?;
+                tester.workflow_start(tester.auto_branch()).await?;
+                tester.post_comment("@bors r-").await?;
+                tester.expect_comments(2).await;
```
Kobzol: This is a specific error I think, so let's check its content. I also wonder why it's two comments, shouldn't it be just one? :thinking: We should ideally combine the messages in to one.

Kobzol: Ah, I only saw the code now, that it sends two comments. It would be nice to merge that into one.

src/bors/handlers/review.rs
```diff
@@ -127,14 +132,50 @@ pub(super) async fn command_unapprove(
...
-            .post_comment(pr.number(), unapprove_non_open_pr_comment())
+            .post_comment(pr_num, unapprove_non_open_pr_comment())
             .await?;
         return Ok(());
     }
 
     db.unapprove(&pr.db).await?;
     handle_label_trigger(&repo_state, pr.number(), LabelTrigger::Unapproved).await?;
-    notify_of_unapproval(&repo_state, pr).await
+    notify_of_unapproval(&repo_state, pr).await?;
```
Kobzol: This will still send two comments, right? What I meant here was to essentially try to cancel the workflows first, store the result into a string, and optionally include that string in the unapproval comment.

Kobzol (commented):
Thanks for merging the comments, it would be great to also do it for the unapproval.

Sakib25800:
> It would be great to also do it for the unapproval.

I misread and merged comments for the push to PR but not the unapproval 🤦‍♂️.

Kobzol:
One test needs a change, otherwise LGTM.

PR #330, opened 2025-06-29
Merge queue

Sakib25800:
This PR adds the complete merge queue process from auto build -> merge to master. 

The merge queue works like so: 
- Every 30s `handle_merge_queue(...)` is called
- It selects a PR that can be merged (based on a set of merge queue priority rules)
- As part of the sorts, successful auto builds come first and pending ones second and a bunch of other rules. But those are 
the two most important. 
- If `handle_merge_queue(...)` sees:
  - Successful auto build, it will merge it 
  - Pending auto build, it will halt the queue until the auto build is done (to prevent simultaneous auto builds)
- From the check suite completed webhook, we attempt to `try_complete_build(...)` where we attempt to determine the build status and post relevant comments

Homu works in exactly the same way, with the only difference being that we use the GitHub checks API rather than the the commit status API.

PR #386, opened 2025-07-23
Improve merge queue error handling 

Sakib25800:
https://github.com/rust-lang/bors/pull/330#discussion_r2218510460

### DIFF
--- a/src/bors/merge_queue.rs
+++ b/src/bors/merge_queue.rs
@@ -183,7 +183,7 @@ pub async fn merge_queue_tick(ctx: Arc<BorsContext>) -> anyhow::Result<()> {
                     tracing::error!("Error starting auto build for PR {pr_num}: {:?}", error);
                     // Unexpected error - the PR will remain in the queue for a retry. This is most
                     // likely a transient GitHub error.
-                    continue;
+                    break;
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


src/bors/merge_queue.rs
```diff
@@ -165,18 +165,24 @@ pub async fn merge_queue_tick(ctx: Arc<BorsContext>) -> anyhow::Result<()> {
...
+                        .update_pr_mergeable_state(&pr, MergeableState::HasConflicts)
+                        .await?;
+                    break;
                 }
                 Err(error) => {
-                    // Unexpected error - the PR will remain in the "queue" for a retry.
                     tracing::error!("Error starting auto build for PR {pr_num}: {:?}", error);
+                    // Unexpected error - the PR will remain in the queue for a retry. This is most
+                    // likely a transient GitHub error.
                     continue;
```
Kobzol: In this case we should still break. We shouldn't just skip the PR because of a transient error. Maybe we'll find out in the future that some of there errors are not so transient though :) But we should have explicit detection of that.

Kobzol:
I'll take a look at all the errors that can happen and we can brainstorm some strategy for this.

Sakib25800:
I guess everything in this PR has been addressed (to some extent). 

> If the PR is not mergeable, we should trigger a mergeability check for it.

Forgot to add this. Will add in another PR. 

Closing.

Issue #389, opened 2025-07-25
Investigate builds without an attached PR


Kobzol:
I saw in production DB logs that there are some builds without an attached PR:
<img width="2087" height="224" alt="Image" src="https://github.com/user-attachments/assets/56b6fb61-cf64-4b06-9596-e6d4662e59cd" />

That's quite weird, we should figure out what's up with that.

PR #392, opened 2025-07-25
Handle build cancellation errors properly

Kobzol:
There are a few places in bors where we try to cancel a pending build. Before, we used to mostly ignore cancellation errors, because we assumed that cancelling the workflows is best effort. While that is true, marking the build itself as cancelled is actually load-bearing! Because if that happens, we shouldn't continue with normal bors logic, otherwise invariants might get broken, and can get e.g. "orphaned" builds (https://github.com/rust-lang/bors/issues/389).

So we need to ideally mark the build status as cancelled first, and also distinguish an error that happens during this from the less important error that happens when external CI workflows cannot be cancelled.

While I was changing this, I also unified code for cancelling auto builds, and realized that we should not delete cancelled auto builds from PRs. Bors has to be able to robustly handle auto builds *attached to PRs*, even if they are failed, cancelled or timed out. So there's no need to actually delete these builds from the DB, as we don't do that anywhere else, e.g. when we cancel try builds, so it was not consistent.

Closes: https://github.com/rust-lang/bors/issues/389

r? @Sakib25800

### DIFF
--- a/src/bors/handlers/refresh.rs
+++ b/src/bors/handlers/refresh.rs
@@ -9,7 +9,7 @@ use std::collections::BTreeMap;
 use crate::bors::PullRequestStatus;
 use crate::bors::RepositoryState;
 use crate::bors::comment::build_timed_out_comment;
-use crate::bors::handlers::trybuild::cancel_build_workflows;
+use crate::bors::handlers::workflow::cancel_build_workflows;
 use crate::bors::mergeable_queue::MergeableQueueSender;
 use crate::database::{BuildModel, BuildStatus};
 use crate::{PgDbClient, TeamApiClient};
@@ -42,7 +42,7 @@ async fn refresh_build(
         if let Some(pr) = db.find_pr_by_build(&build).await? {
             tracing::info!("Cancelling build {build:?}");
             if let Err(error) =
-                cancel_build(&repo.client, db, &build, CheckRunConclusion::TimedOut).await
+                cancel_build_workflows(&repo.client, db, &build, CheckRunConclusion::TimedOut).await
             {
                 tracing::error!(
                     "Could not cancel workflows for SHA {}: {error:?}",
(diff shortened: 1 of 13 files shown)


src/bors/handlers/refresh.rs
```diff
@@ -20,33 +21,61 @@ pub async fn refresh_pending_builds(
...
+            // Note that we could write an explicit query for finding these orphaned builds,
+            // but that could be quite expensive. Instead we piggyback on the existing logic
+            // for timed out builds; if a build is still pending and has no PR attached, then
+            // there likely won't be any additional event that could mark it as finished.
+            // So eventually all such builds will arrive here
+            tracing::warn!(
+                "Detected orphaned pending without a PR, marking it as time outed: {build:?}"
+            );
+            db.update_build_status(build, BuildStatus::Cancelled)
+                .await?;
```
Sakib25800: This should be `BuildStatus::Timeouted`.

src/bors/handlers/review.rs
```diff
@@ -1387,7 +1343,8 @@ merge_queue_enabled = true
                 Commit pr-1-sha has been unapproved.
 
                 Auto build cancelled due to unapproval. Cancelled workflows:
-                - https://github.com/rust-lang/borstest/actions/runs/1
+
+                - https://github.com/workflows/Workflow1/1
```
Sakib25800: I guess since this is a test-specific issue, it's not an actual issue, but the workflow URLs seem to have regressed.

I think in `src/tests/mocks/workflow.rs:115-118`, the URL format should be updated.

src/bors/handlers/refresh.rs
Kobzol: We can't actually know the real outcome unless we query GitHub, but you're right that time outed is a better fallback than cancelled, and it's what the log said. Good catch!

src/bors/handlers/review.rs
Kobzol: It's not really a regression, it's just that now we take the URL of the workflow from the actual workflow webhooks (which is IMO more correct). Before we were trying to reconstruct them by hand, but that was hacky (and didn't correspond with the mocked test data).

That being said, using the actual URl format that is used by GitHub doesn't sound like a bad idea, as you have suggested. So did that :+1:

Issue #334, opened 2025-07-01
Hide outdated comment about mergeability


Kobzol:
Once a PR becomes mergeable again, we should hide old comments about mergeability (https://github.com/rust-lang/rust/pull/143255#issuecomment-3021618048).

JerryQ17:
rla is able to hide outdated comments ([link](https://github.com/rust-lang/rust-log-analyzer/blob/26a046039b5ba078cd9c647f4bd29c81dec2678b/src/github.rs#L246-L326)), we can reuse that logic.

If no one is working on this, I would like to have a try. I'm not familiar with GraphQL, so it'll take some time for me to complete it.😁

Kobzol:
Triagebot also has similar logic. It would be great if we can do it without GraphQL, but maybe it's the best way.

We can either remember the node IDs of comments that we sent, along with some metadata, or we can scan comments on the PR from GH API. But the latter could be quite exppensive, not sure if there's a way to find comments by a specific author through the REST API.

JerryQ17:
> Triagebot also has similar logic. It would be great if we can do it without GraphQL, but maybe it's the best way.

I think it's the only way. I can't find a REST API to minimize a comment.

> We can either remember the node IDs of comments that we sent, along with some metadata, or we can scan comments on the PR from GH API. But the latter could be quite exppensive, not sure if there's a way to find comments by a specific author through the REST API.

The node id is required to minimize the comment anyway, remembering it will make the impl simpler.

If we scan comments to get those posted by bors (either by a REST API or manually filter with `GithubRepositoryClient::is_comment_internal`), it's still difficult to determine which comment to hide from its contents. Maybe the metadata in `<!-- homu: {} -->` will be useful.

Kobzol:
We could use the metadata, but if we need the node ID anyway, then remembering it in the DB is a better idea anyway.

PR #396, opened 2025-07-27
Comment label mechanism

JerryQ17:
This PR adds the ability to identify a comment, via the comment label.

This is the prerequisite of #334: To hide outdated comments, you need to identify which comment to hide.

cc @Kobzol since you created that issue

## Structure

The labels has a tree structure, mirroring of how bors handles events. This is currently a proof-of-concept implementation, only minimal branches are included.

```text
CommentLabel
├── Repository
│   └── Comment
│       └── Command
│           └── Try
│               └── TryBuildStarted
└── Global
    ├── InstallationsChanged
    ├── RefreshConfig
    ├── RefreshPermissions
    ├── CancelTimedOutBuilds
    ├── RefreshPullRequestMergeability
    ├── RefreshPullRequestState
    └── ProcessMergeQueue
```

This tree structure enables you to find not only a specific comment with `TryBuildStarted` label, but also a group of comments, such as all `Global` comments.

## API

For correctness issues, all APIs are provided by the root label `CommentLabel`. For non-root labels, you must call `.into()` to convert it to the root label. See `private::Label` trait's doc comment.

There are 5 methods:

- `impl From<CommentLabel> for PgLTree`
- `impl TryFrom<PgLTree> for CommentLabel`
- `impl Display for CommentLabel`
- `impl FromStr for CommentLabel`
- `CommentLabel::is_partial`: Partial labels can only be used to search child labels. You cannot create a comment with a partial label. A later PR that implements hiding comments will need this to check the label.

## Implementation

Labels are stored as `LTREE` in database, this requires enabling the `ltree` extension. It looks like `root.mid1.mid2.leaf` in string. (If you disagree with this, I can change it to`TEXT`)

The tree structure is difficult to maintain manually, I wrote some decl macros to simplify the label definition. They have simple syntax and is not very mysterious😈. Take the following two as exapmle:

```rust
pub enum CommentLabel {
    Repository(Option<RepositoryLabel>) = "repository",
    Global(Option<GlobalLabel>) = "global",
}

pub enum GlobalLabel {
    InstallationsChanged = "installations_changed",
    RefreshPullRequestMergeability = "refresh_pull_request_mergeability",
    ...
}
```

It's straight forward that:

```rust
let label = CommentLabel::Global(Some(GlobalLabel::RefreshPullRequestMergeability)); // A full label
let string = "global.refresh_pull_request_mergeability";
assert_eq!(label.to_string(), string);
assert_eq!(string.parse().unwrap(), label);

[...]

PR #397, opened 2025-07-29
Add comment minimization logic and minimize try build started comments

JerryQ17:
This PR implemented the functionality to minimize comments, following the steps in [this comment](https://github.com/rust-lang/bors/pull/396#issuecomment-3126093064).

However, the test infrastructure do not support the newly added GitHub GraphQL API, and is causing 12 tests to fail. I'm not familiar with the test infra so I need some help to fix them.

In my opinion, to make tests pass, I need to:

1. Mock the `graphql` POST endpoint in `GitHubMockServer::start` in `src/tests/mocks/github.rs`
2. Store the minimized comment's node_id somewhere. (Where to store it?)
3. Return `{}`, since the only GraphQL request we make is `minimizeComment`, which does not care about the return value
4. Provide a check method to make sure the comment is minimized successfully

Fixes: https://github.com/rust-lang/bors/issues/334


src/bors/comment.rs
```diff
@@ -25,6 +27,31 @@ pub enum CommentMetadata {
     TryBuildCompleted { merge_sha: String },
 }
 
+/// A label for a comment, used to identify the comment.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum CommentLabel {
```
Kobzol: We already have labels in bors that are triggered based on various actions, which correspond to GH issue labels, so I think that the name `CommentLabel` could be confusing. I would suggest something like `CommentTag`.

src/bors/comment.rs
```diff
@@ -25,6 +27,31 @@ pub enum CommentMetadata {
...
+    TryBuildStarted,
+}
+
+impl Display for CommentLabel {
```
Kobzol: Why do we need these `Display` and `FromStr` impls?

migrations/20250729051122_create_comment.up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE IF NOT EXISTS comment (
+  id SERIAL PRIMARY KEY,
+  repository TEXT NOT NULL,
+  pr_number BIGINT NOT NULL,
+  label TEXT NOT NULL,
```
Kobzol: `label` -> `tag`

src/bors/handlers/trybuild.rs
```diff
@@ -137,14 +138,22 @@ pub(super) async fn command_try_build(
                         cancelled_workflow_urls,
                     ),
                 )
-                .await
+                .await?;
+            db.insert_comment(
```
Kobzol: I was thinking of attaching the tags directly to the `Comment` struct, and then having a function that would take the GH client and a DB handle and insert both. But since we only need the tags at a single place right now, it would be a bit overkill, so let's leave it like it is for now, and I'll just leave this comment here for future reference.

src/database/client.rs
```diff
@@ -327,4 +328,27 @@ impl PgDbClient {
     ) -> anyhow::Result<Vec<PullRequestModel>> {
         get_merge_queue_prs(&self.pool, repo, tree_priority.map(|p| p as i32)).await
     }
+
+    pub async fn get_comments(
```
```suggestion
    pub async fn get_tagged_bot_comments(
```
Just to make it more obvious that it's just (some, tagged) comments made by the bot.

src/database/client.rs
```diff
@@ -327,4 +328,27 @@ impl PgDbClient {
...
+        &self,
+        repo: &GithubRepoName,
+        pr_number: PullRequestNumber,
+        label: CommentLabel,
+    ) -> anyhow::Result<Vec<CommentModel>> {
+        get_comments(&self.pool, repo, pr_number, label).await
+    }
+
+    pub async fn insert_comment(
```
```suggestion
    pub async fn record_tagged_bot_comment(
```

src/database/client.rs
```diff
@@ -327,4 +328,27 @@ impl PgDbClient {
...
+        &self,
+        repo: &GithubRepoName,
+        pr_number: PullRequestNumber,
+        label: CommentLabel,
+        node_id: &str,
+    ) -> anyhow::Result<()> {
+        insert_comment(&self.pool, repo, pr_number, label, node_id).await
+    }
+
+    pub async fn delete_comment(&self, comment: &CommentModel) -> anyhow::Result<()> {
```
```suggestion
    pub async fn delete_tagged_bot_comment(&self, comment: &CommentModel) -> anyhow::Result<()> {
```

src/database/mod.rs
```diff
@@ -510,3 +511,40 @@ pub struct RepoModel {
     pub tree_state: TreeState,
     pub created_at: DateTime<Utc>,
 }
+
+/// Represents a comment that will be minimized later.
```
Kobzol:
```suggestion
/// Represents a tagged comment made by the bors GitHub app that can be later minimized.
```

src/database/mod.rs
```diff
@@ -510,3 +511,40 @@ pub struct RepoModel {
...
+pub struct CommentModel {
+    pub id: PrimaryKey,
+    /// The GitHub repository this comment belongs to.
+    pub repository: GithubRepoName,
+    /// The pull request number this comment is associated with.
+    pub pr_number: PullRequestNumber,
+    /// The label of the comment.
+    pub label: CommentLabel,
```
Kobzol:
```suggestion
    pub tag: CommentTag,
```

src/database/mod.rs
```diff
@@ -510,3 +511,40 @@ pub struct RepoModel {
...
+        <String as sqlx::Type<sqlx::Postgres>>::type_info()
+    }
+}
+
+impl sqlx::Encode<'_, sqlx::Postgres> for CommentLabel {
+    fn encode_by_ref(
+        &self,
+        buf: &mut sqlx::postgres::PgArgumentBuffer,
+    ) -> Result<sqlx::encode::IsNull, BoxDynError> {
+        <String as sqlx::Encode<sqlx::Postgres>>::encode(self.to_string(), buf)
```
Kobzol: Let's implement the DB (de)serialization here, no need to tie it with `FromStr`/`ToString`. I know that some existing code in bors already does it, but it's not a great design.

Kobzol (commented):
Thank you, it looks great! Left some comments. Once you resolve them, I will take a look at adding the necessary test infra.

src/bors/comment.rs
JerryQ17: They are used in `src/database/mod.rs` (`to_string`, `parse`):

```rust
impl sqlx::Encode<'_, sqlx::Postgres> for CommentLabel {
    fn encode_by_ref(
        &self,
        buf: &mut sqlx::postgres::PgArgumentBuffer,
    ) -> Result<sqlx::encode::IsNull, BoxDynError> {
        <String as sqlx::Encode<sqlx::Postgres>>::encode(self.to_string(), buf)
    }
}

impl sqlx::Decode<'_, sqlx::Postgres> for CommentLabel {
    fn decode(value: sqlx::postgres::PgValueRef<'_>) -> Result<Self, BoxDynError> {
        <String as sqlx::Decode<sqlx::Postgres>>::decode(value)?
            .parse()
            .map_err(anyhow::Error::into_boxed_dyn_error)
    }
}
```

JerryQ17: I will move them to `sqlx::Encode` and `sqlx::Decode`.

JerryQ17:
Rebased

JerryQ17:
Oops, I forgot to modify the "Add `CommentLabel`" commit message😂

Kobzol:
Could you please rebase on top of master once https://github.com/rust-lang/bors/pull/402 lands? I did a large refactoring of the test suite in order to make testing GraphQL endpoints easier/possible.

Kobzol:
Thanks! I had to do some other fixes, but finally got the tests working. I also tested it on a live repo, seems to work well!

PR #406, opened 2025-08-07
Fast forward error handling

Sakib25800:
- If the base branch fast-forwarding fails:
  - We no longer mark the check run as failed since this does not accurately reflect the error
  - Errors 
    - `BranchUpdateError::Conflict` (409) -> Post comment and mark build as failed 
    - `BranchUpdateError::ValidationFailed` (422) -> Post comment and mark build as failed 
    - `_` -> Post comment and retry. If retries fail then post comment and mark build as failed

### DIFF
--- a/src/bors/handlers/pr_events.rs
+++ b/src/bors/handlers/pr_events.rs
@@ -658,10 +658,7 @@ mod tests {
         BorsBuilder::new(pool)
             .github(gh_state_with_merge_queue())
             .run_test(async |tester: &mut BorsTester| {
-                tester.post_comment("@bors r+").await?;
-                tester.expect_comments((), 1).await;
-                tester.process_merge_queue().await;
-                tester.expect_comments((), 1).await;
+                tester.start_auto_build(()).await?;
                 tester.wait_for_pr((), |pr| pr.auto_build.is_some()).await?;
                 tester.workflow_start(tester.auto_branch().await).await?;
                 tester.push_to_pr(()).await?;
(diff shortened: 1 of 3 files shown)


src/github/api/operations.rs
```diff
@@ -177,6 +181,9 @@ async fn update_branch(
 
     match status {
         StatusCode::OK => Ok(()),
+        StatusCode::NOT_FOUND => Err(BranchUpdateError::BranchNotFound(branch_name)),
+        StatusCode::CONFLICT => Err(BranchUpdateError::Conflict(branch_name)),
+        StatusCode::UNPROCESSABLE_ENTITY => Err(BranchUpdateError::ValidationFailed(branch_name)),
         _ => Err(BranchUpdateError::BranchNotFound(branch_name)),
```
Kobzol: This should be NetworkError, right?

Sakib25800: Thanks for the catch, should be `BranchUpdateError::InternalServerError`.

src/bors/merge_queue.rs
```diff
@@ -702,4 +610,128 @@ mod tests {
             &["main-sha1", "merge-0-pr-1", "merge-1-pr-3", "merge-2-pr-2"],
         );
     }
+
+    #[sqlx::test]
+    async fn auto_build_push_conflict_comment(pool: sqlx::PgPool) {
+        run_merge_queue_test(pool, async |tester: &mut BorsTester| {
+               start_auto_build(tester).await?;
```
Kobzol: This is using internal API in a test, which I'd like to avoid, because it then tests things that don't happen in the real-world. If you often need to test a PR having a successful auto build, could you please create a helper test function that will approve the PR, run the merge queue and let it start the auto build instead?

Kobzol:
Omg, I'm stupid. You already had a helper function, it was just named the same as `merge_queue::start_auto_build`, so I thought you're calling that function directly :facepalm: Well, at least as a method on the builder such confusion won't happen anymore :) Please fix the CI issue and I'll merge it.

Kobzol (approved):
Thanks!

PR #410, opened 2025-08-15
Improve auto build startup recovery

Sakib25800:
Merge conflict => update PR mergeable state and post comment 
GitHub error => by the time we get to this point the request would have been retried so remove the PR from the queue 
Database error => abort the tick and hope that this doesn't happen again

### DIFF
--- a/src/bors/merge_queue.rs
+++ b/src/bors/merge_queue.rs
@@ -115,7 +115,7 @@ async fn process_repository(repo: &RepositoryState, ctx: &BorsContext) -> anyhow
                     );
 
                     ctx.db
-                        .update_pr_mergeable_state(&pr, MergeableState::HasConflicts)
+                        .update_pr_mergeable_state(&pr, MergeableState::Unknown)
                         .await?;
                     repo.client
                         .post_comment(pr.number, merge_conflict_comment(&gh_pr.head.name))
@@ -825,7 +825,7 @@ mod tests {
                 .await;
             tester.start_auto_build(()).await?;
             tester
-                .wait_for_pr((), |pr| pr.mergeable_state == MergeableState::HasConflicts)
+                .wait_for_pr((), |pr| pr.mergeable_state == MergeableState::Unknown)
                 .await?;
             Ok(())
         })


src/bors/merge_queue.rs
```diff
@@ -100,20 +102,39 @@ async fn process_repository(repo: &RepositoryState, ctx: &BorsContext) -> anyhow
...
                     continue;
                 }
-                Err(error) => {
-                    // Unexpected error - the PR will remain in the "queue" for a retry.
-                    tracing::error!("Error starting auto build for PR {pr_num}: {:?}", error);
+                Err(StartAutoBuildError::GitHubError(error)) => {
+                    tracing::debug!(
+                        "Failed to start auto build for PR {pr_num} due to a GitHub error: {error:?}"
+                    );
+                    // TODO: Remove the PR from the queue (somehow...)
```
Sakib25800: @Kobzol I'm not sure how you want to go about this. 

If there's a GitHub error, it could mean either something is wrong with GitHub or there is a legitimate issue with this PR. Either way, I'm assuming we want to just remove this PR from the queue, but I'm not sure how.

I guess we could just lower the priority of this PR and throw it to the back of the queue? Of the two possibilities, it's likely that there's just an issue with GitHub, so maybe in the time it takes for this PR to reach the front again the issue will have resolved (somehow).

Kobzol: If there is a legitimate issue with a PR, we should detect that issue, and handle it explicitly. If this happens in practice, we'll just have to scan the logs and modify bors to handle that case. I would start with that approach, only if we saw that this happens often for various weird reasons, then we could do what you suggest (mark the PR as failed). But I assume that most of the time this will actually be a network error, so in that case I would just skip the tick.

So the `continue;` here should become `break;`, at least for now.

src/bors/merge_queue.rs
```diff
@@ -100,20 +102,39 @@ async fn process_repository(repo: &RepositoryState, ctx: &BorsContext) -> anyhow
...
-                Ok(false) => {
-                    tracing::debug!("Failed to start auto build for PR {pr_num}");
+                Err(StartAutoBuildError::MergeConflict) => {
+                    let gh_pr = repo.client.get_pull_request(pr.number).await?;
+
+                    tracing::debug!(
+                        "Failed to start auto build for PR {pr_num} due to merge conflict"
+                    );
+
+                    ctx.db
```
Kobzol, line 117: I wonder if marking it as unknown would be safer, as in theory there could be a race condition where the PR gets updated in-between the merge attempt and this line (or we could receive a webhook about it being updated, and the merge queue would overwrite the result from a webhook handler). But this should be incredibly rare, and I think that we update mergeability after every bors comment on a PR, so it should be easy to "refresh" a PR.

src/bors/merge_queue.rs
```diff
@@ -183,16 +204,32 @@ async fn process_repository(repo: &RepositoryState, ctx: &BorsContext) -> anyhow
...
     ctx: &BorsContext,
-    pr: PullRequestModel,
-) -> anyhow::Result<bool> {
+    pr: &PullRequestModel,
+) -> anyhow::Result<(), StartAutoBuildError> {
     let client = &repo.client;
 
-    let gh_pr = client.get_pull_request(pr.number).await?;
-    let base_sha = client.get_branch_sha(&pr.base_branch).await?;
+    let gh_pr = client
```
Kobzol, line 224: If the mergeability status in the GH representation of the PR here tells us that there is a conflict, we might not even try to start the auto build :thinking: That being said, probably it's safer to try anyway, because the mergeability status on the PR could be inaccurate.

src/bors/merge_queue.rs
```diff
@@ -210,27 +247,32 @@ async fn start_auto_build(
...
-    .await?
+    .await
     {
-        MergeResult::Conflict => return Ok(false),
-        MergeResult::Success(merge_sha) => merge_sha,
+        Ok(MergeResult::Success(merge_sha)) => merge_sha,
+        Ok(MergeResult::Conflict) => {
+            return Err(StartAutoBuildError::MergeConflict);
+        }
+        Err(_) => return Err(StartAutoBuildError::MergeConflict),
```
Kobzol: This could be a network error, we shouldn't treat it as a merge conflict. So we should map this to `GitHubError`.

src/bors/merge_queue.rs
```diff
@@ -210,27 +247,32 @@ async fn start_auto_build(
...
         .attach_auto_build(
-            &pr,
+            pr,
             AUTO_BRANCH_NAME.to_string(),
             merge_sha.clone(),
             base_sha,
         )
-        .await?;
+        .await
+        .map_err(StartAutoBuildError::DatabaseError)?;
```
Kobzol, line 274: Could you add a comment below this line that says something like "Below this line, this function should always return `Ok`, because the auto build was started and recorded in the DB".

src/bors/merge_queue.rs
Sakib25800: Sanity checks are on my to-do list for the next PR.

src/bors/merge_queue.rs
Sakib25800: Hmm. Yeah, setting it to unknown would be safer I think. 

Though, I'm not sure what the posted comment should say?

Kobzol: Well the comment should still be the same, that there was a merge conflict. How we mark it in the DB is just an implementation detail of bors.

src/bors/merge_queue.rs
Kobzol: Ok. In general when we try to merge PR, we should check that it's still approved/open/etc., you know the drill :grin:

PR #413, opened 2025-08-24
Add merge queue sanity checks

Sakib25800:
- Add sanity checks before starting an auto build

### DIFF
--- a/src/bors/merge_queue.rs
+++ b/src/bors/merge_queue.rs
@@ -12,10 +12,10 @@ use crate::bors::comment::{
     merge_conflict_comment,
 };
 use crate::bors::{PullRequestStatus, RepositoryState};
-use crate::database::{BuildStatus, MergeableState, PullRequestModel};
+use crate::database::{BuildStatus, MergeableState, OctocrabMergeableState, PullRequestModel};
 use crate::github::api::client::{CheckRunOutput, GithubRepositoryClient};
 use crate::github::api::operations::{BranchUpdateError, ForcePush};
-use crate::github::{CommitSha, MergeError};
+use crate::github::{CommitSha, MergeError, PullRequest};
 use crate::utils::sort_queue::sort_queue_prs;
 
 enum MergeResult {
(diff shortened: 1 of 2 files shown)


Sakib25800:
I don't think there are really any checks worth adding before marking a build as `BuildStatus::Success` in `workflow.rs`. If you think there are checks worth adding, then I'll add them.

src/bors/merge_queue.rs
```diff
@@ -211,6 +215,23 @@ pub enum StartAutoBuildError {
...
+    SanityCheckFailed(anyhow::Error),
+}
+
+async fn verify_pr_state(gh_pr: &PullRequest, pr: &PullRequestModel) -> anyhow::Result<()> {
+    anyhow::ensure!(
+        gh_pr.mergeable_state == OctocrabMergeableState::Clean,
+        "PR not mergeable"
+    );
+    anyhow::ensure!(gh_pr.status == PullRequestStatus::Open, "PR not opened");
+    anyhow::ensure!(
```
Kobzol: Could you please split the error? To have a different error message when the PR is not approved at all, and when the approved SHA doesn't match the head SHA.

src/bors/merge_queue.rs
```diff
@@ -225,10 +246,12 @@ async fn start_auto_build(
         .get_pull_request(pr.number)
         .await
         .map_err(StartAutoBuildError::GitHubError)?;
-    let base_sha = client
-        .get_branch_sha(&pr.base_branch)
+    let base_sha = gh_pr.base.sha.clone();
```
Kobzol: Wait, this is actually a really big change. Does `gh_pr.base.sha` really contain the *latest* SHA of the master branch? I worry that it could be outdated (https://github.com/orgs/community/discussions/39880, https://github.com/orgs/community/discussions/59677).

Could you please check it on some test repo if the value is always up-to-date, even if the base branch updates while the PR is open?

Sakib25800: Nice catch! `gh_pr.base.sha`is the SHA of the target branch at the time the PR was created and does not get updated.

Reverted.

Kobzol (approved):
Thanks! As one of the next steps, could you please unify the `attempt_merge` functions between auto/try builds?

PR #415, opened 2025-08-25
Unify `attempt_merge()`

Sakib25800:
- Unify the separates auto and try attempt merge functions

### DIFF
--- a/src/bors/handlers/info.rs
+++ b/src/bors/handlers/info.rs
@@ -7,7 +7,7 @@ use std::sync::Arc;
 
 pub(super) async fn command_info(
     repo: Arc<RepositoryState>,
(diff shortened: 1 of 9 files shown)


Kobzol (commented):
Thanks! Could you also please generalize `create_merge_commit_message` for auto builds (it should already be prepared for that, just add `Auto` variant to `MergeType`) and use it in `merge_queue`? You can move `create_merge_commit_message` to `bors/mod.rs`, since it will be shared now.

Kobzol (approved):
I was wondering if we could get rid of the clones, not so much because of performance, but just to avoid making a couple more things cloneable unnecessarily. And it seems to work, so I pushed the change.

Thank you!

PR #422, opened 2025-08-26
Refeactor merge queue tests and change how the merge queue is executed

Kobzol:
I went through the merge queue tests to examine what they are doing. While doing that, I found a lot of suboptiomal things in the test suite API, particularly around error reporting and error handling. I ran into many deadlocks and timeouts, so I tried to make the errors from the test suite be as nice as possible.

Apart from refactoring tests, I also changed how the merge queue is executed. Now the queue is checked every 5 seconds. If there was an interesting event in the meantime, or at least 30s has elapsed since the last tick, a new tick is executed. This provides an upper bound on the merge queue execution (it won't run more than once per 5 seconds), which is useful to avoid overloading the network APIs and to handle transient errors robustly.

It is also better for tests. Before, the merge queue was triggered kind of randomly by various events in tests, which wasn't ideal. Now if we don't call `process_merge_queue` in tests, the queue won't run.

@Sakib25800 Let me know what do you think!

### DIFF
--- a/src/bin/bors.rs
+++ b/src/bin/bors.rs
@@ -32,8 +32,12 @@ const MERGEABILITY_STATUS_INTERVAL: Duration = Duration::from_secs(60 * 10);
 /// How often should the bot synchronize PR state.
 const PR_STATE_PERIODIC_REFRESH: Duration = Duration::from_secs(60 * 10);
 
-/// How often should the bot process the merge queue.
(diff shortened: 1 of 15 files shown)


Kobzol:
Yeah, I tried to make the helper functions more atomic, to have more control over what happens, but that also means that even a simple merge queue tests contains a bunch of function calls.

PR #428, opened 2025-09-08
Add `@bors retry`

Sakib25800:
If PR is approved and there is a previously failed auto build, it will be wiped so that the merge queue can pick it up again, 
else it will send an error comment. 

I believe the old bors had the ability to retry try builds as well, so this is a difference that should be noted. However, I've seen no one really use it for retrying try builds.

### DIFF
--- a/src/bors/command/mod.rs
+++ b/src/bors/command/mod.rs
@@ -129,4 +129,6 @@ pub enum BorsCommand {
     OpenTree,
     /// Set the tree closed with a priority level.
     TreeClosed(Priority),
+    /// Retry a previously run (auto) build.
+    Retry,
 }
--- a/src/bors/handlers/help.rs
+++ b/src/bors/handlers/help.rs
@@ -39,6 +39,7 @@ fn format_help() -> &'static str {
         BorsCommand::SetRollupMode(_) => {}
         BorsCommand::OpenTree => {}
         BorsCommand::TreeClosed(_) => {}
+        BorsCommand::Retry => {}
     }
 
     r#"
@@ -65,6 +66,7 @@ You can use the following commands:
     - Optionally, you can specify a `<parent>` SHA with which will the PR be merged. You can specify `parent=last` to use the same parent SHA as the previous try build.
     - Optionally, you can select a comma-separated list of CI `<jobs>` to run in the try build.
 - `try cancel`: Cancel a running try build
+- `retry`: Retry a previously run (auto) build
 - `info`: Get information about the current PR
 
 ## Repository management
@@ -109,6 +111,7 @@ mod tests {
                 - Optionally, you can specify a `<parent>` SHA with which will the PR be merged. You can specify `parent=last` to use the same parent SHA as the previous try build.
                 - Optionally, you can select a comma-separated list of CI `<jobs>` to run in the try build.
             - `try cancel`: Cancel a running try build
+            - `retry`: Retry a previously run (auto) build
             - `info`: Get information about the current PR
 
             ## Repository management
--- a/src/bors/handlers/retry.rs
+++ b/src/bors/handlers/retry.rs
@@ -1,0 +1,95 @@
+use std::sync::Arc;
+
+use crate::PgDbClient;
+use crate::bors::handlers::{PullRequestData, deny_request, has_permission};
+use crate::bors::merge_queue::MergeQueueSender;
+use crate::bors::{Comment, RepositoryState};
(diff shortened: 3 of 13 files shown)


src/bors/command/mod.rs
```diff
@@ -129,4 +129,6 @@ pub enum BorsCommand {
     OpenTree,
     /// Set the tree closed with a priority level.
     TreeClosed(Priority),
+    /// Retry a previously run (auto) build.
```
Kobzol: The name of the command is actually a bit misleading. I think it would be better to explain more accurately what it does.

```suggestion
    /// Clear a failed auto build status from an approved PR. This will cause the merge queue to attempt to start a new auto build and retry merging the PR again.
```

src/bors/handlers/help.rs
```diff
@@ -65,6 +66,7 @@ You can use the following commands:
     - Optionally, you can specify a `<parent>` SHA with which will the PR be merged. You can specify `parent=last` to use the same parent SHA as the previous try build.
     - Optionally, you can select a comma-separated list of CI `<jobs>` to run in the try build.
 - `try cancel`: Cancel a running try build
+- `retry`: Retry a previously run (auto) build
```
Kobzol: Same here.

src/bors/handlers/retry.rs
```diff
@@ -0,0 +1,95 @@
...
+use crate::permissions::PermissionType;
+
+pub(super) async fn command_retry(
+    repo_state: Arc<RepositoryState>,
+    db: Arc<PgDbClient>,
+    pr: PullRequestData<'_>,
+    author: &GithubUser,
+    merge_queue_tx: &MergeQueueSender,
+) -> anyhow::Result<()> {
+    if !has_permission(&repo_state, author, pr, PermissionType::Try).await? {
```
Kobzol: I think that with the "new" implementation that is only for auto builds, it makes more sense to check the `review` permission instead. The set of people who have only the try permission and not the review permission is small anyway.

src/bors/handlers/retry.rs
```diff
@@ -0,0 +1,95 @@
...
+        deny_request(&repo_state, pr.number(), author, PermissionType::Try).await?;
+        return Ok(());
+    }
+
+    let pr_model = pr.db;
+
+    if pr_model.stalled() {
```
Kobzol:
```suggestion
    if pr_model.is_stalled() {
```
To keep consistency with `is_approved` and similar methods.

src/bors/handlers/retry.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+async fn notify_of_invalid_retry_state(
+    repo: &RepositoryState,
+    pr_number: PullRequestNumber,
+) -> anyhow::Result<()> {
+    repo.client
+        .post_comment(
+            pr_number,
+            Comment::new(":exclamation: This PR must be approved and have a previously failed auto build to be retried".to_string())
```
Kobzol:
```suggestion
            Comment::new(":exclamation: You can only retry pull requests that are approved and have a previously failed auto build".to_string())
```

src/bors/handlers/retry.rs
```diff
@@ -0,0 +1,95 @@
...
+        run_test(pool, async |tester: &mut BorsTester| {
+            tester.approve(()).await?;
+            tester.start_auto_build(()).await?;
+            tester
+                .workflow_full_failure(tester.auto_branch().await)
+                .await?;
+            tester.expect_comments((), 1).await;
+            tester.post_comment(Comment::from("@bors retry")).await?;
+            tester.process_merge_queue().await;
+            tester.expect_comments((), 1).await;
```
Kobzol: Let's check exactly what the comment is, and make the test "end-to-end", i.e. actually finish the merge and check that it was completed succesfully.

src/database/mod.rs
```diff
@@ -381,6 +381,15 @@ impl PullRequestModel {
             ApprovalStatus::NotApproved => None,
         }
     }
+
+    /// Approved but with a failed auto build.
+    pub fn stalled(&self) -> bool {
```
Kobzol: We should document this in one of the design/doc markdown files, but it doesn't need to happen in this PR, we are kind of lagging with updating these doc files anyway.

Sakib25800: Will add in future.

Kobzol (approved):
Thank you!

PR #429, opened 2025-09-09
Enable Bors to process commands from PR description when opened

mostafa630:
This PR introduces a new functionality by using  `process_pr_description_commands` that allows commands written in the PR description to be processed in the same way as regular pull request comments.

- added `create_pr_description_comment` to convert a PR description into a `PullRequestComment`
- reuse the existing `handle_comment` function
- updated `handle_pull_request_opened`  signature to include` ctx: Arc<BorsContext>` and `merge_queue_tx: MergeQueueSender` because  `handle_comment`  function needs them.
- call the `process_pr_description_commands` inside the `handle_pull_request_opened`


Kobzol:
Hi, thanks for the PR! I thought that we'd generate the PR comment in the webhook handler itself, so that the rest of the codebase doesn't even know about where the comment is coming from, but this also works.

Could you please add a test for this? You can copy-paste `enqueue_prs_on_pr_opened`, and in the `open_pr` closure, set something to the PR's description (e.g. start a try build or set the PR's priority), and then check that bors reacted to it. You can check other tests for examples, look for `@bors` commands in them.

PR #432, opened 2025-09-16
Add failed count to queue stats

Sakib25800:
A failed PR is a PR with a failed build - either try or auto.

### DIFF
--- a/src/bors/merge_queue.rs
+++ b/src/bors/merge_queue.rs
@@ -145,6 +145,7 @@ async fn process_repository(repo: &RepositoryState, ctx: &BorsContext) -> anyhow
             QueueStatus::ReadyForMerge(approval_info, auto_build) => {
                 handle_successful_build(repo, ctx, &pr, &auto_build, &approval_info, pr_num)
                     .await?;
+                break;
             }
             QueueStatus::Approved(..) => {
                 if handle_start_auto_build(repo, ctx, &pr, pr_num).await? {


src/database/mod.rs
```diff
@@ -390,6 +390,17 @@ impl PullRequestModel {
                 .as_ref()
                 .is_some_and(|build| build.status.is_failure())
     }
+
+    /// Failed try or auto build.
+    pub fn is_failed(&self) -> bool {
+        self.auto_build
+            .as_ref()
+            .is_some_and(|build| build.status.is_failure())
+            || self
```
Kobzol: We don't clear try build status after a push happens on a PR, IIRC, and in general in new bors try and auto builds are completely separated (as they should be), which was not the case in homu. So I wouldn't take try builds into account at all here, and just consider auto builds.

src/github/server.rs
```diff
@@ -150,16 +150,20 @@ async fn queue_handler(
...
-        } else {
-            0
-        };
-        (approved + is_approved, rolled_up + is_rolled_up)
-    });
+    let (approved_count, rolled_up_count, failed_count) =
+        prs.iter()
+            .fold((0, 0, 0), |(approved, rolled_up, failed), pr| {
+                (
+                    approved + if pr.is_approved() { 1 } else { 0 },
```
Kobzol: Note (not necessarily for this PR): I don't like the fact that we have several ways of determining a PR "status", but it only returns atomic information and it can be misleading.

As an example, here we can actually have the counts not "add up" to the total number of PRs, because you can have a PR that is both approved and failed, so it will be counted twice. I always found this very confusing in the homu queue page.

I think that a better way of representing the PR statuses in the queue page in a way where the statuses are mutually exclusive. So something like this:
```rust
enum QueueStatus {
    /// Approved + auto build running
    Pending(ApprovalInfo, BuildModel)
    /// Approved + failed auto build
    Stalled(ApprovalInfo, BuildModel),
    /// Approved, no auto build started yet or a failed auto build was reset with `@bors retry`
    Approved,
    NotApproved
}
```

Ideally we would also use this combined status in the merge queue itself (there will be some duplication with the SQL query, but that's fine).

src/bors/handlers/workflow.rs
```diff
@@ -407,13 +409,18 @@ pub async fn maybe_cancel_auto_build(
...
 ) -> anyhow::Result<Option<String>> {
-    let Some(auto_build) = &pr.auto_build else {
-        return Ok(None);
-    };
-    if auto_build.status != BuildStatus::Pending {
-        return Ok(None);
+    match pr.queue_status() {
+        QueueStatus::Pending(_, _) => {
+            // Auto build is in progress, proceed with cancellation
+        }
```
Kobzol:
```suggestion
    let auto_build = match pr.queue_status() {
        QueueStatus::Pending(_, build) => {
            // Auto build is in progress, proceed with cancellation
            build
        }
```
to avoid the `unwrap` below.

src/bors/merge_queue.rs
```diff
@@ -130,112 +133,130 @@ async fn process_repository(repo: &RepositoryState, ctx: &BorsContext) -> anyhow
...
+        match pr.queue_status() {
+            QueueStatus::NotApproved => unreachable!(),
+            QueueStatus::Stalled(..) => unreachable!(),
+            QueueStatus::Pending(..) => {
+                // Build in progress - stop queue. We can only have one PR being built
+                // at a time.
+                tracing::info!("PR {pr_num} has a pending build - blocking queue");
+                break;
+            }
+            QueueStatus::Approved(approval_info) => {
```
Kobzol: What do you think about adding `QueueStatus::ReadyForMerge`? It would only be useful for the `merge_queue` itself, not the status page, but it seems weird to represent both "approved and nothing happened yet" and "approved and CI already passed" in `QueueStatus`, it would be harder to make a mistake if we represented the successful auto build case explicitly.

src/bors/merge_queue.rs
```diff
@@ -130,112 +133,130 @@ async fn process_repository(repo: &RepositoryState, ctx: &BorsContext) -> anyhow
...
                 }
             }
+        }
+    }
+
+    Ok(())
+}
+
+/// Handle a successful auto build by pointing the base branch to the merged commit.
+/// Returns true if the queue should break (success or fatal error), false to continue.
```
Kobzol: It doesn't seem like the function ever returns `false`?

src/github/server.rs
```diff
@@ -150,16 +151,24 @@ async fn queue_handler(
...
-    let (approved_count, rolled_up_count) = prs.iter().fold((0, 0), |(approved, rolled_up), pr| {
-        let is_approved = if pr.is_approved() { 1 } else { 0 };
-        let is_rolled_up = if matches!(pr.rollup, Some(RollupMode::Always)) {
-            1
-        } else {
-            0
-        };
-        (approved + is_approved, rolled_up + is_rolled_up)
-    });
+    let (approved_count, pending_count, stalled_count, rolled_up_count) = prs.iter().fold(
```
Kobzol: Pending will always be exactly 0 or 1, and any pending PR should always be visible at the very top of the queue, so I don't think that we have to render this count explicitly. I think that makes sense to distinguish "in queue" (approved and CI is NOT failed) and "failed" (approved and CI is failed), plus the rollup count.

Kobzol (commented):
Thank you for introducing the `QueueStatus`. Added a few more comments.

src/bors/merge_queue.rs
Sakib25800: My bad, I copied the other function but this one should only need to `continue`. 

BTW, this was `break`ing before but I've made it `continue` now. This is something I was meant to do before as you mentioned - https://github.com/rust-lang/bors/pull/386#issuecomment-3108777508.

Kobzol: Hmm. I would say that we now indeed have a way of handling this kind of robustly with the merge queue run limits, and with the sanity check. That being said, I don't trust GitHub enough to always return us the latest state of PRs "immediately" after a merge, I can imagine a situation where we merge a rollup PR, and then we immediately continue to a next PR that was included in a rollup, and the sanity check will return us that the PR is still open.

Let's say "for added safety", I would still want to be on the safe(r) side, and put a `break` here. If we give GitHub a few extra seconds, hopefully it will "always" be able to catch up.

templates/queue.html
```diff
@@ -58,6 +58,8 @@ <h1>
         {% match pr.queue_status() %}
         {% when Approved(_) %}
         approved
+        {% when ReadyForMerge(_, _) %}
+        approved
```
Kobzol: This state would only be visible for a few seconds usuallly, but I think that we could actually show it, for easier debugging. So maybe `ready for a merge`?

Kobzol (approved):
Thank you! :)

PR #434, opened 2025-09-21
Add workflow links to try build started comments

mostafa630:
This PR introduces support for adding workflow links to the comment that gets created when a try build starts.

### Implementation details

Within `handle_workflow_started`, we now invoke a new function:
`add_workflow_links_to_try_build_start_comment`.

That function performs the following steps:

1.  Retrieve the build using its commit_sha.

2.  Find the PR associated with that build.

3.  Locate the TryBuildStarted comment for that PR.

4.  Fetch the workflows linked to the build.

5.  Retrieve the comment’s current content (via GraphQL).

6.  Append the workflow links to the content to generate the updated version.

7.  Update the comment with this new content.

### tests 
getting the comment content and updating it  require us to extend the `github.rs` mock file to handle them
so I change the logic to check if the request is  query or mutation
then handle each case based on operation name

### DIFF
--- a/src/github/api/client.rs
+++ b/src/github/api/client.rs
@@ -457,12 +457,31 @@ impl GithubRepositoryClient {
             variables: V,
         }
 
+        #[derive(serde::Deserialize, Debug)]
+        struct Error {
+            #[allow(unused)]
+            message: String,
+        }
+
+        #[derive(serde::Deserialize)]
+        struct RawResponse<T> {
+            errors: Option<Vec<Error>>,
+            #[serde(flatten)]
+            result: T,
+        }
+
         let response = self
             .client
-            .graphql::<T>(&Payload { query, variables })
+            .graphql::<RawResponse<T>>(&Payload { query, variables })
             .await
             .context("GraphQL request failed")?;
-        Ok(response)
+
+        let errors = response.errors.unwrap_or_default();
+        if !errors.is_empty() {
+            Err(anyhow::anyhow!("Query ended with error(s): {errors:?}"))
+        } else {
+            Ok(response.result)
+        }
     }
 
     /// Hides a comment on an Issue, Commit, Pull Request, or Gist.
@@ -517,6 +536,11 @@ impl GithubRepositoryClient {
 
         #[derive(Deserialize)]
         struct Output {
+            data: OutputInner,
+        }
+
+        #[derive(Deserialize)]
+        struct OutputInner {
             node: Option<IssueCommentNode>,
         }
 
@@ -527,17 +551,16 @@ impl GithubRepositoryClient {
 
         tracing::debug!(node_id, "Fetching comment content");
 
-        let output: Output =
-            perform_retryable("get_comment_content", RetryMethod::default(), || async {
-                self.graphql::<Output, Variables>(QUERY, Variables { node_id })
-                    .await
-                    .context("Failed to fetch comment content")
-            })
-            .await?;
+        let output = perform_retryable("get_comment_content", RetryMethod::default(), || async {
+            self.graphql::<Output, Variables>(QUERY, Variables { node_id })
+                .await
+                .context("Failed to fetch comment content")
+        })
+        .await?;
 
-        match output.node {
+        match output.data.node {
             Some(comment) => Ok(comment.body),
-            None => anyhow::bail!("No comment found for node_id: {}", node_id),
+            None => anyhow::bail!("No comment found for node_id: {node_id}"),
         }
     }
     pub async fn update_comment_content(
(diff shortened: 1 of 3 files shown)


mostafa630:
@Kobzol
I’ve completed the changes for this PR , Please let me know if any further adjustments are needed.

src/bors/handlers/workflow.rs
```diff
@@ -53,14 +58,63 @@ pub(super) async fn handle_workflow_started(
...
+        payload.url.clone(),
         payload.run_id.into(),
-        payload.workflow_type,
+        payload.workflow_type.clone(),
         WorkflowStatus::Pending,
     )
     .await?;
 
+    if payload.branch == TRY_BRANCH_NAME {
+        add_workflow_links_to_try_build_start_comment(repo, db, payload).await?;
```
Kobzol: We already have the `BuildModel` loaded in this function, we can just pass it to the `add_workflow_links_to_try_build_start_comment` function, so that we don't have to load it again from the DB and deal with error handling.

src/bors/handlers/workflow.rs
```diff
@@ -53,14 +58,63 @@ pub(super) async fn handle_workflow_started(
...
+    };
+    let Some(pr) = db.find_pr_by_build(&build).await? else {
+        tracing::warn!("PR for build not found");
+        return Ok(());
+    };
+    let comments = db
+        .get_tagged_bot_comments(&payload.repository, pr.number, CommentTag::TryBuildStarted)
+        .await?;
+
+    let Some(try_build_comment) = comments.first() else {
```
Kobzol: We should ~always have at most one such comment in the DB, but race conditions can mean that multiple were created. In that case, we should modify the last comment, not the first one. Please:
- Change `first()` to `last()`
- Add `ORDER BY created_at` to the query in `get_tagged_bot_comments`

src/bors/handlers/workflow.rs
```diff
@@ -53,14 +58,63 @@ pub(super) async fn handle_workflow_started(
...
+    };
+
+    let workflows = db.get_workflow_urls_for_build(&build).await?;
+
+    let mut comment_content = repo
+        .client
+        .get_comment_content(&try_build_comment.node_id)
+        .await?;
+
+    let new_content = append_work_flow_links_to_comment(&mut comment_content, workflows);
```
Kobzol:
```suggestion
    let new_content = append_workflow_links_to_comment(&mut comment_content, workflows);
```

src/github/api/client.rs
```diff
@@ -498,6 +498,109 @@ impl GithubRepositoryClient {
...
+                .context("Failed to update comment")
+            })
+            .await?;
+
+        match output.update_comment {
+            Some(updated_comment) => {
+                tracing::debug!("Updated comment content: {}", updated_comment.comment.body);
+                Ok(())
+            }
+            None => anyhow::bail!("Failed to update comment: {}", node_id),
```
Kobzol:
```suggestion
            None => anyhow::bail!("Failed to update comment: {node_id}"),
```

src/bors/comment.rs
```diff
@@ -198,6 +198,19 @@ pub fn try_build_started_comment(
     Comment::new(msg)
 }
 
+pub fn append_work_flow_links_to_comment(
+    comment_content: &mut String,
+    workflow_links: Vec<String>,
+) -> String {
```
Kobzol: This function does not need to return anything, because it already modifies its argument (which is a mutable reference).
```suggestion
) {
```

src/tests/mocks/github.rs
```diff
@@ -108,40 +108,79 @@ async fn mock_graphql(github: Arc<tokio::sync::Mutex<GitHubState>>, mock_server:
...
-                let data: Variables = serde_json::from_value(body.variables).unwrap();
+                    let data: Variables = serde_json::from_value(body.variables).unwrap();
 
-                // We have to use e.g. `blocking_lock` to lock from a sync function.
-                // It has to happen in a separate thread though.
-                std::thread::spawn(move || {
-                    github.blocking_lock().add_hidden_comment(HiddenComment {
-                        node_id: data.node_id,
-                        reason: data.reason,
+                    let response = serde_json::json!({
```
Kobzol, line 191: Here we'll also need to lock the `github` state, find the comment by the node ID, and return its actual content, or update it. I can modify the test infra if you're unsure how to do it, let me know :)

mostafa630:
@Kobzol 
thanks a lot for the review and feedback  ,  I’ve made the suggested changes. I’m not fully sure about the tests and test infra, so it would be great if you could add them. Thanks again for your support!

Kobzol:
Added the tests, and did some test refactoring before that. Thank you, this looks pretty nice!

mostafa630:
Thanks a lot for merging this and for taking care of the tests and final touches  I really appreciate it  ,  happy I could contribute and help move things forward!

PR #442, opened 2025-10-02
Add table interactivity 

Sakib25800:
This PR adds:
- Grouping 
- Sorting 
- Search

I chose to use DataTable as the table library since it had the easiest setup and I found it the most ergonomic for our use case. 

Currently the design looks poor as I plan to do the design as the last step - functionality first.

### DIFF
--- a/templates/queue.html
+++ b/templates/queue.html
@@ -3,11 +3,27 @@
 {% block title %}Bors queue - {{ repo_name }} {% if tree_state.is_closed() %} [TREECLOSED] {% endif %}{% endblock %}
 
 {% block head %}
+<link rel="stylesheet" href="https://cdn.datatables.net/2.3.4/css/dataTables.dataTables.min.css" />
 <style>
     main {
         max-width: 100rem;
         width: 100%;
         margin: 0 auto;
+    }
+
+    .table-wrapper {
+        overflow-x: auto;
+        -webkit-overflow-scrolling: touch;
+    }
+
+    table {
+        min-width: 100%;
+        white-space: nowrap;
+    }
+
+    table th,
+    table td {
+        padding: 0.5rem;
     }
 </style>
 {% endblock %}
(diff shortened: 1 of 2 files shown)


templates/queue.html
```diff
@@ -123,11 +135,47 @@ <h1>
...
 
 <script>
-    let table = new DataTable('#table', {
+    let table = new DataTable("#table", {
         paging: false,
         info: false
     });
+
+    // Extract clean status text from HTML, removing links and "(try)" suffix
+    function getStatusText(html) {
```
Kobzol: It would be a bit cleaner if we set the "raw" status (without `(try)`) on the given table cell HTML element e.g. using a `data-XXX` attribute. To avoid this parsing.

templates/queue.html
```diff
@@ -151,10 +146,34 @@ <h1>
         return text.replace(/\s*\(try\)\s*$/, "").trim();
     }
 
+    // Default table sort order
+    const statusOrder = {
```
Kobzol: Does this essentially override the order that we provide on the backend? :thinking: It's not so simple as just order by status, as it actually contains the actual queue order. It would be nice to keep that.

Sakib25800: So you're saying you want to call `sort_queue()` on the PRs on the backend and not have this frontend sorting? That way the queue ordering reflects the real merge queue.

Kobzol: Yeah. Essentially:
- When you open the page, the table should be pre-sorted by the backend.
- Then if you re-sort the page via UI, you will change the ordering.

I'm not sure if it will be possible to get to the original ordering once you start messing with the individual sort columns, but I don't think that's a huge issue, you can always just refresh the page, and most people won't be resorting anyway.

src/database/mod.rs
```diff
@@ -416,6 +416,21 @@ impl PullRequestModel {
             },
         }
     }
+
+    /// Get the build status to display on the queue page.
+    pub fn status_text(&self) -> String {
```
Kobzol: Could you please move this to `templates.rs`? This is a very specific UI thing related only to a single web page, and shouldn't live in the database module.

It can be just a free function.

src/utils/sort_queue.rs
```diff
@@ -1,63 +1,40 @@
...
-        None => {
-            if pr.is_approved() {
-                1 // Approved but no build - should be prioritized
-            } else {
-                2 // No status
-            }
-        }
+fn get_queue_status_priority(status: &QueueStatus) -> u32 {
+    match status {
+        QueueStatus::ReadyForMerge(_, _) => 0,
```
Kobzol, line 33: Seeing this function, I wonder if we should add the mergeability status into `QueueStatus`, and have something like `QueueStatus::Unmergeable`. Not sure if that would help with anything though, and definitely not for this PR anyway.

Kobzol (approved):
Thank you!

src/utils/sort_queue.rs
Sakib25800: We wouldn't be able to use that variant in the merge queue itself since it filters out un-mergeable PRs.

Kobzol: That's not really different from `NotApproved` or `Stalled`, which should already never be returned from the SQL query, right?

Sakib25800: True. I guess there's no use for that variant (at least now).

PR #451, opened 2025-11-01
Add rollup selection column UI

Sakib25800:
This PR adds UI for selecting PRs for rollups. 
The next PR will add the actual logic.

### DIFF
--- a/templates/queue.html
+++ b/templates/queue.html
@@ -109,7 +109,7 @@
 
         <tbody>
             {% for pr in prs %}
-            <tr>
+            <tr data-rollupable="{{ pr.is_rollupable() }}">
                 <td class="select-checkbox"></td>
                 <td>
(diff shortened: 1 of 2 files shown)


templates/queue.html
```diff
@@ -154,39 +212,135 @@ <h1>
...
             ],
-            order: []
+            order: [],
+            select: {
+                style: "multi",
+                selector: "td.select-checkbox",
+                headerCheckbox: true,
+                selectable: (rowData, tr, index) => {
+                    let table = $("#table").DataTable();
+                    let approvedByIndex = table.column("approved_by:name").index();
```
Kobzol: This should also check that the PR is not `rollup=never`. It would be great if we could calculate the property "is allowed to be rolled up" on the backend, rather than here, and then just read it in the frontend. Maybe we could do that using a `data-` attribute again?

Kobzol: Awesome, I like the new approach much more, thanks.

PR #454, opened 2025-11-04
feat: Don't allow the merging of PRs when they don't have a clean state [WIP]

### LINKED ISSUE #452: Do not allow approving PRs with a merge conflict
Currently, it is possible to `r+` a PR even if it has a merge conflict. If that happens, bors should instead return an error.

Gustrb:
I still have some stuff do do before marking it as "Ready to Review"
- [x] Write some tests
- [ ] Validate locally everything works fine
- [ ] Validate the text message, I added some simple copy, might not be good enough

I am just starting with this so might take some time :) All advice is appreciated

Closes #452

### DIFF
--- a/src/bors/comment.rs
+++ b/src/bors/comment.rs
@@ -277,7 +277,9 @@ Hint: Remove **{keyword}** from this PR's title when it is ready for review.
 }
 
 pub fn approve_non_clean_pr() -> Comment {
-    Comment::new(r":clipboard: Looks like this PR is not ready to be merged,.".to_string())
+    Comment::new(
+        r":clipboard: Looks like this PR is not ready to be merged, ignoring approval.".to_string(),
+    )
 }
 
 pub fn approve_blocking_labels_present(blocking_labels: &[&str]) -> Comment {
--- a/src/bors/handlers/review.rs
+++ b/src/bors/handlers/review.rs
@@ -1221,6 +1221,20 @@ mod tests {
     }
 
     #[sqlx::test]
+    async fn approve_dirty_pr(pool: sqlx::PgPool) {
+        run_test(pool, async |tester: &mut BorsTester| {
+            tester.edit_pr((), |pr| pr.dirty_pull_request()).await?;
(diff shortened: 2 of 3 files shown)


bncer:
@Gustrb I set up local environment by hard coding some required fields.
I can help to setup, if you have some issues.

src/bors/handlers/review.rs
```diff
@@ -81,6 +83,10 @@ async fn check_pr_approval_validity(
         return Ok(Some(approve_non_open_pr_comment()));
     }
 
+    if matches!(pr.github.mergeable_state, MergeableState::Dirty) {
```
Kobzol, line 86: We do indeed treat both `Dirty` and `Blocked` as having conflicts. Could you please construct `database::MergeableState`  from the GH state using its `From` impl, and then check if the resulting state is mergeable? To use the same logic everywhere. Thanks!

src/bors/comment.rs
```diff
@@ -276,6 +276,12 @@ Hint: Remove **{keyword}** from this PR's title when it is ready for review.
     ))
 }
 
+pub fn approve_non_clean_pr() -> Comment {
+    Comment::new(
+        r":clipboard: Looks like this PR is not ready to be merged, ignoring approval.".to_string(),
```
Kobzol, line 281:
```suggestion
        r":clipboard: This pull request has merge conflicts. These have to be resolved before the PR can be approved.".to_string(),
```

src/tests/mocks/pull_request.rs
```diff
@@ -206,6 +206,10 @@ impl PullRequest {
         self.status = PullRequestStatus::Draft;
     }
 
+    pub fn dirty_pull_request(&mut self) {
```
Kobzol, line 209:
```suggestion
    pub fn make_merge_conflict(&mut self) {
```

Kobzol (commented):
Thank you! Left some comments, but otherwise the PR looks great.

PR #457, opened 2025-11-11
Get rid of username notification checker in merge commit messages

bncer:
Closes  #455

Github is going to implement this feature [on December 8th](https://github.blog/changelog/2025-11-07-removing-notifications-for-mentions-in-commit-messages/).

So it should work now and when I run on test repo I got a notification from [commit message](https://github.com/bncer/boring-test/pull/4) and an email received: 
[bncer/boring-test] Check if notification from @opendatavibe happens (2592771)
Inbox
 <notifications@github.com> Unsubscribe
00:34 (13 minutes ago)

PR #458, opened 2025-11-13
Add rollup logic

Sakib25800:
This PR adds the rollup logic. You can select certain PRs and then create a rollup. 

I did not add the "create a similar PR" text in the rollup PR description as I'm not sure it's used - I can add that in another PR if needed. 

The rollup logic works very similarly to [homu's](https://github.com/rust-lang/homu/blob/8ead3f8e30d1273a665d40d682912df727a1c403/homu/server.py#L238).

### DIFF
--- a/src/github/rollup.rs
+++ b/src/github/rollup.rs
@@ -1,12 +1,8 @@
-use crate::bors::RollupMode;
-use crate::database::QueueStatus;
-use crate::templates::{HtmlTemplate, PullRequestStats, QueueTemplate};
-
 use super::GithubRepoName;
 use super::error::AppError;
 use super::server::ServerStateRef;
 use anyhow::Context;
-use axum::extract::{Path, Query, State};
+use axum::extract::{Query, State};
 use axum::http::StatusCode;
 use axum::response::{IntoResponse, Redirect};
 use octocrab::OctocrabBuilder;


src/github/server.rs
```diff
@@ -134,6 +165,234 @@ async fn help_handler(State(state): State<ServerStateRef>) -> impl IntoResponse
...
+            StatusCode::INTERNAL_SERVER_ERROR,
+            format!("Failed to create rollup: {error}"),
+        )
+            .into_response()),
+    }
+}
+
+/// Creates a rollup PR by merging multiple approved PRs into a single branch
+/// in the user's fork, then opens a PR to the upstream repository.
+async fn create_rollup(
```
Sakib25800: This function is quite big. I tried a few ways to split it up into multiple functions, but that involved a lot of parameters being passed around, which I was not a fan of, as well as just being generally harder to follow the rollup logic.

But I can split it up if needed.

src/bin/bors.rs
```diff
@@ -49,6 +49,14 @@ struct Opts {
     #[arg(long, env = "PRIVATE_KEY")]
     private_key: String,
 
+    /// GitHub OAuth client ID for rollups.
+    #[arg(long, env = "CLIENT_ID")]
```
Kobzol, line 53: Could you please make these optional? It's already a bit difficult to deploy bors locally, and this would add another roadblock. Making rollups is not strictly needed for testing bors locally, so it would be nice if it worked even without creating an OAuth app. Not to mention that testing OAuth locally is a bit tricky anyway :sweat_smile:

src/bin/bors.rs
```diff
@@ -214,10 +222,12 @@ fn try_main(opts: Opts) -> anyhow::Result<()> {
         }
     };
 
+    let oauth_config = OAuthConfig::new(opts.client_id.clone(), opts.client_secret.clone());
```
Kobzol: Then here, store `Option<OAuthConfig>` in the `ServerState`. If either client ID or secret is set, both have to be set, otherwise it should be an error.

src/github/server.rs
```diff
@@ -134,6 +162,74 @@ async fn help_handler(State(state): State<ServerStateRef>) -> impl IntoResponse
...
+    let client = reqwest::Client::new();
+    let token_response = client
+        .post("https://github.com/login/oauth/access_token")
+        .form(&[
+            ("client_id", state.oauth.client_id()),
+            ("client_secret", state.oauth.client_secret()),
+            ("code", &callback.code),
+        ])
+        .send()
+        .await?
```
Kobzol: Please include some `.context` here, otherwise if the request fails, it might be difficult to trace what network request failed.

src/github/server.rs
```diff
@@ -134,6 +162,74 @@ async fn help_handler(State(state): State<ServerStateRef>) -> impl IntoResponse
...
+#[derive(serde::Deserialize)]
+struct OAuthState {
+    pr_nums: Vec<u32>,
+}
+
+async fn oauth_callback_handler(
+    Query(callback): Query<OAuthCallbackQuery>,
+    State(state): State<ServerStateRef>,
+) -> Result<impl IntoResponse, AppError> {
+    let oauth_state: OAuthState = serde_json::from_str(&callback.state)
```
Kobzol: Please add some `info` or `debug` logging here so that we have logs about the OAuth handler being called.

src/github/server.rs
```diff
@@ -134,6 +162,74 @@ async fn help_handler(State(state): State<ServerStateRef>) -> impl IntoResponse
...
+    // Extract access token
+    let oauth_token_params: HashMap<String, String> =
+        url::form_urlencoded::parse(token_response.as_bytes())
+            .into_owned()
+            .collect();
+    let access_token = oauth_token_params
+        .get("access_token")
+        .ok_or_else(|| anyhow::anyhow!("No access token in response"))?;
+
+    match create_rollup(state, oauth_state.pr_nums, access_token).await {
```
Kobzol: Making the rollup can take a long time. Have you tried what are the timeouts currently configured in axum (I'm not sure). Just to make sure that we won't timeout the HTTP request itself needlessly.

templates/queue.html
```diff
@@ -328,5 +329,30 @@ <h1>
...
     });
+
+    rollupContinueButton.addEventListener("click", () => {
+        const scopes = ["public_repo", "workflow"];
+
+        // Gather PR numbers
+        let selectedRows = table.rows({ selected: true }).nodes().toArray();
+        let nums = selectedRows
+            .map(row => {
+                let numberCell = row.cells[1];
```
Kobzol, line 349: Can we again encode the pull request number with a `data-` attribute on the given cell? To avoid parsing it needlessly from a link.

src/github/server.rs
```diff
@@ -134,6 +162,74 @@ async fn help_handler(State(state): State<ServerStateRef>) -> impl IntoResponse
     })
 }
 
+#[derive(serde::Deserialize)]
+struct OAuthCallbackQuery {
+    code: String,
+    state: String,
+}
+
+#[derive(serde::Deserialize)]
```
Kobzol: Could you please add a comment here that says where to find the definition of this JSON message (in the JS code in a template)? Otherwise it's quite non-obvious where does it come from.

src/github/server.rs
```diff
@@ -134,6 +162,74 @@ async fn help_handler(State(state): State<ServerStateRef>) -> impl IntoResponse
     })
 }
 
+#[derive(serde::Deserialize)]
+struct OAuthCallbackQuery {
+    code: String,
```
Kobzol: Could you please add doc comments that explain what are those two attributes? Thanks.

src/github/server.rs
```diff
@@ -134,6 +162,74 @@ async fn help_handler(State(state): State<ServerStateRef>) -> impl IntoResponse
...
+            .into_owned()
+            .collect();
+    let access_token = oauth_token_params
+        .get("access_token")
+        .ok_or_else(|| anyhow::anyhow!("No access token in response"))?;
+
+    match create_rollup(state, oauth_state.pr_nums, access_token).await {
+        Ok(pr_url) => Ok(Redirect::temporary(&pr_url).into_response()),
+        Err(error) => Ok((
+            StatusCode::INTERNAL_SERVER_ERROR,
```
Kobzol: More a note to myself: Regarding the long rollup creation time: as a follow-up, it would be cool to:
- Start an async background task to create the rollup
- Immediately redirect the user to a special bors page that would show something like "Waiting to create a rollup" with a spinner, and redirect them to the PR once the rollup has been created (maybe communicate that with a websocket or something?)

Because right now, when you create the rollup, you then stare into a "stuck" webpage for sometimes a long time, and it's unclear whether something is happening or if some error has occurred and you should do a F5.

src/github/server.rs
```diff
@@ -213,21 +218,179 @@ async fn oauth_callback_handler(
     }
 }
 
+/// Creates a rollup PR by merging multiple approved PRs into a single branch
+/// in the user's fork, then opens a PR to the upstream repository.
 async fn create_rollup(
```
Kobzol: Could you please move this function (and the corresponding OAuth handler and its serde structs) into a separate `rollup.rs` module? Just to make it easier to look for things related to rollups.

src/github/server.rs
```diff
@@ -213,21 +218,179 @@ async fn oauth_callback_handler(
...
+    for num in pr_nums {
+        match state
+            .db
+            .get_pull_request(
+                &GithubRepoName::new(&repo_owner, &repo_name),
+                (num as u64).into(),
+            )
+            .await?
+        {
+            Some(pr) if pr.approver().is_some() => rollup_prs.push(pr),
```
Kobzol: Maybe we should use the `queue_status` instead and make more states invalid? For example, if a PR has a pending build, I don't think we should allow putting it into a rollup :thinking:

We also have to check the rollup state; `rollup=never` PRs must not be allowed to be rolled up.

src/github/server.rs
```diff
@@ -213,21 +218,179 @@ async fn oauth_callback_handler(
...
+        pr_nums,
+    } = oauth_state;
+
     let gh_client = OctocrabBuilder::new()
         .user_access_token(access_token.to_string())
         .build()?;
     let user = gh_client.current().user().await?;
+    let username = user.login;
+
+    // Ensure user has a fork
```
Kobzol: Another log would be nice here, to see which user is attempting to make the rollup. Actually, it would be great if we could wrap all the logic below in a `tracing::span`, so that any logs/errors are attached to a span that shows the repo, PR numbers and username that were involved in the rollup creation.

src/github/server.rs
```diff
@@ -213,21 +218,179 @@ async fn oauth_callback_handler(
...
+        .map_err(|error| {
+            anyhow::anyhow!("Could not create rollup branch {branch_name}: {error}",)
+        })?;
+
+    let mut successes = Vec::new();
+    let mut failures = Vec::new();
+
+    // Merge each PR's commits into the rollup branch
+    for pr in rollup_prs {
+        let pr_github = gh_client
```
Kobzol: Loading the PRs can take some times, as it's up to e.g. 20 roundtrips to GitHub. As a follow-up, it would be great to do some mini-benchmark to see how long it takes to load e.g. 20 PRs sequentially, vs creating e.g. `FuturesUnordered` or something that can buffer futures and load e.g. 5 PRs at a time, to make this faster, as it can be trivially parallelized.

src/github/server.rs
```diff
@@ -213,21 +218,179 @@ async fn oauth_callback_handler(
...
+        }
+
+        let head_sha = pr_github.head.sha.clone();
+        let merge_msg = format!(
+            "Rollup merge of #{} - {}, r={}\n\n{}\n\n{}",
+            pr.number.0,
+            pr_github.head.ref_field,
+            pr.approver().unwrap_or("unknown"),
+            pr.title,
+            suppress_github_mentions(&pr_github.body.unwrap_or_default())
```
Kobzol: `suppress_github_mentions` is no longer needed to be used here (https://github.com/rust-lang/bors/pull/457).

src/github/server.rs
```diff
@@ -213,21 +218,179 @@ async fn oauth_callback_handler(
...
+                successes.push(pr);
+            }
+            Err(error) => {
+                if let octocrab::Error::GitHub { source, .. } = &error {
+                    if source.status_code == http::StatusCode::CONFLICT {
+                        failures.push(pr);
+                        continue;
+                    }
+
+                    anyhow::bail!("Merge failed due to conflict: {}", source.message);
```
Kobzol: This isn't actually due to a conflict, because that is handled above, right?

src/github/server.rs
```diff
@@ -203,7 +208,7 @@ async fn oauth_callback_handler(
         .get("access_token")
         .ok_or_else(|| anyhow::anyhow!("No access token in response"))?;
 
-    match create_rollup(state, oauth_state.pr_nums, access_token).await {
+    match create_rollup(state, oauth_state, access_token).await {
         Ok(pr_url) => Ok(Redirect::temporary(&pr_url).into_response()),
         Err(error) => Ok((
```
Kobzol: Let's also log the error here, otherwise we'd have no record of it.

Kobzol (commented):
This looks awesome, thank you so much! I went through it and left a bunch of comments.

Apart from the comments, it would be great to have tests for this (doesn't need to be in this PR, let's do it in a follow-up). I'm imagining something like this:
1) The test context will gain a new method, something like `.rollup(<list-of-prs>)` (we can just hardcode the default repo name inside of it).
2) The method will post a request to the `/oauth/callback` endpoint
3) We will have to mock any GitHub requests done by the rollup handler.
4) We wait for the rollup to be done (should be enough to wait for the OAuth callback HTTP request finishing) and then assert that the rollup PR was created.

src/github/server.rs
Kobzol: But maybe if we parallelize the fetching of PRs from GitHub, it will be "fast enough" and this is overkill.

src/github/server.rs
Sakib25800: No timeout has been set so it should be indefinite.

src/github/server.rs
Sakib25800: Will benchmark the parallelisation to see if it's fast enough.

src/github/server.rs
```diff
@@ -182,6 +182,10 @@ async fn oauth_callback_handler(
     Query(callback): Query<OAuthCallbackQuery>,
     State(state): State<ServerStateRef>,
 ) -> Result<impl IntoResponse, AppError> {
+    let oauth_config = state.oauth.as_ref().ok_or_else(|| {
```
Kobzol: This will be returned to the user (or rather GitHub). We should also log it, that will be more useful to the app developer.

src/github/rollup.rs
```diff
@@ -126,8 +126,13 @@ async fn create_rollup(
             )
             .await?
         {
-            Some(pr) if pr.approver().is_some() => rollup_prs.push(pr),
-            Some(_) => anyhow::bail!("PR #{num} is not approved"),
+            Some(pr) => {
+                if !pr.is_rollupable() {
+                    tracing::warn!("PR #{num} cannot be included in rollup, skipping");
```
Kobzol: I think that this should be a hard error.

Kobzol (commented):
Changes look great so far! :)

src/github/rollup.rs
Kobzol: Sorry, I wasn't clear. What I meant was to cancel the rollup and return an error.

Sakib25800: Oh, that makes more sense.

Kobzol:
Thank you! The PR needs a rebase, and maybe a squash (17 commits is a bit too much), but other than that I think I'm ok with merging! We can take a look at parallelizing the PR fetch later. Awesome job! :heart:

Kobzol (approved):
Thank you!

Issue #371, opened 2025-07-20
Display the time of a running build on the queue page


Kobzol:
See https://github.com/rust-lang/homu/issues/64.

Kobzol:
It would also be nice to show expected end time (based on historical average of CI times).

LimpSquid:
@bncer Unfortunately I did not have the time to work on it yet. Feel free to pick it up.

bncer:
@LimpSquid Thank you!

Kobzol:
Sure! I think that @Sakib25800 is working (or planning to work) on other things related to rollups now, so this should be free to be picked up.

Maybe I'd start with showing how long the current build takes. The UX might be a bit tricky to avoid having too much data in the table, but maybe it won't be that wide and it will be fine.

Eventually, it would be cool to compute the historical average, and then display a mini progress bar of expected duration :)

PR #459, opened 2025-11-16
Add elapsed time in pending status

bncer:
Closes [#371](https://github.com/rust-lang/bors/issues/371)

Display elapsed time

Currently it's not live, it shows within template.

### DIFF
--- a/src/templates.rs
+++ b/src/templates.rs
@@ -3,6 +3,7 @@ use crate::database::{
 };
 use askama::Template;
 use axum::response::{Html, IntoResponse, Response};
+use chrono::{DateTime, Local, Utc};
 use http::StatusCode;
 
 /// Build status to display on the queue page.
@@ -67,6 +68,12 @@ pub struct QueueTemplate {
     pub prs: Vec<PullRequestModel>,
     pub tree_state: TreeState,
     pub oauth_client_id: Option<String>,
+}
+
+impl QueueTemplate {
+    fn to_local_time(&self, time: DateTime<Utc>) -> DateTime<Local> {
+        time.into()
+    }
 }
 
 #[derive(Template)]
--- a/templates/queue.html
+++ b/templates/queue.html
@@ -156,7 +156,7 @@
                 {% let pending_build = crate::templates::get_pending_build(pr) %}
                 <td class="status" data-status="{{ crate::templates::status_text(pr) }}"
                     {% if let Some(build) = pending_build %}
-                    title="{{ build.created_at.format("%d.%m.%Y %H:%M:%S") }} UTC"
+                    title="Started at {{ to_local_time(*build.created_at).format("%d.%m.%Y %H:%M:%S") }}"
                     data-created-at="{{ build.created_at.timestamp_millis() }}"
                     {% endif %}>
 


Kobzol:
I think we can do that in several steps (multiple PRs). It is enough to just show the elapsed time in the first PR (this one). Thank you! I will take a look tomorrow.

src/templates.rs
```diff
@@ -68,3 +70,18 @@ pub struct QueueTemplate {
 #[derive(Template)]
 #[template(path = "not_found.html")]
 pub struct NotFoundTemplate {}
+
+pub fn get_pending_build(pr: &PullRequestModel) -> Option<&BuildModel> {
+    if let Some(auto_build) = &pr.auto_build
```
Kobzol, line 84: Let's only show auto builds for now. We don't currently show try builds in the queue page at all. Maybe we can change that in the future, but that would require more changes.

templates/queue.html
```diff
@@ -340,5 +386,40 @@ <h1>
...
+        const remainingSeconds = totalSeconds % 3600;
+        const minutes = Math.floor(remainingSeconds / 60);
+
+        let output = '';
+        if (hours > 0) {
+            output += `${hours}h`;
+        }
+
+        if (minutes > 0) {
+            output += `${minutes}m`;
```
Kobzol, line 429: Would be nice to add a space before the minutes if the hours were already filled in :)

templates/queue.html
```diff
@@ -340,5 +386,40 @@ <h1>
...
+            if (createdAtMs > 0) {
+                const elapsedMs = now - createdAtMs;
+                const formattedTime = formatElapsedTime(elapsedMs);
+                let display = td.querySelector('.elapsed-time-display');
+                display.textContent = formattedTime;
+            }
+        });
+    }
+
+    updateElapsedTimes();
```
Kobzol, line 450: Maybe we could wrap this in a `setInterval` and update it every second, so that it doesn't get stale if you have the website opened for a longer time?

templates/queue.html
```diff
@@ -340,5 +386,40 @@ <h1>
...
+        if (output.length > 0) {
+            return ` (${output})`;
+        }
+        return '';
+    }
+
+    function updateElapsedTimes() {
+        const now = Date.now();
+        document.querySelectorAll('[data-created-at]').forEach(td => {
+            const createdAtMs = parseInt(td.getAttribute('data-created-at'), 10);
```
Kobzol:
```suggestion
            const createdAtMs = parseInt(td.dataset.createdAt, 10);
```

templates/queue.html
```diff
@@ -114,12 +154,18 @@ <h1>
                 <td>
                     <a href="{{ repo_url }}/pull/{{ pr.number }}">{{ pr.number.0 }}</a>
                 </td>
-                <td data-status="{{ crate::templates::status_text(pr) }}">
+                {% let pending_build = crate::templates::get_pending_build(pr) %}
+                <td class="status" data-status="{{ crate::templates::status_text(pr) }}"
+                    data-created-at="{% if let Some(build) = pending_build %}
```
Kobzol: Could you output the attribute only when the build is actually pending? Then the JS code will only iterate the pending PRs. The list of PRs can be otherwise quite long (~thousand), so no need to run the JScode on all of that.

templates/queue.html
```diff
@@ -31,6 +47,30 @@
     td.select-checkbox {
         width: 2.5rem;
     }
+    table.dataTable > tbody > tr > td.status {
+	position: relative;
```
Kobzol: Could you indent the CSS to match the other code around? Thanks!

src/templates.rs
bncer: It's strange, I was able to create with '@bors try' command and it shows.
Screenshot From 2025-11-18 00-44-19

templates/queue.html
bncer: Thank you :-)

src/templates.rs
Kobzol: Ah, I forgot this is a thing, sorry. Well, I would rather delete this for now, tbh :laughing: Let's just deal with auto builds.

templates/queue.html
```diff
@@ -114,12 +154,18 @@ <h1>
                 <td>
                     <a href="{{ repo_url }}/pull/{{ pr.number }}">{{ pr.number.0 }}</a>
                 </td>
-                <td data-status="{{ crate::templates::status_text(pr) }}">
+                {% let pending_build = crate::templates::get_pending_build(pr) %}
+                <td class="status" data-status="{{ crate::templates::status_text(pr) }}"
```
Kobzol, line 157: One more idea 😆 When I hover on the pending build, it would be nice to show when it started. Could you add a `title` for cells with a pending build, which shows the time when it started, e.g. with a `<day>. <month>. <year> <hour>:<month>:<second> format`? Thanks!

bncer: Yeah, I will do it. The only thing I don't understand why <title> should be used on cells? Title will show in `<head>` section.

Kobzol: Sorry, I meant the `title="foo"` attribute (https://www.geeksforgeeks.org/html/html-title-attribute/), not the `<title>` element :)

bncer:
> Could you share the code where that happens?

[in src/database/operations.rs in upsert_pull_request function](https://github.com/rust-lang/bors/blob/3bb5fa203bbd1b8138738371a7d67f646d7b5a5a/src/database/operations.rs#L130C1-L190C2)
and [src/database/operations.rs in update_mergeable_states_by_base_branch function](https://github.com/rust-lang/bors/blob/3bb5fa203bbd1b8138738371a7d67f646d7b5a5a/src/database/operations.rs#L301C1-L352C11)

After several hours I built bors and it worked without any errors. But I tried to reproduce this by: `cargo clean`, `rm -rf .sqlx` and it fails again 😆

As a result `cargo sqlx prepare -- --all-targets`  shows modifications on:
modified:   .sqlx/query-367bf2de86a8a92b56b2be296cef119a7835ab5fd09326265a4a1403d0d6ad19.json
modified:   .sqlx/query-d16ca5631fcdd4216b1bf3a7b1672b8721d72dead28356fbbc69f84029d08132.json
```
"nullable": [
-      false,
-      false,
-      false,
-      false,
-      false,
-      false,
+      true,
+      true,
+      true,
+      true,
+      true,
+      true, ...
```

bncer:
With sqlx's exclamation mark (!) I lowered number of errors, but clear reason of different behavior blows my mind 
```
### diff --git a/src/database/operations.rs b/src/database/operations.rs
index 6d31e9b..bce49ee 100644
--- a/src/database/operations.rs
+++ b/src/database/operations.rs
@@ -151,22 +151,22 @@ pub(crate) async fn upsert_pull_request(
             )
             SELECT
                 pr.id,
-                pr.repository as "repository: GithubRepoName",
+                pr.repository as "repository!: GithubRepoName",
                 pr.number as "number!: i64",
                 pr.title,
                 pr.author,
-                pr.assignees as "assignees: Assignees",
+                pr.assignees as "assignees!: Assignees",
                 (
                     pr.approved_by,
                     pr.approved_sha
                 ) AS "approval_status!: ApprovalStatus",
-                pr.status as "pr_status: PullRequestStatus",
+                pr.status as "pr_status!: PullRequestStatus",
                 pr.priority,
-                pr.rollup as "rollup: RollupMode",
-                pr.delegated_permission as "delegated_permission: DelegatedPermission",
+                pr.rollup as "rollup!: RollupMode",
+                pr.delegated_permission as "delegated_permission!: DelegatedPermission",
                 pr.base_branch,
-                pr.mergeable_state as "mergeable_state: MergeableState",
-                pr.created_at as "created_at: DateTime<Utc>",
+                pr.mergeable_state as "mergeable_state!: MergeableState",
+                pr.created_at as "created_at!: DateTime<Utc>",
                 try_build AS "try_build: BuildModel",
                 auto_build AS "auto_build: BuildModel"
             FROM upserted_pr as pr
```

templates/queue.html
```diff
@@ -90,7 +130,7 @@ <h1>
   </div>
 
   <div class="table-wrapper">
-    <table id="table">
+    <table id="table" class="stripe">
```
Kobzol, line 133: Oh, this was a botched rebase 😱 Thank you for finding that!

templates/queue.html
```diff
@@ -114,12 +153,19 @@ <h1>
                 <td data-pr-number="{{ pr.number.0 }}">
                     <a href="{{ repo_url }}/pull/{{ pr.number }}">{{ pr.number.0 }}</a>
                 </td>
-                <td data-status="{{ crate::templates::status_text(pr) }}">
+                {% let pending_build = crate::templates::get_pending_build(pr) %}
+                <td class="status" data-status="{{ crate::templates::status_text(pr) }}"
+                    {% if let Some(build) = pending_build %}
+                    title="{{ build.created_at.format("%d.%m.%Y %H:%M:%S") }} UTC"
```
Kobzol:
```suggestion
                    title="Started at {{ to_local_time(*build.created_at).format("%d.%m.%Y %H:%M:%S") }}"
```
Just to make it obvious what is the date about. We should also convert it to local time. I tried to do it directly in the template, but it seems.. difficult 😆

Let's add this to Rust:
```
impl QueueTemplate {
    fn to_local_time(&self, time: DateTime<Utc>) -> DateTime<Local> {
        time.into()
    }
}
```
And this to the template:

Issue #468, opened 2025-11-23
Failing to load permissions for repo


bncer:
Command `cargo run ...` building, but fails on running:
```
2025-11-23T22:05:41.099116Z ERROR bors: Error: Failed to load repository bncer/boring-test: Cannot load repository Some("bncer/boring-test"): Could not load permissions for r
epository bncer/boring-test

Caused by:
    Cannot load review users: Cannot load users from team API: reqwest::Error { kind: Status(404, None), url: "https://team-api.infra.rust-lang.org/v1/permissions/bors.boring_test.review.json" }
```
`boring_test` - it's my github repo to test bors

It's because [permissions/bors.rust.review.json](https://team-api.infra.rust-lang.org/v1/permissions/bors.rust.review.json) does not exist for dev environment and json doesn't contain other contributors github UserId.

@Kobzol Can I work on that? 
What method do you suggest to handle local development?

bncer:
I will make as soon as I have a time

PR #477, opened 2025-12-05
Note about private key on local development experience

bncer:
Related to a [comment](https://github.com/rust-lang/bors/issues/468#issuecomment-3568404708)
It's about how to help people to setup local environment in terminal (bash, fish) to test bors.

I found essential part which triggered the fail, it's one `\n`  at the beginning of env var. Main reason for the cause is unknown - it may be terminal providing, arg parser...

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -129,7 +129,7 @@ can still be done if needed.
 2. Run bors locally, and configure environment variables and/or command-line parameters for it:
    - Set `APP_ID` to the ID of the created GitHub app.
    - Set `WEBHOOK_SECRET` to the webhook secret of the app.
-   - Set `PRIVATE_KEY` to the private key of the app.
+   - Set `PRIVATE_KEY` to the private key of the app. **Note:** If you face an error related to the private key, you need to add `\n` at the beginning of environment variable: `\n-----BEGIN RSA PRIVATE KEY`
    - (optional) Set `WEB_URL` to the public URL of the website of the app.
    - (optional) Set `CMD_PREFIX` to the command prefix used to control the bot (e.g. `@bors`).
    - (optional) Set `PERMISSIONS` `"data/permissions"` directory path to list users with permissions to perform try/review.


docs/development.md
```diff
@@ -129,7 +129,7 @@ can still be done if needed.
 2. Run bors locally, and configure environment variables and/or command-line parameters for it:
    - Set `APP_ID` to the ID of the created GitHub app.
    - Set `WEBHOOK_SECRET` to the webhook secret of the app.
-   - Set `PRIVATE_KEY` to the private key of the app.
+   - Set `PRIVATE_KEY` to the private key of the app. **Note:** If you face an error related to the private key, you need to add `\n` at the beginning of environment variable: `\n-----BEGIN RSA PRIVATE KEY`
```
Kobzol, line 132: I tried it in my terminal and it was enough to `export PRIVATE_KEY=<paste from GitHub>`. But the important part was then using `"${PRIVATE_KEY}"` rather than just `${PRIVATE_KEY}` if I explicitly referred to the env. var. later.

This note seems a bit random, to be honest 😅 Did you figure out what was the problem and why did this fix it?

bncer: I tried simply paste without `\n` at env var.  And just put like that: "\n${PRIVATE_KEY}" it worked :smile:
My setup Fedora on bash/fish

Haven't understand the problem. I'm thinking about is it worth dig out further.

bncer: My error message looks following:
`error: unexpected argument '-----BEGIN RSA PRIVATE KEY-----
MII .......
.../Wt
-----END RSA PRIVATE KEY-----' found
`
`
Usage: bors --app-id <APP_ID> --private-key <PRIVATE_KEY> --webhook-secret <WEBHOOK_SECRET> --db <DB>
`

bncer: I found the cause in [a clap test](https://github.com/clap-rs/clap/blob/53a69215e481c237bff5599f93bbcc7bdc6218c5/tests/builder/error.rs#L121C1-L129C41)
`'-----BEGIN...'` treated as a new argument with `--`. It resolves when I provide backslash character escaping `'\-----BEGIN...'` or `'\n-----BEGIN...'`

Clap seems to put additional backslash to special characters - `\\\\-----BEGIN RSA PRIVATE KEY-----\\n`

bncer: You can try to test this by:
`--db '--postgres://bors:bors@localhost:5432/bors'` it will fail the same way :smile:

Kobzol: Good find. With knowledge comes power :) Can you try if https://github.com/rust-lang/bors/pull/478 fixes the issue for you? If yes, we don't need to document anything.

bncer: Thank you! It works smoothly  :smile:

Kobzol: Awesome :)

Issue #486, opened 2025-12-13
Show expected time on CI builds


bncer:
HI!

I will continue on [Display the time of a running build on the queue page](https://github.com/rust-lang/bors/issues/371#issuecomment-3524068014). Specifically, show expected time in form of progress bar.

I need some guidance on source code where to implement.

I did some research and my best guess was to implement [db.update_build_status](https://github.com/rust-lang/bors/blob/aee15fdf9b52117c631e8b2bd68fe56ddc245783/src/bors/handlers/workflow.rs#L271)'s similar logic (naming a new function `update_build_running_time`) and create `running_time` column in `build` table.

PR #487, opened 2025-12-13
Store duration in database

bncer:
Related to #486 

Note:
- running_time from [BorsRepositoryEvent::WorkflowCompleted::WorkflowRunCompleted](https://github.com/rust-lang/bors/blob/9de8ed204ca6283816b6911f106835000a5f57c9/src/server/webhook.rs#L327C1-L334C15), which calculates duration
- Currently in this PR try builds written to the `duration` column of the `build` table for possible future use. Try build duration records should not effect performance, we will filter out only successfully completed auto builds.


migrations/20260113181302_add_duration_to_build.up.sql
```diff
@@ -0,0 +1,2 @@
+-- Add up migration script here
+ALTER TABLE build ADD COLUMN duration INTERVAL;
```
Kobzol, line 2: Huh, first time I see this (Postgre)SQL data type. Cute.

src/bors/handlers/workflow.rs
```diff
@@ -269,6 +269,8 @@ async fn maybe_complete_build(
     };
 
     db.update_build_status(&build, status).await?;
+    db.update_build_duration(&build, payload.running_time)
```
Kobzol: Let's modify the duration in the `db.update_build_status` call, to avoid doing two back-to-back SQL queries unnecessarily. I plan to extend the `update_build_status` function in the near future anyway.

Kobzol (commented):
Hi, thank you for the PR! Could you also add code that loads the duration from the DB into the build model? Just so that we can check that the `Interval` type can be correctly deserialized into Rust.

src/bors/handlers/workflow.rs
bncer: I was about to add duration in db.update_build_status, but saw this function was used elsewhere which changes only status.

Kobzol: That's fine, it can just be `Option<Duration>`.

bncer:
Sorry. I'm busy this week. I will do as soon as I have time

bncer:
I also tried to look into marcos `sqlx::query_as!` with `cargo expand` but couldn't do it properly so far.

bncer:
WIP. Do not review immediately, please. 
It was a bit tricky to convert database's interval to rust's duration.
Maybe I shouldn't touch the extension of update_build_{status, duration}, but I tried to show you my version of solution.
I will wrap up everything and finish as soon as possible.

src/database/operations.rs
```diff
@@ -643,15 +646,23 @@ WHERE repository = $1
...
 ) -> anyhow::Result<()> {
-    measure_db_query("update_build_status", || async {
+    measure_db_query("update_build_column", || async {
         sqlx::query!(
-            "UPDATE build SET status = $1 WHERE id = $2",
-            status as _,
+            r#"
+UPDATE build
+SET
+    status   = COALESCE($1, status),
```
Kobzol:
```suggestion
    status   = $1,
```

src/database/mod.rs
```diff
@@ -376,6 +387,45 @@ impl sqlx::Decode<'_, sqlx::Postgres> for BuildKind {
...
+        let days = total_us / 86_400_000_000;
+        let microseconds = total_us % 86_400_000_000;
+
+        let interval = PgInterval {
+            months: 0,
+            days: days as i32,
+            microseconds,
+        };
+
+        interval.encode_by_ref(buf)
```
Kobzol:
```suggestion
        self.0.encode_by_ref(buf)
```
`sqlx` implements `Encode` (but for some reason not `Decode`) for `Duration` directly.

src/bors/build_queue.rs
```diff
@@ -95,8 +95,16 @@ pub async fn handle_build_queue_event(
                         // First try to complete builds, and only then timeout then
                         // Because if the bot was offline for some time, we want to first attempt to
                         // actually finish the build, otherwise it might get instantly timeouted.
-                        if !maybe_complete_build(&repo, db, &build, &pr, &merge_queue_tx, None)
-                            .await?
+                        if !maybe_complete_build(
```
Kobzol, line 98: It's a shame to lose the build duration if we complete it from here, and not the workflow run completed event. Furthermore, to be fully general, we should ideally support having multiple workflows per build, not just one.

A more general solution would be to read the workflow runs in `maybe_complete_build`, and figure out the duration from them (take the minimum of start, maximum of end, and subtract). I'd store `Option<Duration>` in `WorkflowRun`, and fill it in `get_workflow_runs_for_commit_sha` if the workflow has completed (either successfully or failed). You can compute the duration by subtracting `run.updated_at - run.created_at`.

Kobzol (commented):
Thank you! Left some comments to simplify the `Interval` operations and about the actual build duration logic.

src/bors/build_queue.rs
bncer: Yeah, most what you are proposing I understood.

Issue #284, opened 2025-05-23
Think about handling PR rollup merge race conditions


Kobzol:
Just a thought I had today: if we have a PR A and a PR B, where B is a rollup containing A, we should make sure that once B is merged, bors won't start immediately merging A, in case the notification that A has been merged arrives too late and bors is very eager. It would be useful to check what homu does to prevent this (maybe it just sleeps for a minute after merging a PR, or something? :) ).

jieyouxu:
I think the answer for homu(?) is it doesn't, rollup people have to be very careful not to include merged PRs or overlapping PRs.

Kobzol:
I think that with the PR sanity checks that we have now, this should be fine, because bors will check whether a PR hasn't been merged immediately before trying to merge it.

It would be still nice to track PR <-> rollup associations to make creating rollups easier though, but that's a for separate issue.

PR #519, opened 2025-12-22
feat(rollups): add rollup_member mapping

### LINKED ISSUE #494: Store mapping of PRs to rollups
When we create a rollup, it would be great to remember in the DB which PRs are a part of it. So that we can provide some visualization of which PRs are currently in a rollup in the queue page.

This would likely be a new DB table (e.g. `rollup_contents`) that would associate a given rollup PR to another PR that is contained in it. Something like (`repository TEXT NOT NULL, rollup_pr BIGINT NOT NULL, member_pr BIGINT NOT NULL`), where `(rollup_pr, member_pr`) should be unique. Also some index would likely be helpful, so that we can quickly find all rollups in which a given PR is located.

Carbonhell:
Closes #494.

I wasn't sure whether to use IDs from the pull_request table to represent PRs in the mapping, or just the GitHub PR number - I went with the latter as there would be no record for the rollup at the time of its creation, as that gets populated via GitHub events, if I understood correctly.

I haven't experienced rollups at all as I haven't (yet?) contributed to a project with bors, but I noticed there's no safety check for multiple rollups containing the same PR(s) being opened. I found a related comment [here](https://github.com/rust-lang/bors/issues/284#issuecomment-2905353395), but I kept it out of this PR to reduce its scope and simplify review. I think such a safety check could be added easily by checking for rollups associated to each candidate PR, and verifying the status of the rollup PR to ignore closed ones. That would still allow one to recreate the same scenario by reopening a previously closed rollup, though.

One thing I'm unsure about is how to handle the error when writing the rollup_contents records in the DB. I think that should only happen either in case of DB connection errors, or if the list of PRs contains duplicates. I assume the former can be propagated to the caller, whereas I'm not sure if the latter is realistic and should be accounted for. I'm thinking we can just ignore the conflicting records in such case - let me know if I should proceed with that.

Edit: sorry, I forgot the migration test files. Will push them tomorrow as soon as possible

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/.sqlx/*.json binary linguist-generated=true
--- a/src/database/client.rs
+++ b/src/database/client.rs
@@ -13,9 +13,13 @@ use super::{
     ApprovalInfo, DelegatedPermission, MergeableState, PrimaryKey, RunId, UpdateBuildParams,
     UpsertPullRequestParams,
 };
+use std::collections::{HashMap, HashSet};
+
 use crate::bors::comment::CommentTag;
 use crate::bors::{BuildKind, PullRequestStatus, RollupMode};
-use crate::database::operations::update_pr_auto_build_id;
+use crate::database::operations::{
+    get_nonclosed_rollups, register_rollup_pr_member, update_pr_auto_build_id,
+};
 use crate::database::{
     BuildModel, CommentModel, PullRequestModel, RepoModel, TreeState, WorkflowModel,
     WorkflowStatus, WorkflowType,
@@ -370,6 +374,32 @@ impl PgDbClient {
 
     pub async fn delete_tagged_bot_comment(&self, comment: &CommentModel) -> anyhow::Result<()> {
         delete_tagged_bot_comment(&self.pool, comment.id).await
+    }
+
+    /// Register the members of a rollup.
+    /// All records are inserted in a single transaction.
+    pub async fn register_rollup_members(
+        &self,
+        rollup: &PullRequestModel,
+        members: &[PullRequestModel],
+    ) -> anyhow::Result<()> {
+        let mut tx = self.pool.begin().await?;
+        for member in members {
+            assert_ne!(rollup.id, member.id);
+            assert_eq!(rollup.repository, member.repository);
+            register_rollup_pr_member(&mut *tx, rollup, member).await?;
+        }
+        tx.commit().await?;
+        Ok(())
+    }
+
+    /// Returns a map of rollup PR numbers to the set of member PR numbers that are part of that rollup.
+    /// Only returns non-closed rollup PRs.
+    pub async fn get_nonclosed_rollups(
+        &self,
+        repo: &GithubRepoName,
+    ) -> anyhow::Result<HashMap<PullRequestNumber, HashSet<PullRequestNumber>>> {
+        get_nonclosed_rollups(&self.pool, repo).await
     }
 }
 
--- a/src/database/operations.rs
+++ b/src/database/operations.rs
@@ -1,6 +1,7 @@
 use chrono::DateTime;
 use chrono::Utc;
 use sqlx::postgres::PgExecutor;
+use std::collections::{HashMap, HashSet};
 
 use super::ApprovalStatus;
 use super::Assignees;
@@ -1087,6 +1088,59 @@ pub(crate) async fn clear_auto_build(
         .execute(executor)
         .await?;
         Ok(())
+    })
+    .await
+}
+
+pub(crate) async fn register_rollup_pr_member(
+    executor: impl PgExecutor<'_>,
+    rollup: &PullRequestModel,
+    member: &PullRequestModel,
+) -> anyhow::Result<()> {
+    measure_db_query("register_rollup_pr_member", || async {
+        sqlx::query!(
+            r#"
+        INSERT INTO rollup_member (rollup, member)
+        VALUES ($1, $2)
+        "#,
+            rollup.id,
+            member.id
+        )
+        .execute(executor)
+        .await?;
+        Ok(())
+    })
+    .await
+}
+
+pub(crate) async fn get_nonclosed_rollups(
+    executor: impl PgExecutor<'_>,
+    repo: &GithubRepoName,
+) -> anyhow::Result<HashMap<PullRequestNumber, HashSet<PullRequestNumber>>> {
+    measure_db_query("get_nonclosed_rollups", || async {
+        let mut stream = sqlx::query!(
+            r#"
+        SELECT rollup.number AS rollup_num, member.number AS member_num
+        FROM rollup_member rm
+            JOIN pull_request rollup ON rollup.id = rm.rollup
+            JOIN pull_request member ON member.id = rm.member
+        WHERE rollup.status IN ('open', 'draft') AND
+              rollup.repository = $1
+            "#,
+            repo as &GithubRepoName
+        )
+        .fetch(executor);
+
+        let mut rollups_map: HashMap<PullRequestNumber, HashSet<PullRequestNumber>> =
+            HashMap::new();
+        while let Some(row) = stream.try_next().await? {
+            let rollup_pr = PullRequestNumber(row.rollup_num as u64);
+            let member_pr = PullRequestNumber(row.member_num as u64);
+
+            rollups_map.entry(rollup_pr).or_default().insert(member_pr);
+        }
+
+        Ok(rollups_map)
     })
     .await
 }
--- a/src/github/rollup.rs
+++ b/src/github/rollup.rs
@@ -183,7 +183,6 @@ async fn create_rollup(
         repo_owner,
         mut pr_nums,
     } = rollup_state;
-
     // Repository where we will create the PR
     let base_repo = GithubRepoName::new(&repo_owner, &repo_name);
 
@@ -388,6 +387,17 @@ async fn create_rollup(
         .await
         .context("Cannot create PR")?;
 
+    // Store the rollup PR into the DB
+    let rollup_db = db
+        .upsert_pull_request(&base_repo, pr.clone().into())
+        .await
+        .context("Cannot store the created rollup into the DB")?;
+
+    // And register its rollup member PRs
+    db.register_rollup_members(&rollup_db, &successes)
+        .await
+        .context("Cannot register rollup members")?;
+
     // Set the rollup label
     gh_client
         .add_labels(pr.number, &["rollup".to_string()])
@@ -406,6 +416,7 @@ mod tests {
         User, default_repo_name, run_test,
     };
     use http::StatusCode;
+    use std::collections::{HashMap, HashSet};
 
     #[sqlx::test]
     async fn rollup_missing_fork(pool: sqlx::PgPool) {
@@ -528,6 +539,17 @@ mod tests {
             make_rollup(ctx, &[&pr2, &pr3, &pr4, &pr5])
                 .await?
                 .assert_status(StatusCode::SEE_OTHER);
+            assert_eq!(
+                ctx.db().get_nonclosed_rollups(&default_repo_name()).await?,
+                HashMap::from([(
+                    PullRequestNumber(6),
+                    HashSet::from_iter(vec![
+                        PullRequestNumber(2),
+                        PullRequestNumber(3),
+                        PullRequestNumber(5)
+                    ])
+                )])
+            );
             Ok(())
         })
         .await;
@@ -650,7 +672,57 @@ mod tests {
             insta::assert_snapshot!(pr_url, @"https://github.com/rust-lang/borstest/pull/4");
 
             ctx.pr(4).await.expect_added_labels(&["rollup"]);
+            assert_eq!(
+                ctx.db().get_nonclosed_rollups(pr2.repo()).await?,
+                HashMap::from([(
+                    PullRequestNumber(4),
+                    HashSet::from_iter(vec![PullRequestNumber(3), PullRequestNumber(2)])
+                )])
+            );
 
+            Ok(())
+        })
+        .await;
+        let repo = gh.get_repo(());
+        insta::assert_snapshot!(repo.lock().get_pr(4).description, @"
+        Successful merges:
+
+         - rust-lang/borstest#2 (Title of PR 2)
+         - rust-lang/borstest#3 (Title of PR 3)
+
+        r? @ghost
+
+        <!-- homu-ignore:start -->
+        [Create a similar rollup](https://bors-test.com/queue/borstest?prs=2,3)
+        <!-- homu-ignore:end -->
+        ");
+    }
+
+    /// Ensure that a PR can be associated to multiple rollups.
+    /// This can happen when opening a new rollup similar to an existing one, before closing the old one.
+    #[sqlx::test]
+    async fn multiple_rollups_same_pr(pool: sqlx::PgPool) {
+        let gh = run_test((pool, rollup_state()), async |ctx: &mut BorsTester| {
+            let pr2 = ctx.open_pr((), |_| {}).await?;
+            let pr3 = ctx.open_pr((), |_| {}).await?;
+            ctx.approve(pr2.id()).await?;
+            ctx.approve(pr3.id()).await?;
+
+            make_rollup(ctx, &[&pr2, &pr3]).await?;
+            make_rollup(ctx, &[&pr3]).await?;
+            assert_eq!(
+                ctx.db().get_nonclosed_rollups(pr2.repo()).await?,
+                HashMap::from([
+                    (
+                        PullRequestNumber(4),
+                        HashSet::from_iter(vec![PullRequestNumber(2), PullRequestNumber(3)])
+                    ),
+                    (
+                        PullRequestNumber(5),
+                        HashSet::from_iter(vec![PullRequestNumber(3)])
+                    )
+                ])
+            );
             Ok(())
         })
         .await;
--- a/src/templates.rs
+++ b/src/templates.rs
@@ -3,9 +3,12 @@ use crate::database::{
     BuildModel, BuildStatus, MergeableState::*, PullRequestModel, QueueStatus, TreeState,
     WorkflowModel,
 };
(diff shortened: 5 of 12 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+/.sqlx/*.json binary linguist-generated=true
```
Kobzol, line 1: Did you have any specific use-case for this? GitHub already marked them as generated before.

src/database/client.rs
```diff
@@ -344,4 +347,39 @@ impl PgDbClient {
     pub async fn delete_tagged_bot_comment(&self, comment: &CommentModel) -> anyhow::Result<()> {
         delete_tagged_bot_comment(&self.pool, comment.id).await
     }
+
+    /// Register the contents of a rollup in the DB.
+    /// The contents are stored as rollup PR number - member PR number records, scoped under a specific repository.
+    /// All records are inserted in a single transaction.
+    pub async fn create_rollup(
```
```suggestion
    pub async fn register_rollup_members(
```
I think that's a bit less misleading.

migrations/20251221160635_rollup_contents.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS rollup_contents
```
Kobzol: I like the member terminology that you have used, so let's go with it.
```suggestion
CREATE TABLE IF NOT EXISTS rollup_member
```

src/database/client.rs
```diff
@@ -344,4 +347,39 @@ impl PgDbClient {
...
+        let mut tx = self.pool.begin().await?;
+        for member_pr_number in member_pr_numbers {
+            create_rollup_pr_content(&mut *tx, repo, rollup_pr_number, member_pr_number).await?;
+        }
+        tx.commit().await?;
+        Ok(())
+    }
+
+    /// Returns the numbers of all the PRs associated with the requested rollup.
+    pub async fn get_rollup_pr_contents(
```
Kobzol: I think that the main operation that we will want to do is load members information for a set of PRs that will be displayed on the queue page. So I'm thinking something like:
1) Load non-closed PRs on the queue page endpoint
2) Do a query that will fetch all members for all rollups in those PRs.

And then in the frontend, we'll build a reverse index, so that we can also associate a PR with its rollup(s).

As an alternative to 1) + 2), we could directly attach a bool whether a given PR is a rollup or not to the return value of 1), and then only ask for members from actual rollups, to reduce the DB traffic slightly. But we'd need to take a look at the query plans to see if it makes sense. And probably perf. will be fine with both approaches.

migrations/20251221160635_rollup_contents.up.sql
```diff
@@ -0,0 +1,11 @@
...
+(
+    -- We have to store the PR through their numbers rather than a FK to the pull_request table. This is due to the rollup PR
+    -- being created at the time of record insertion in this table, so we won't have an entry in pull_request for it
+    rollup_pr_number  BIGINT NOT NULL,
+    member_pr_number  BIGINT NOT NULL,
+    repository TEXT   NOT NULL,
+
+    PRIMARY KEY (repository, rollup_pr_number, member_pr_number)
+);
+CREATE INDEX rollup_contents_member_pr_number_idx ON rollup_contents (repository, member_pr_number);
```
Kobzol: As per my other comment, I think that it makes more sense to create the index on the `rollup_pr_number` attribute, as we will be loading members of rollups, and not looking up rollups for individual PRs, at least for the queue page.

.gitattributes
Carbonhell: Sorry, yes, it's mostly for the `binary` part, I have a windows workstation and the sqlx command to generate those files was causing diffs due to line endings. I copied this file from the docs.rs repo to avoid this issue. I can remove it from the PR as it's out of scope, if you prefer.

Kobzol: Oh, I see. Let's keep it then, no real harm in that.

migrations/20251221160635_rollup_contents.up.sql
Carbonhell: I will proceed removing this index, then. For the index on the `rollup_pr_attribute` I assume you mean including the `repository` attribute as well, like the query in your review. That should already be covered by the index generated by the primary key, as I've purposefully ordered the columns by putting the `member_pr_number` attribute last, as per https://www.postgresql.org/docs/current/indexes-multicolumn.html .

Kobzol: Yeah, also including repository. Fair enough, as long as the index is used for the query I posted, it should be fine.

I ran some simulations with 20k PRs and 3k rollup_member rows on a local Postgres instance, and `EXPLAIN ANALYZE` was still choosing a sequential scan over rollup members anyway. So maybe the index won't be even needed so much. But that shouldn't be an issue for some time anyway.

Carbonhell: Disclaimer, I don't consider myself an expert in the optimization of Postgres query plans, but I think it's mostly a matter of the planner not considering worth using the index with such a low amount of data.

For reference, with 118925 rollup member records all specific to only a single repository, of which only 3617 are related to an open/draft rollup (100 rollups) and with rollups averaging on 10 members each, the query I wrote (slightly tweaked to use a join instead of a subquery) uses the index:

```
Nested Loop  (cost=0.71..509.37 rows=214 width=16) (actual time=0.068..0.633 rows=3617 loops=1)
  ->  Index Scan using pull_request_status_idx on pull_request pr  (cost=0.29..23.80 rows=90 width=13) (actual time=0.040..0.056 rows=100 loops=1)
"        Index Cond: (status = ANY ('{open,draft}'::text[]))"
        Filter: (repository = 'repo'::text)
  ->  Index Only Scan using rollup_member_pkey on rollup_member rm  (cost=0.42..5.00 rows=40 width=21) (actual time=0.003..0.004 rows=36 loops=100)
        Index Cond: ((repository = 'repo'::text) AND (rollup_pr_number = pr.number))
        Heap Fetches: 0
Planning Time: 0.533 ms
Execution Time: 0.708 ms
```
Which I was able to reproduce by truncating the table and regenerating the test data, so it should be a realistic assumption.

src/database/client.rs
Carbonhell: Judging by the query plan I shared in the other thread, I don't think it's needed. For now, the PR implements an operation to perform the 2nd query in a standalone way, which is expected to be used in the queue endpoint handler (is there an issue for that already? I can look into that after this PR gets merged)

Carbonhell:
Sorry for the long delay (and the force-pushes :eyes:), the PR should be ready for another review now. Thank you in advance :)

src/github/rollup.rs
```diff
@@ -672,12 +733,16 @@ mod tests {
...
-        .await
+        let response = ctx
+            .api_request(rollup_request(
+                &rollup_user().name,
+                default_repo_name(),
+                &prs,
+            ))
+            .await;
+        // Trigger a refresh so that the new rollup PR can be upserted in the pull_request table
+        ctx.refresh_prs().await;
```
Kobzol: This shows that our test suite currently lacks "webhook feedback", where changing something in the GitHub test model should generate new webhooks into bors automatically :) Good catch.

src/github/rollup.rs
```diff
@@ -598,6 +620,45 @@ mod tests {
...
+            let pr3 = ctx.open_pr((), |_| {}).await?;
+            ctx.approve(pr2.id()).await?;
+            ctx.approve(pr3.id()).await?;
+
+            make_rollup(ctx, &[&pr2, &pr3]).await?;
+            // Simulate a maintainer closing the rollup due to a CI failure requiring popping a PR from the contents
+            ctx.set_pr_status_closed(4).await?;
+            make_rollup(ctx, &[&pr3]).await?;
+            // Ensure both rollups have the requested PRs
+            assert_eq!(
```
Kobzol, line 713: Shouldn't this contain two rollup PRs, rather than just one? 🤔

Kobzol (commented):
Thank you, it looks good!

migrations/20251221160635_rollup_contents.up.sql
Kobzol: I liked the subquery more, as it gave me more confidence that the WHERE filter will be pushed down to the PR table :laughing: But it seems that Postgres is smart enough to pre-filter the open/draft PRs during the JOIN, and not only after. Looks good, thanks for the experiment.

Kobzol:
I wouldn't actually mind having some basic usage of the new queries in the same PR - it often turns out that a new API will need to be changed once it's actually used :laughing: Would you like to take a stab at that?

Some ideas:
1. If a rollup is currently being tested (it has state *pending*), it should not be possible to add its members to a new rollup (the rollup checkboxes should be disabled for the members)
2. Each rollup in the queue could have some color associated to it, and we could show some small rectangle or circle before/after the rollup PR number or title, in the queue table. And then we could also add this rectangle/circle at the same spot for the rollup members, to make it visually obvious which members belong to which rollups.
3. When you hover on a rollup, its members could be highlighted.

I think that 3. might be the simplest to implement, and it would be enough to test that the queries that we have here make sense. The rest could be done in follow-up PRs.

src/github/rollup.rs
Carbonhell: No, as the first rollup is closed (it corresponds to PR 4, closed in line 634 above)

Kobzol: Right, but I thought that the test should test that we can have a single PR attached to two (or more) rollups, but here the `get_nonclosed_rollups` function returns only a single rollup/PR mapping anyway.

Carbonhell: 🤦 sorry, you're right.
The test initially used the queries that I implemented before your review. When refactoring the code to use the new query, I left this single assert to verify that the following common scenario works:
- a rollup is opened with PRs 2 and 3
- the tests fail due to PR 2
- the rollup is closed, and a similar one is opened with only PR 3 as member
- the resulting rollup membership is registered correctly, regardless of the previous rollup.

The comment above this line still refers to the old asserts where I was checking both rollups. That said, reading the test again I think it's unnecessarily contrived - I'll tweak it to remove the rollup closing procedure and to check if both rollups are returned with the expected member PRs. This should also be more similar to what happens in practice, which based on what you said, new rollups may be created before the old one gets closed. Thank you for the feedback!

src/templates.rs
```diff
@@ -55,6 +58,17 @@ pub struct PullRequestStats {
     pub failed_count: usize,
 }
 
+/// A displayable view of a map of rollups to their members, meant to be used with data attributes
+pub struct RollupsInfo {
+    pub rollups: HashMap<PullRequestNumber, HashSet<PullRequestNumber>>,
```
Carbonhell, line 63: FYI: I've added the `RollupsInfo` struct as I was using it in my tests with the icon per rollup member to store a reverse hashmap (rollup PR member -> rollups containing it) to use within the template. I plan to introduce this reverse hashmap in my next PR. Not sure if there are easier ways with Askama - I'm not familiar with it, I assume it's better to prepare the data as needed on the Rust side, and then embed it without performing complex operations in the template itself to keep it simple

Carbonhell:
As agreed on Zulip, for now I've only implemented the hover functionality, as seen in the following screenshot (mouse hovering on a rollup not shown):
image

I've also fixed the issue with the multiple_rollups_same_pr test from the last review.
The PR should be ready for review. Thanks in advance :)

migrations/20260106212918_rollup_member.up.sql
```diff
@@ -0,0 +1,11 @@
+-- Add up migration script here
+CREATE TABLE IF NOT EXISTS rollup_member
+(
+    -- We have to store the PR through their numbers rather than a FK to the pull_request table. This is due to the rollup PR
+    -- being created at the time of record insertion in this table, so we won't have an entry in pull_request for it
+    rollup_pr_number  BIGINT NOT NULL,
+    member_pr_number  BIGINT NOT NULL,
+    repository TEXT   NOT NULL,
```
Kobzol:
```suggestion
    repository TEXT   NOT NULL,
    rollup_pr_number  BIGINT NOT NULL,
    member_pr_number  BIGINT NOT NULL,
```
Just a nit, it's easier to read like this when inspecting the DB state manually :) The migration data SQL file will also need to be updated after this change.

src/database/client.rs
```diff
@@ -336,4 +339,29 @@ impl PgDbClient {
...
+    ) -> anyhow::Result<()> {
+        let mut tx = self.pool.begin().await?;
+        for member_pr_number in member_pr_numbers {
+            register_rollup_pr_member(&mut *tx, repo, rollup_pr_number, member_pr_number).await?;
+        }
+        tx.commit().await?;
+        Ok(())
+    }
+
+    /// Returns a map of rollup PR numbers to the set of member PR numbers that are part of that rollup.
```
Kobzol, line 396:
```suggestion
    /// Returns a map of rollup PR numbers to the set of member PR numbers that are part of that rollup.
    /// Only returns non-closed rollup PRs.
```

src/database/operations.rs
```diff
@@ -1064,3 +1065,60 @@ pub(crate) async fn clear_auto_build(
...
+        let mut stream = sqlx::query!(
+            r#"
+        SELECT rm.rollup_pr_number, rm.member_pr_number
+        FROM rollup_member rm
+        JOIN pull_request pr
+          ON pr.repository = rm.repository
+         AND pr.number = rm.rollup_pr_number
+        WHERE rm.repository = $1
+          AND pr.status IN ('open', 'draft')
+        ORDER BY rm.rollup_pr_number;
```
Kobzol:
```suggestion
```
Doesn't seem to be required, since we store the numbers in a hashmap/hashset anyway? Might help with cache locality for the hashmap, but that's not really needed here, and the DB won't have to sort the rollups.

src/templates.rs
Kobzol: Yeah, preparing complicated state on the Rust side is definitely a good idea!

web/templates/queue.html
```diff
@@ -450,8 +464,43 @@ <h1>
...
+            }
+            rollupPrMembers = rollupPrMembers.split(",");
+            for (const rollupPrMember of rollupPrMembers) {
+                tbody.querySelector(`tr[data-pr-number="${rollupPrMember}"]`).classList.add("rollup-hover");
+            }
+        }
+        tbody.addEventListener("mouseover", handler);
+        tbody.addEventListener("mouseleave", clear);
+
+        return () => {
```
Kobzol, line 520: It's nice that you implemented the event removal, but we don't call it anywhere anyway, so 🤷 I'd just remove it to shorten the code.

web/templates/queue.html
```diff
@@ -450,8 +464,43 @@ <h1>
...
+        const tbody = document.querySelector("#table tbody");
+        if (!tbody) {
+            return () => {};
+        }
+        const clear = () => {
+            tbody.querySelectorAll(".rollup-hover").forEach(el => el.classList.remove("rollup-hover"));
+        };
+        const handler = (event) => {
+            // Reset highlight
+            clear()
```
Kobzol:
```suggestion
            clear();
```

web/templates/queue.html
```diff
@@ -450,8 +464,43 @@ <h1>
...
+            const rowElement = event.target.closest("tr");
+            if (!rowElement) {
+                return;
+            }
+            let rollupPrMembers = rowElement.dataset.rollupPrMembers;
+            if (!rollupPrMembers) {
+                return
```
Kobzol:
```suggestion
                return;
```

web/templates/queue.html
Carbonhell: Will do. One related note: do you think it might be worth to support hovering on mobile devices too? I was thinking of hooking the hover handle to the touchstart event (and clear on touchend).

Kobzol: Hmm, not sure, because on click we select the PR - how do you actually distinguish hover and click on mobile? :thinking: If it can be done reliably, then might be useful, but I'd leave it for a future PR :)

src/github/rollup.rs
```diff
@@ -388,6 +387,14 @@ async fn create_rollup(
         .await
         .context("Cannot create PR")?;
 
+    db.register_rollup_members(
```
Kobzol: I was thinking about (many!) follow-up features that we could implement once this PR lands, and I realized that the problem with the missing PR id can be solved very easily. We can simply pre-insert the PR into the DB immediately here:
```rust
let pr_db = db.upsert_pull_request(..., pr.into());
```
and then use `pr_db.id` to record the rollup mapping in the DB.

Since we anyway have to do a `JOIN` with the PR table when selecting the rollup mapping due to filtering out open PRs, I think that storing the IDs in the mapping would be simpler, after all, would waste less DB space, and would provide proper FK mapping in the DB.

Could you please change it to store `(FK to PR table, rollup PR number)` in the `rollup_member` table? Sorry for the late notice 😅

Carbonhell: That makes sense! Some questions:
1. I assume the fact we perform another upsert from the webhook event received by GitHub for the same PR doesn't matter, as it'd just write the same data. Adding a check to prevent writing PRs that are already in the table in the webhook logic seems pointless - is my assumption correct?
2. Wouldn't it make sense to change the member PR number in the same way? I assume we're guaranteed to have an entry in the PR table, as from what I know, rollups can only be created from the queue page, which implies PRs shown there have an entry in the table. Therefore we could use a FK for member PRs too for the same reasons. I think the only tradeoff is that writes would be slightly more expensive, but it shoudn't be a concern with the low amount of writes, and it might help future with future query patterns, if any.

I'm always looking for issues to contribute to (mostly during weekends, when i'm not too tired from working in a startup haha). If you want, write down the ideas in GitHub issues and I'll take care of them!

Kobzol: 1. Yeah, the open PR webhook is not load-bearing. It can in theory race with the rollup handler, but that shouldn't matter much, Postgres will ensure atomicity of the PR creation.
2. A very good point! Yeah, I agree. In all cases where we will query this table, the corresponding PRs should already be in the DB. Including when we record the member entries.

Btw, after this change you can remove `refresh_prs` from rollup tests.

Okay, I will create issues for the features, once this PR lands.

PR #547, opened 2026-01-06
fix(ci): fail if deployment goes wrong

marcoieni:
In https://docs.aws.amazon.com/cli/latest/reference/ecs/wait/services-stable.html I didn't find a way to override the default values (15 seconds and 40 attempts).


marcoieni:
This is what many people suggest online, so I would say let's test it!

Kobzol:
Ok :) Let's try.

PR #552, opened 2026-01-06
Add CI cache

Kobzol:
Hopefully this will make PR and merge queue builds faster.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -4,15 +4,28 @@
     "config:recommended",
     ":maintainLockFilesMonthly",
     "docker:disable",
+    "helpers:pinGitHubActionDigests"
   ],
   packageRules: [
     {
-      matchCategories: ["rust"],
-      updateTypes: ["patch"],
+      matchCategories: [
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -24,6 +28,9 @@ jobs:
           toolchain: 1.88.0
           components: clippy, rustfmt
 
+      - name: Rust cache
+        uses: Swatinem/rust-cache@v2.8.2
```
marcoieni: a better security practice is to pin the hash of the commit, as the tags can be changed.

marcoieni: we should have renovate to update it! :)

Kobzol: Pinned. We have renovatebot, not sure if it handles GHA actions though.

marcoieni (approved):
check https://github.com/rust-lang/crates-io-auth-action/commit/4c7b24b740172de793d7d4a68571a6b14caef23c to see how renovate expects the version

PR #577, opened 2026-01-07
Add `BuildCompleted` homu-compatible comment metadata to test successful comments

Kobzol:
It is currently required by triagebot, although I would like to get rid of it eventually.


Urgau (approved):
Looks good to me compared to what triagebot [is expecting](https://github.com/rust-lang/triagebot/blob/c19f6d743e6a5c8a413d0ecd0a6f3d608a029c94/src/handlers/rustc_commits.rs#L42-L63).

PR #594, opened 2026-01-09
Implement a `@bors cancel` command

Kobzol:
This command allows cancelling a currently running auto build on a given PR, thus "yielding" its slot in the queue. The usual motivation is to start testing a different PR, which has gotten to the front of the queue in the meantime.


src/bors/command/mod.rs
```diff
@@ -130,6 +130,8 @@ pub enum BorsCommand {
     /// Set the tree closed with a priority level.
     TreeClosed(Priority),
     /// Clear a failed auto build status from an approved PR.
-    /// This will cause the merge queue to attempt to start a new auto build and retry merging the PR again.
+    /// This will cause the merge queue to eventually attempt to try to test the PR again.
     Retry,
+    /// Cancel an auto build currently running on a given PR.
```
Urgau:
```suggestion
    /// Cancel an auto build currently running on a given PR (without removing it from the queue).
```

Issue #597, opened 2026-01-09
Visited links don't have a different color


bjorn3:
The color of links is overridden, but there a rule for `a:visited` is missing, so visited links look the same as non-visited links.

https://github.com/rust-lang/bors/blob/de760673076b0128590f1147ae3abfa89d0ac853/web/templates/base.html#L118-L136

PR #603, opened 2026-01-10
#597 different color

mohit2162004:
Added a simple css code in the html file . now user able see both the diffrence 

Fixes: https://github.com/rust-lang/bors/issues/597


Kobzol (approved):
Thank you! I'm not sure if the contract is different enough, but we'll see what others think.

PR #637, opened 2026-01-13
Block PR approval when merge conflicts are present

### LINKED ISSUE #452: Do not allow approving PRs with a merge conflict

enthropy7:
added new func + test

fixes #452

Prevents `r+` command from approving PRs with merge conflicts by adding conflict checks in `check_pr_approval_validity()`. The implementation uses `MergeableState::from()` to convert GitHub mergeable state and checks both GitHub and database states to catch conflicts even when GitHub state hasn't been refreshed yet. Returns `approve_merge_conflict_comment()` error message instead of approving the PR.

added tests for all conflict scenarios: dirty, blocked, unknown state with DB conflicts, and clean state with DB conflicts.

### DIFF
--- a/src/bors/handlers/review.rs
+++ b/src/bors/handlers/review.rs
@@ -463,7 +463,7 @@ mod tests {
 
     use crate::bors::TRY_BRANCH_NAME;
     use crate::bors::merge_queue::AUTO_BUILD_CHECK_RUN_NAME;
-    use crate::database::{DelegatedPermission, MergeableState, OctocrabMergeableState, TreeState};
+    use crate::database::{DelegatedPermission, OctocrabMergeableState, TreeState};
     use crate::tests::BorsTester;
     use crate::tests::default_repo_name;
     use crate::{


src/bors/handlers/mod.rs
```diff
@@ -367,6 +367,14 @@ async fn handle_comment(
     )
     .await?;
 
+    // Reload PR from DB after mergeability update, to ensure we have the latest state
```
Kobzol: In theory it might be useful if our DB functions updated the in-memory models, but that's a larger refactoring, and I'm not sure if it's a good idea altogether.

This would add another DB call into a hot-path (handling a comment), which shouldn't really be needed. `set_pr_mergeability_based_on_user_action` could just return an updated DB model to fix this, but I think that even that is not needed - see my other comment :)

Kobzol: In theory it might be useful if our DB functions updated the in-memory DB models, but that's a larger refactoring, and I'm not sure if it's a good idea altogether.

This would add another DB call into a hot-path (handling a comment), which shouldn't really be needed - see my other comment.

src/bors/handlers/review.rs
```diff
@@ -124,6 +125,15 @@ async fn check_pr_approval_validity(
         return Ok(Some(approve_blocking_labels_present(&blocking_labels)));
     }
 
+    let mergeable_state = MergeableState::from(pr.github.mergeable_state.clone());
```
Kobzol, line 128: What we get from GitHub is always the latest state. If we know it, there's not much point in checking the DB state. So here let's check only the GH state - if it has conflicts, we report a conflict. If it is OK, we should treat it as being OK, even though in our DB it's still marked as having conflicts. And if it's unknown, it will have to be reloaded, and there's no point in waiting for that here, so we should just allow it.

src/bors/mergeability_queue.rs
```diff
@@ -700,11 +700,16 @@ pub async fn set_pr_mergeability_based_on_user_action(
...
     if mergeable_state == MergeableState::Unknown {
-        // Eagerly update the last known state in the DB. This will essentially avoid posting the
-        // conflict notification, because the last state will be unknown.
-        db.set_pr_mergeable_state(&db_pr.repository, db_pr.number, mergeable_state)
-            .instrument(span)
-            .await?;
+        // Only update to Unknown if the DB doesn't already have a known conflict state.
+        // We should not overwrite HasConflicts with Unknown, as that would lose important
+        // information about merge conflicts that should block approval.
+        if db_pr.mergeable_status() != MergeableState::HasConflicts {
```
Kobzol: It is actually load bearing that we always update the mergeability here 😅 Definitely shouldn't be changed because of this PR's needs! :) When the mergeability changes based on user action, we want to avoid sending mergeability notifications, and if we added the `if` here, that might happen.

src/bors/mergeability_queue.rs
```diff
@@ -717,10 +722,14 @@ pub async fn set_pr_mergeability_based_on_user_action(
 
         // If the status is unknown, we assume that the PR will be refreshed soon by some other
         // means, and we do not do anything here
-        if db_pr.mergeable_status() != MergeableState::Unknown {
-            // If it is not unknown, we assume that for some reason this PR has not been updated
-            // properly in the DB. We will thus force update the state here, without sending a
-            // notification, and without unapproving the PR.
+        // Also, do not overwrite HasConflicts with any other state, as that would lose critical
+        // merge conflict information needed to block PR approval.
+        if db_pr.mergeable_status() != MergeableState::Unknown
```
Kobzol: Same here, please do not modify this function :)

Kobzol (commented):
Thank you, looks great! Mergeability checking is very tricky, and I just finished a big reworking of it, so some of the invariants might not be documented properly yet, but this PR shouldn't need to change how it works - it should be enough to simply block approvals if the PR is currently unmergeable on GitHub.

src/bors/mergeability_queue.rs
enthropy7: thanks for all comments, i think a might overthink a problem a little bit xD. everything should be fine now

src/bors/handlers/review.rs
```diff
@@ -1563,4 +1569,106 @@ labels_blocking_approval = ["proposed-final-comment-period", "final-comment-peri
         })
         .await;
     }
+
+    #[sqlx::test]
+    async fn approve_pr_with_merge_conflict_dirty(pool: sqlx::PgPool) {
```
Kobzol, line 1574: Tbh I think that this test is enough. I would like to avoid using any functions that modify the DB in tests, as that "cheats"; I want to only use end-to-end/integration logic in tests if possible, to simulate realistic-ish scenarios.

The test for the `Blocked` state is IMO also not needed; if we want to test the mapping between GH/DB mergeable tests, we should have specific tests for that, rather than piggybacking on this functionality (I don't think we need those tests at the moment though).

Kobzol (approved):
Thank you! :)

PR #645, opened 2026-01-16
Implement `@bors squash` (take two)

Kobzol:
Second attempt, this time using `libgit2` to perform the push of the squashed commit to a fork, and with tests.

I hid the `libgit2` dependency behind a feature, to make it easier for contributors who can't build that dependency (although it is enabled by default).


src/bors/handlers/squash.rs
```diff
@@ -0,0 +1,387 @@
...
+            .author
+            .email
+            .clone()
+            .unwrap_or_else(|| bors_commit_author().email),
+    });
+
+    // Create the squashed commit on the source repository.
+    // We take the parents of the first commit, and the tree of the last commit, to create the
+    // squashed commit.
+    let commit_msg = &pr.github.title;
```
bjorn3, line 98: This would lose the original commit messages. I don't know of an easy heuristic for that though. Another option may be to pick the PR description and remove the bors ignored part, all closes commands and the r?. Or maybe allow giving the commit message as argument to `@bors squash`?

Kobzol: I implemented this as an MVP, mostly to check permissions. There is a pretty big design aspect left, and that is how to generate the commit message, exactly as you say :laughing: Let's discuss that on Zulip.

src/bors/handlers/squash.rs
```diff
@@ -0,0 +1,387 @@
...
+
+    let mut source_remote = repo.remote_anonymous(&source_repo_url)?;
+    let mut fetch_options = FetchOptions::new();
+    fetch_options.remote_callbacks(callbacks());
+
+    tracing::debug!("Fetching commit");
+    // Fetch the commit from the source repo
+    source_remote
+        .fetch(&[commit_sha.as_ref()], Some(&mut fetch_options), None)
+        .with_context(|| anyhow::anyhow!("Cannot fethc commit {commit_sha} from {source_repo}"))?;
```
bjorn3, line 218: Wouldn't this clone the entire repo? Can you use `fetch_options.depth(1)` to at least skip all history? Looks like libgit2 doesn't support partial clones though, so you can't do the equivalent of `--filter=tree:0` to avoid downloading any blobs and trees at all.

Kobzol: ...crap, I thought it was too easy. Yeah, this will have to be handled somehow (possibly with a disk cache :( ).

Issue #606, opened 2026-01-10
feature: write run duration into `auto_build_succeeded_comment`


matthiaskrgr:
something like  :sunny: `Test successful after 3hours 21 minutes`  would useful from time to time.

Kobzol:
https://github.com/rust-lang/bors/pull/487 has been merged, so this is now unblocked :)

PR #650, opened 2026-01-18
Include build duration in success comment

reddevilmidzy:
close: #606

### DIFF
--- a/src/bors/comment.rs
+++ b/src/bors/comment.rs
@@ -409,6 +409,7 @@ pub fn auto_build_succeeded_comment(
     approved_by: &str,
     merge_sha: &CommitSha,
     base_ref: &str,
+    duration: Option<Duration>,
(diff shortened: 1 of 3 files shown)


src/bors/comment.rs
```diff
@@ -417,10 +418,17 @@ pub fn auto_build_succeeded_comment(
         .collect::<Vec<_>>()
         .join(", ");
 
+    let duration = if let Some(dur) = duration {
+        format_duration(dur)
+    } else {
+        "Unknown duration".to_string()
+    };
```
reddevilmidzy, line 425: How should we handle it when the duration is None? (Even though it seems unlikely to happen🙄)

Kobzol: I'd avoid including the `Duration` row in that case, but this is also fine.

Issue #430, opened 2025-09-13
Escaping job regex via backtick does not work


RalfJung:
I issued the command
```
@bors try jobs=`dist-various-*`
```
which did not work:
> Unknown value for argument "jobs". Run @bors2 help to see available commands.

That error is not super helpful, I know about the command, I just don't know how to escape `*`... and the help text does not indicate that it seems.

I expected this to work similar to "try-jobs=`regex`" in the PR description.

Kobzol:
I wouldn't even say that it's real syntax, it's mostly just a hack to prevent GitHub from rendering the command as Markdown. Writing `bors try jobs=foo-*` can be interpreted by bors perfectly fine, it's just that GH might render it in a weird way.

I'll take a look at improving the error message.

RalfJung:
> I wouldn't even say that it's real syntax, it's mostly just a hack to prevent GitHub from rendering the command as Markdown. Writing bors try jobs=foo-* can be interpreted by bors perfectly fine, it's just that GH might render it in a weird way.

Yeah I understand, but it does look odd and makes it harder for a human to tell (from looking at the comment) what jobs were run. So if `\*` works that seems preferable.

PR #652, opened 2026-01-19
Allow specifying a custom commit message for `@bors squash`

Kobzol:
By defaut, it will now include the messages of all squashed commits. The commit message can now be overwritten using `@bors squash msg="Foo bar"`.

### DIFF
--- a/src/bors/handlers/squash.rs
+++ b/src/bors/handlers/squash.rs
@@ -509,7 +509,11 @@ mod tests {
             ctx.approve(()).await?;
             ctx.post_comment("@bors squash").await?;
             ctx.run_gitop_queue().await?;
-            ctx.expect_comments((), 1).await;
+            insta::assert_snapshot!(ctx.get_next_comment_text(()).await?, @"
+            :hammer: 2 commits were squashed into sha2-reauthored-to-git-user.
+
+            The pull request was unapproved.
+            ");
             ctx.pr(()).await.expect_unapproved();
 
             Ok(())


src/bors/command/parser.rs
```diff
@@ -165,15 +175,75 @@ fn parse_command(input: &str, parsers: &[ParserFn]) -> ParseResult {
...
+        let mut in_quotes = false;
+        let mut quote_start = None;
+        let mut quote_end = None;
+
+        // Parse a `bare` part or a `key=value` part
+        while i < len {
+            let b = bytes[i];
+
+            if in_quotes {
+                if b == b'"' {
```
programmerjake, line 207: should this support escaping quotes?

Kobzol: Could be done as a follow-up, but I'd wait for a use-case :laughing: It also can't deal with multi-line commit messages, backticks, etc., so it is limited anyway.

programmerjake: well, you're depending on `serde_json` anyway, why not use that to parse the string? something like:
```rust
// Check if value starts with quote
if i < len && bytes[i] == b'"' {
    let mut de = serde_json::de::Deserializer::from_slice(&bytes[i..]).into_iter();
    match String::deserialize(&mut de) {
        Ok(value) => {
            i += de.byte_offset();
            // handle string
        }
        Err(e) => {
            // handle error
        }
    }
}
```

Kobzol: Good idea! I tried it, but sadly it looks like `pulldown_cmark` removes the quote escapes :( The interaction between our parser and Markdown is a bit complicated (https://github.com/rust-lang/bors/issues/430#issuecomment-3290866324).

PR #675, opened 2026-01-26
Warn about unknown reviewers when approving

reddevilmidzy:
part of #191

Check for invalid reviewers and include a warning in comments.

### DIFF
--- a/src/bors/handlers/review.rs
+++ b/src/bors/handlers/review.rs
@@ -518,6 +518,7 @@ mod tests {
     use crate::bors::TRY_BRANCH_NAME;
     use crate::bors::merge_queue::AUTO_BUILD_CHECK_RUN_NAME;
     use crate::database::{DelegatedPermission, OctocrabMergeableState, TreeState};
+    use crate::permissions::PermissionType;
     use crate::tests::default_repo_name;
     use crate::tests::{BorsTester, Commit};
     use crate::{
@@ -565,8 +566,16 @@ approved = ["+approved"]
 
     #[sqlx::test]
     async fn approve_on_behalf(pool: sqlx::PgPool) {
-        run_test(pool, async |ctx: &mut BorsTester| {
-            let approve_user = "user1";
+        let approve_user_id = 200;
+        let approve_user = "user1";
+        let mut gh = GitHub::default();
+        gh.add_user(User::new(approve_user_id, approve_user));
+        gh.default_repo().lock().permissions.users.insert(
+            User::new(approve_user_id, approve_user),
+            vec![PermissionType::Review],
+        );
+
+        run_test((pool, gh), async |ctx: &mut BorsTester| {
             ctx.post_comment(format!(r#"@bors r={approve_user}"#).as_str())
                 .await?;
             insta::assert_snapshot!(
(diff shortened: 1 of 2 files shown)


src/bors/handlers/review.rs
reddevilmidzy, line 644: @Kobzol I have a question!
In this case, would it be better to include the same warning, or would it be better to mock it and test it as if the usernames foo, ... exist?

Kobzol: It would be slighty better to do the latter, because this test should be about normalizing usernames, not about checking if they exist.

src/bors/handlers/review.rs
```diff
@@ -97,6 +111,40 @@ fn normalize_approvers(approvers: &str) -> String {
         .join(",")
 }
 
+/// Check if the specified reviewers exist as GitHub users or teams.
+/// Returns comma-separated string of unknown reviewer names, or None if all exist.
+async fn check_unknown_reviewers(repo_state: &RepositoryState, reviewers: &str) -> Option<String> {
+    let mut unknown_reviewers = Vec::new();
+
+    for reviewer in reviewers.split(',') {
```
Kobzol: To avoid re-parsing them again, we could make `normalize_approvers` return `Vec<String>`, and then make this function take a `Vec`, and return `Vec<String>` (of unknown reviewers). And then we can join the normalized vec back to a string in `(normalized, unknown)`.

src/bors/handlers/review.rs
```diff
@@ -97,6 +111,40 @@ fn normalize_approvers(approvers: &str) -> String {
...
+        let user_exists = repo_state
```
Kobzol: I think we can just check through the team API that we already have loaded in permissions, to avoid asking the GitHub API. We can simply add a new function to the permissions checker that asks whether a given username or a team exists or not.

src/bors/handlers/review.rs
reddevilmidzy: I've applied your suggestion about method signature. 
For `unknown`, it has the type `Option<String>`, but did you mean to change it to `String` here as well?

src/bors/handlers/review.rs
reddevilmidzy: Shouldn't this API(https://team-api.infra.rust-lang.org/bors.review.json) be used to retrieve the team from this code? 
https://github.com/rust-lang/bors/blob/3c11e943a23a1b72abc965fee53699aaa7ac0744/src/permissions.rs#L121-L123

I haven't checked this API, so I'm not sure if it's implemented correctly.

src/bors/handlers/review.rs
Kobzol: I meant to turn `Option<String>` into `Vec<String>`. Really it's `Option<Vec<String>>`, but the option is redundant, since the vec can just be empty.

src/bors/handlers/review.rs
Kobzol: Hmm, I didn't express myself very well :laughing: Originally, I wanted to check usernames from https://team-api.infra.rust-lang.org/v1/people.json and team names from https://team-api.infra.rust-lang.org/v1/teams.json. That being said, we would have to load these two new resources, and maybe change the Permissions stuff to be more like TeamApi stuff, because it's no longer just about permissions anymore.

That is a larger change than what I intended for this PR. I think that as a first step, checking that the reviewers have `r+` access is good enough. So please just change it to only check usernames of actual reviewers (`review` permission), not the people with only try access (`try` permission).

src/bors/handlers/review.rs
reddevilmidzy: Ah, I understand. I changed `Option<String>` to `Vec<String>`!

src/bors/handlers/review.rs
reddevilmidzy: Thanks for the clarification! Changed to only check review permission usernames. 
I'll tackle the people.json/teams.json approach in a follow-up PR 👍

PR #678, opened 2026-01-28
Fix deploy workflows

Kobzol:


marcoieni:
in this way you checkout twice:
https://github.com/rust-lang/bors/blob/794384c7e8ef20973b4205272125418cf246a9c5/.github/actions/deploy-ecs/action.yml#L12

you probably should remove the checkout from the shared action.

PR #682, opened 2026-01-28
Add TeamDirectory and load people/teams names

reddevilmidzy:
Follow up to: https://github.com/rust-lang/bors/issues/675

The logic that previously checked with r+ permission has been changed to a method that checks by retrieving people and team names from the team API.

Fixes: https://github.com/rust-lang/bors/issues/191

### DIFF
--- a/src/bors/handlers/review.rs
+++ b/src/bors/handlers/review.rs
@@ -783,6 +783,49 @@ approved = ["+approved"]
     }
 
     #[sqlx::test]
+    async fn approve_with_unknown_team(pool: sqlx::PgPool) {
+        run_test(pool, async |ctx: &mut BorsTester| {
+            ctx.post_comment("@bors r=nonexistent-team").await?;
+            insta::assert_snapshot!(
+                ctx.get_next_comment_text(()).await?,
+                @r###"
+            :pushpin: Commit pr-1-sha has been approved by `nonexistent-team`
+
+            It is now in the [queue](https://bors-test.com/queue/borstest) for this repository.
+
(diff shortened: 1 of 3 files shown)


src/permissions.rs
```diff
@@ -51,19 +81,24 @@ impl UserPermissions {
         }
     }
 
-    /// Checks if the given username is a valid reviewer.
-    /// The usernames are checked in a case-insensitive manner.
-    pub fn has_reviewer(&self, username: &str) -> bool {
-        self.review_usernames.contains(&username.to_lowercase())
+    pub fn user_exists(&self, username: &str) -> bool {
+        self.directory.user_exists(&username.to_lowercase())
```
Kobzol: This is a bit inconsistent, as the `TeamDirectory` lowercases internally, but then checks usernames without lowercasing, and the lowercasing happens here. I'd move it to `TeamDirectory` directly.

src/permissions.rs
```diff
@@ -90,20 +125,24 @@ impl TeamApiClient {
     ) -> anyhow::Result<UserPermissions> {
         tracing::info!("Reloading permissions for repository {repo}");
 
-        let (review_users, review_usernames) = self
+        let review_users = self
```
Kobzol: This doesn't happen that often, but still, as a follow-up improvement, we could fire those 4 requests concurrently.

src/permissions.rs
reddevilmidzy: good idea!

src/permissions.rs
reddevilmidzy: Great! Changed to using `try_join!`

PR #744, opened 2026-04-20
Batch mergeability checks on base branch pushes

kyokuping:
Closes #707

### DIFF
--- a/src/bors/handlers/pr_events.rs
+++ b/src/bors/handlers/pr_events.rs
@@ -16,7 +16,6 @@ use crate::bors::{AUTO_BRANCH_NAME, BorsContext, hide_tagged_comments};
 use crate::bors::{PullRequestStatus, RepositoryState};
 use crate::database::{PullRequestModel, UpsertPullRequestParams};
 use crate::github::CommitSha;
-use crate::utils::text::pluralize;
 use std::sync::Arc;
 
 pub(super) async fn handle_pull_request_edited(
@@ -243,29 +242,22 @@ pub(super) async fn handle_push_to_branch(
     mergeability_queue: &MergeabilityQueueSender,
     payload: PushToBranch,
 ) -> anyhow::Result<()> {
-    let affected_prs = db
-        .set_stale_mergeability_status_by_base_branch(repo_state.repository(), &payload.branch)
+    db.set_stale_mergeability_status_by_base_branch(repo_state.repository(), &payload.branch)
         .await?;
 
-    if !affected_prs.is_empty() {
-        tracing::info!(
-            "Adding {} {} to the mergeability queue due to a new commit pushed to base branch `{}`",
-            affected_prs.len(),
-            pluralize("PR", affected_prs.len()),
-            payload.branch
-        );
+    tracing::info!(
+        "Adding a batch to the mergeability queue due to a new commit pushed to base branch `{}`",
+        payload.branch
+    );
 
-        // Try to find an auto build that matches this SHA
-        let merged_pr = find_pr_by_merged_commit(&repo_state, &db, CommitSha(payload.sha))
-            .await
-            .ok()
-            .flatten()
-            .map(|pr| pr.number);
+    // Try to find an auto build that matches this SHA
+    let merged_pr = find_pr_by_merged_commit(&repo_state, &db, CommitSha(payload.sha))
+        .await
+        .ok()
+        .flatten()
+        .map(|pr| pr.number);
 
-        for pr in affected_prs {
-            mergeability_queue.enqueue_pr(&pr, merged_pr);
-        }
-    }
+    mergeability_queue.enqueue_batch(repo_state.repository(), &payload.branch, merged_pr);
 
     Ok(())
 }
--- a/src/bors/handlers/pr_events.rs
+++ b/src/bors/handlers/pr_events.rs
@@ -836,41 +836,6 @@ report_merge_conflicts = false
     }
 
     #[sqlx::test(migrator = "crate::MIGRATOR")]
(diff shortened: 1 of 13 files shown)


src/bors/mergeability_queue.rs
```diff
@@ -555,63 +646,132 @@ impl MergeabilityQueueReceiver {
...
+                );
+                return Ok(());
+            }
 
-    Ok(())
+            let repo_state = ctx.get_repo(repo)?;
+
+            let mut after = None;
+            loop {
+                let (fetched_prs, cursor) = repo_state
```
Kobzol: We usually abstract paging within the client calls that deal with the GH API, so let's:
- Rename `get_pull_requests_batch` to `get_pull_requests_by_base_branch`
- Handle paging inside `get_pull_requests_by_base_branch`

The most performant way would be to overlap loading the PR pages from GitHub and then querying our database, but I'd leave that for a separate PR, as this change is already complicated enough :)

src/bors/mergeability_queue.rs
```diff
@@ -555,63 +646,132 @@ impl MergeabilityQueueReceiver {
...
+                    .await?;
+
+                for pr in fetched_prs {
+                    let new_mergeable_state = pr.mergeable_state.clone();
+                    if new_mergeable_state == OctocrabMergeableState::Unknown {
+                        match pr.status {
+                            PullRequestStatus::Open | PullRequestStatus::Draft => {
+                                tracing::info!("Mergeability status unknown, scheduling retry.");
+                                mq_tx.enqueue_retry(mq_item);
+                                return Ok(());
```
Kobzol: If we hit a single PR that is in the unknown state, we would stop the whole check, but that is unnecessary. Let's store the PRs that were unknown in a `Vec`, and then after we go through the whole batch, decide whether we then retry the whole batch (if there are, say, more than 20 such PRs) or if we enqueue the individual PRs one-by-one (to avoid downloading 1000 PRs again if only a few of them are left with unknown mergeability).

src/bors/mergeability_queue.rs
```diff
@@ -555,63 +646,132 @@ impl MergeabilityQueueReceiver {
...
+                                    "Mergeability status unknown, but pull request is no longer open."
+                                );
+                            }
+                        }
+                    } else if let Some(db_pr) = ctx
+                        .db
+                        .get_pull_request(repo_state.repository(), pr.number)
+                        .await?
+                    {
+                        update_pr_with_known_mergeability(
```
Kobzol: Can you please wrap this in a span that mentions the PR number? Because currently in the log we only have this:
```
TRACE Mergeability check{item=Batch(PullRequestBatchToCheck { repo: kobzol/bors-kindergarten2, base_branch: "main", attempt: 2, conflict_source: ... })}
```
So it's not really possible to see from the log which PR is getting updated.

src/bors/mergeability_queue.rs
```diff
@@ -555,63 +646,132 @@ impl MergeabilityQueueReceiver {
...
+            ref base_branch,
+            ref attempt,
             conflict_source,
-        )
-        .await?;
-    } else {
-        tracing::warn!("Cannot find DB pull request for {fetched_pr:?}");
-    }
+        }) => {
+            if *attempt >= BATCH_MAX_RETRIES {
```
Kobzol: Could you please add logging at the start and end of the batch check? And record the time it took. So that we can easily see in logs how fast it is.

src/bors/handlers/pr_events.rs
```diff
@@ -243,29 +242,22 @@ pub(super) async fn handle_push_to_branch(
     mergeability_queue: &MergeabilityQueueSender,
     payload: PushToBranch,
 ) -> anyhow::Result<()> {
-    let affected_prs = db
-        .set_stale_mergeability_status_by_base_branch(repo_state.repository(), &payload.branch)
+    db.set_stale_mergeability_status_by_base_branch(repo_state.repository(), &payload.branch)
```
Kobzol: Let's re-add the previous functionality, where we check the `affected_prs`, and only schedule the batch check if at least one PR was affected.

src/github/mod.rs
```diff
@@ -142,6 +142,31 @@ pub struct Branch {
     pub sha: CommitSha,
 }
 
+#[derive(Debug)]
+pub struct PullRequestSummary {
```
Kobzol:
```suggestion
/// Basic information about a PR needed to handle mergeability checks and label changes.
pub struct PullRequestInfo {
```

Kobzol (commented):
I'm really sorry that it took me so long to take a look at this, I didn't have much time for reviews in the past 2-3 months.

The implementation looks great! I tested it on a live repository and it seems to work as expected, it would be cool to get rid of the thousands useless GH API requests to reload the mergeability of all open PRs.

I left some comments regarding the implementation. I haven't looked at the failing tests yet.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/744/edf49e32b0c874622d434400e24fa704af3bdbe8..0bfda8fc55f720a2befbc9e9e56d70e8bf26494a)*

Kobzol:
Thank you! Sorry, I was on a vacation last week. The changes look good, I now also went through the tests, and found that there is one ~obsolete test that we don't need anymore after this change. I also rebased over `main`, as there was a conflict. Let's see if CI passes.

Kobzol:
Looks like it does! I'll merge it now to test this on the staging environment. Thanks a lot for working on this!

Kobzol (approved):

Kobzol:
I tested this on staging and it looks good! I'll probably deploy it to production today to see how it behaves with ~1100 open PRs :sweat_smile:

PR #757, opened 2026-05-23
build: Reduce test build times by tweaking build config

osiewicz:
It seems that with `cargo t --no-run` we're quite blocked on frontend producing CGUs. The following changes reduced incremental build time of `bors` in test scenarios by 40%.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -90,3 +90,4 @@ opt-level = 3
 
 [profile.test]
 codegen-units = 32
+debug = "limited"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,3 +87,6 @@ debug = 1
 
 [profile.dev.package.sqlx-macros]
 opt-level = 3
+
+[profile.test]
+codegen-units = 32


Kobzol:
Thanks for taking a look! Sorry, I'm swamped at the moment, so it might take me some time before I can test this. I'll try to take a look after RustMeet.

Kobzol:
I have some bad news =D

Some notes:
- The default duration for me at the time of testing was around 17-17.5s.
- Setting the CGU count to 32 increases the number of crates compiled in `cargo test` from 455 to 529. Using this:
  ```toml
  [profile.test]
  codegen-units = 32
  
  [profile.test.build-override]
  codegen-units = 32
  ```
  brings it back to 455.
- With CGU=32, the rebuild time is essentially the same, even slightly worse, but that is probably noise.
- With CGU=32 and build-override CGU=32, it's the same as above.
- Setting `debug = "limited"` is what really helps, as it brings the time from ~17s down to ~10s. But I often debug and step through code when testing, so the debuginfo = "limited" change is kinda limiting ( :) ) for me. I might use some kind of `fast-test` profile though for the case where I don't need to debug.

It is clear that debuginfo generation is a big contributor to the slow compile time here (though 10s is still very bad :laughing:).

Moving tests to a separate file, so that only tests (and not the bors `lib` target) is recompiled after a test change, gets the time down to 9s. So the (re-)compilation of tests themselves is just slow, even ignoring the Cargo inefficiency of recompiling also the lib.

PR #769, opened 2026-06-18
Extract Co-authored-by trailers from messages when squashing commits

### LINKED ISSUE #768: Scan existing "Co-authored-by" records when squashing a commit
https://github.com/rust-lang/bors/pull/767 introduced adding "Co-authored-by" (https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) trailer to the commit message when bors does `@bors squash`.

We should ideally also scan the messages of all squashed commits to find pre-existing "Co-authored-by" trailers, parse them out, and add the found authors to the trailer that we generate.

cuba0001:
Fixes #768

### DIFF
--- a/src/bors/handlers/squash.rs
+++ b/src/bors/handlers/squash.rs
@@ -18,6 +18,8 @@ use std::collections::HashSet;
 use std::fmt::Write;
 use std::sync::Arc;
 
+const CO_AUTHORED_BY_TRAILER: &str = "Co-authored-by";
+
 /// Entry point for the squash command.
 /// This function validates the command and enqueues the actual work to the gitops queue.
 pub(super) async fn command_squash(
@@ -127,8 +129,6 @@ pub(super) async fn command_squash(
     // So we instead take the authorship from the commits.
     // We choose the first commit, as it seems like a reasonable expectation that it will be made
     // by the PR author.
-    // Later we also add "Co-authored-by" trailers to add all other authors from the squashed
-    // commits.
     let commit_author = first_commit.author.clone().unwrap_or_else(|| CommitAuthor {
         name: pr.github.author.username.clone(),
         email: pr
@@ -274,13 +274,18 @@ fn add_coauthored_authors(
     commits: &[Commit],
     author: &CommitAuthor,
 ) -> String {
-    // Gather all commit authors.
-    // Note that we ignore previous "Co-authored-by" trailers in the original commit messages, to
-    // simplify the implementation.
-    let mut authors = commits
-        .iter()
-        .filter_map(|commit| commit.author.clone())
-        .collect::<HashSet<CommitAuthor>>();
+    // Gather all commit authors and existing "Co-authored-by" trailers.
+    let mut authors = HashSet::new();
+
+    for commit in commits {
+        if let Some(author) = commit.author.clone() {
+            authors.insert(author);
+        }
+        for coauthor in parse_coauthors(&commit.message) {
+            authors.insert(coauthor);
+        }
+    }
+
     // Remove the author of the final commit, because that will be attributed through normal git
     // commit authorship.
     authors.remove(author);
@@ -299,7 +304,7 @@ fn add_coauthored_authors(
     for author in authors {
         writeln!(
             commit_msg,
-            "Co-authored-by: {} <{}>",
+            "{CO_AUTHORED_BY_TRAILER}: {} <{}>",
             author.name, author.email
         )
         .unwrap();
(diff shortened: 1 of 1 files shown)


cuba0001:
There is one implementation detail worth noting. To keep it simple, we iterate over the commits twice. The first pass is needed when generating the default squash commit message, if no custom `@bors squash msg="..."` message was specified. In that case, we strip existing `Co-authored-by` trailers from the original commit messages. Otherwise they would stay in the generated body and then be added again to the final trailer block. I think this is fine for up to 250 commits, which is the current upper bound, but please let me know if you would prefer this to be done in a single pass.

src/bors/handlers/squash.rs
```diff
@@ -311,11 +314,48 @@ fn add_coauthored_authors(
...
     let mut msg = format!("{pr_title}\n\n");
     for commit in commits {
-        writeln!(msg, "* {}", commit.message).unwrap();
+        writeln!(msg, "* {}", strip_coauthors(&commit.message)).unwrap();
     }
     msg
 }
 
+fn strip_coauthors(message: &str) -> String {
+    // We have to manually strip co-authored-by trailers if overridden commit message is not specified
```
Kobzol, line 325: The duplication doesn't really matter, I think, because only "Co-authored-by" in the end of the commit message will be interpreted by GitHub, I think. But I agree that it's nicer to strip it.

src/bors/handlers/squash.rs
```diff
@@ -311,11 +314,48 @@ fn add_coauthored_authors(
...
+    // or it will be duplicated later
+    let lines = message
+        .lines()
+        .filter(|line| !line.trim_start().starts_with("Co-authored-by:"))
```
Kobzol: Could you please put "Co-authored-by" into a constant, shared by the various functions in this file that use it? Thanks!

src/bors/handlers/squash.rs
```diff
@@ -311,11 +314,48 @@ fn add_coauthored_authors(
...
+
+fn parse_coauthors(message: &str) -> Vec<CommitAuthor> {
+    message.lines().filter_map(parse_coauthor).collect()
+}
+
+fn parse_coauthor(line: &str) -> Option<CommitAuthor> {
+    if !line.trim_start().starts_with("Co-authored-by:") {
+        return None;
+    }
+    let (_, value) = line.trim().split_once(':')?;
```
Kobzol: This could be rewritten in a more "Parse, don't validate" approach, e.g. like this:
```suggestion
    let line = line.trim();
    let (start, value) = line.split_once(':')?;
    if start != "Co-authored-by" {
        return None;
    }
```

src/bors/handlers/squash.rs
```diff
@@ -311,11 +314,48 @@ fn add_coauthored_authors(
...
+
+    let value = value.trim();
+    let (name, email) = value.split_once('<')?;
+    let name = name.trim();
+    let email = email.strip_suffix('>')?.trim();
+
+    // Should we omit if no email presents?
```
Kobzol: We don't really currently require the name or e-mail to be non-empty, they just have to be present somehow. I think it should be fine to remove the condition.

src/bors/handlers/squash.rs
```diff
@@ -311,11 +314,48 @@ fn add_coauthored_authors(
...
+        .collect::<Vec<_>>();
+
+    lines.join("\n").to_string()
+}
+
+fn parse_coauthors(message: &str) -> Vec<CommitAuthor> {
+    message.lines().filter_map(parse_coauthor).collect()
+}
+
+fn parse_coauthor(line: &str) -> Option<CommitAuthor> {
```
Kobzol, line 345: Could you write a few unit tests for this function, please? Thanks!

Kobzol (commented):
Thank you, it looks great! Two passes are fine. I left a few minor comments.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/769/4b03ef5a1243ac9424c30d0e7f4c697a647b2d4d..200e171693fb4fe894995a61453d0df065dc2df2)*

src/bors/handlers/squash.rs
cuba0001: Done.

src/bors/handlers/squash.rs
cuba0001: Hope I got you correctly, done.

src/bors/handlers/squash.rs
cuba0001: Removed.

src/bors/handlers/squash.rs
cuba0001: It definitely should have tests. Done.

Kobzol (approved):
Thank you, looks great! :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/769/4b03ef5a1243ac9424c30d0e7f4c697a647b2d4d..36c9fb280a28f5ed1231a0d80690f7f715b8ccd2)*

PR #778, opened 2026-07-08
Allow attaching notes to PRs

### LINKED ISSUE #772: Support providing reason for `rollup=never` or `rollup=iffy`
Maybe something like `@bors r+ rollup=never subtree update`? It could be named and quoted if desired.

Kobzol:
This PR makes it possible to attach a note to a PR, using one of the following three methods:
- `@bors r+ note=<note>`
- `@bors p=5 <note>`
- `@bors rollup=never <note>`

Note that I used an explicit `note` parameter in the approval command, because it already has two parametrs with arbitrary ordering (`rollup` and `p`), and it was a bit tricky making that work with taking the "rest" of the line as the note. But it could be improved in the future.

The notes are then displayed when hovering on a PR on the queue page. PRs with a note are shown with an asterisk next to their PR number:
image

Fixes: https://github.com/rust-lang/bors/issues/772


Kobzol:
@JonathanBrouwer Let me know if the syntax (and web presentation) is fine for you :)

JonathanBrouwer:
This looks great, thanks!

JonathanBrouwer:
After this PR is merged we should make rust-timer set a note

Kobzol:
> After this PR is merged we should make rust-timer set a note

What note should it set? That a perf. benchmark was done on it? :thinking: Such PRs are already marked with `rollup=never`.

JonathanBrouwer:
> What note should it set? That a perf. benchmark was done on it? :thinking: Such PRs are already marked with `rollup=never`.

My goal would be to be able to easily separate PRs that are rollup=never because of perf from other ones

Kobzol:
I see. I wonder if we should have some special marker for them then, e.g. just a note that says `perf`, and bors could detect that and show e.g. a little stopwatch next to the PR row, or something like that.

PR #782, opened 2026-07-13
Add Beaver favicon to bors

Kobzol:
Suggested in https://rust-lang.zulipchat.com/#narrow/channel/496228-t-infra.2Fbors/topic/favicon/with/609803508.


web/assets/img/favicon.png
marcoieni, line 1: Image

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/782/573b1fb5be18692649a3cf6f97c7cd7dfc312551..5ce0556a2e40a72153c00f16a0f8695976800d8e)*

Kobzol: Fair enough, I replaced it with `.ico`.

web/templates/base.html
```diff
@@ -6,6 +6,8 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
   {% endblock %}
   <title>{% block title %}{% endblock %}</title>
+  <link rel="icon" type="image/png"
+        href="/assets/img/favicon.ico"/>
```
Noratrieb: if it's a png, why is it called .ico? .png works just fine and then GitHub will atually show it. you can downscale the png to make it smaller of course, or use a jpeg or webp

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/782/573b1fb5be18692649a3cf6f97c7cd7dfc312551..e52f2cc2c9932efe53b1019167aa73bea9b01ac5)*

Kobzol: Sorry, the type there is wrong. IIRC .ico was the only format supported by old browsers, but I guess we can just ignore that and use a modern format these days. I started with web dev a long time ago, and I'm still used to thinking about supporting IE 6/7 and stuff :laughing:

Kobzol: I resized the PNG to 48x48 instead of using `.ico`. It's now 4x smaller than the .ico at the same resolution, lol.

Noratrieb: cool cool! yeah ico is a terrible format :D

PR #786, opened 2026-07-23
Add support for launching EC2 runners

Kobzol:
To use optimized EC2 instances for certain CI jobs.


src/ec2/ec2-runner-script.sh
```diff
@@ -0,0 +1,54 @@
...
+    sudo --login -u ubuntu git init rust
+    sudo --login -u ubuntu git -C rust remote add origin https://github.com/Mark-Simulacrum/rust
+    sudo --login -u ubuntu git -C rust config --local gc.auto 0
+    sudo --login -u ubuntu git -C rust fetch --no-tags --prune --no-recurse-submodules --depth=2 origin \
+        +1c8426f372151fd686fd27497827d402d66855f8:refs/remotes/origin/automation/bors/auto
+    sudo --login -u ubuntu git -C rust checkout --progress --force -B automation/bors/auto refs/remotes/origin/automation/bors/auto
+    sudo --login -u ubuntu bash -c 'cd rust && ulimit -n $(ulimit -n -H) && ./src/ci/scripts/checkout-submodules.sh'
+    sudo --login -u ubuntu bash -c 'cd rust && ulimit -n $(ulimit -n -H) && cargo run --manifest-path src/ci/citool/Cargo.toml run-local dist-x86_64-linux'
+    sleep 30
+    shutdown now
```
Mark-Simulacrum: Probably should just delete this :)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/786/4502268b9e2b69d945ad9cebeaeebaeb0d5690ec..568914c8eab04c9245055126202da990bd12b452)*

src/ec2/ec2-runner-script.sh
```diff
@@ -0,0 +1,54 @@
...
+
+if $INSTALL_RUNNER; then
+    # Provision github actions runner
+    # https://github.com/actions/runner/blob/main/docs/start/envlinux.md
+    apt install -y liblttng-ust1t64 libkrb5-3 zlib1g libssl3 libicu78
+    apt install -y docker.io docker-buildx jq python3-pip
+    usermod -a -G docker ubuntu
+    systemctl start docker
+    sudo --login -u ubuntu bash -c 'mkdir actions-runner'
+    sudo --login -u ubuntu bash -c 'cd actions-runner && curl -o runner.tar.gz -L https://github.com/actions/runner/releases/download/v2.335.1/actions-runner-linux-x64-2.335.1.tar.gz'
```
Mark-Simulacrum: I think the runner will self-update, but we'll probably want to figure out a way to bump the tag here automatically (e.g., https://github.com/actions/runner/releases/tag/v2.336.0 has been released since this was written).

I'm not sure if we can just always use latest or not, this sort of implies no:

> Note: Actions Runner follows a progressive release policy, so the latest release might not be available to your enterprise, organization, or repository yet.

IIRC with the old setup we periodically broke and had to bump this constant to unbreak ourselves (30 days after the new release), but there we had disabled the self-update...

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/786/4502268b9e2b69d945ad9cebeaeebaeb0d5690ec..568914c8eab04c9245055126202da990bd12b452)*

src/ec2/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+        .arg("ec2")
+        .arg("run-instances")
+        .arg("--region")
+        .arg(&ec2.region)
+        .arg("--image-id")
+        .arg(format!("resolve:ssm:{image_name}"))
+        .arg("--instance-type")
+        .arg(instance_type)
+        .arg("--launch-template")
+        .arg("LaunchTemplateName=gha-runner,Version=14")
```
Mark-Simulacrum, line 116: Let's add a FIXME or some other tweak to use Latest here before we ship to production.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/786/4502268b9e2b69d945ad9cebeaeebaeb0d5690ec..568914c8eab04c9245055126202da990bd12b452)*

Cargo.toml
```diff
@@ -61,6 +61,7 @@ chrono = "0.4"
 arc-swap = "1"
 itertools = "0.15"
 rand = { version = "0.10", features = ["alloc"] }
+uuid = { version = "1", features = ["v4"] }
```
Mark-Simulacrum: If this is a net-new dependency we can just `format!("{:x}", rand::random::<u128>())` for basically the same effect.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/786/4502268b9e2b69d945ad9cebeaeebaeb0d5690ec..568914c8eab04c9245055126202da990bd12b452)*

Kobzol: Yes, it was new. Ok, fair enough, replaced it.

src/ec2/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+    let Some(image_name) = ec2.images.get(label.image_name) else {
+        return Err(anyhow::anyhow!(
+            "EC2 runner image name {} not found",
+            label.image_name
+        ));
+    };
+
+    // Emulate a "UUID" to avoid adding dependency on the uuid crate just for this one line.
+    let runner_name = format!("{:x}", rand::random::<u128>());
+    let jit_config = match ec2.jit_runner {
```
Mark-Simulacrum, line 66: Maybe worth a comment for future-us on why we have this split, namely that the org-level permission is useful to allow constraining what can execute on the runner to specific workflows while the per-repo one can't do the same? Unless I'm wrong, I see the repo function also takes a runner_group_id...?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/786/4502268b9e2b69d945ad9cebeaeebaeb0d5690ec..e473dce4b0e702f1e840fd50cf4676bd1c04004e)*

src/ec2/ec2-runner-script.sh
```diff
@@ -0,0 +1,41 @@
...
+apt install -y liblttng-ust1t64 libkrb5-3 zlib1g libssl3 libicu78
+apt install -y docker.io docker-buildx jq python3-pip
+usermod -a -G docker ubuntu
+systemctl start docker
+sudo --login -u ubuntu bash -c 'mkdir actions-runner'
+
+# Note: the runner should self-update
+sudo --login -u ubuntu bash -c 'cd actions-runner && curl -o runner.tar.gz -L https://github.com/actions/runner/releases/download/v2.335.1/actions-runner-linux-x64-2.335.1.tar.gz'
+sudo --login -u ubuntu bash -c 'cd actions-runner && tar xzf runner.tar.gz'
+sudo --login -u ubuntu bash -c 'cd actions-runner && ./run.sh --jitconfig "$JITCONFIG"'
```
Mark-Simulacrum, line 39: I guess this script doesn't have `set -euxo pipefail` right now, so even if this all fails we should still reach the shutdown... I think we should leave this as-is for now but I'll continue poking at the AMI stuff so we can remove some of the fallibility in the early parts of this.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/786/4502268b9e2b69d945ad9cebeaeebaeb0d5690ec..e473dce4b0e702f1e840fd50cf4676bd1c04004e)*

src/ec2/ec2-runner-script.sh
```diff
@@ -0,0 +1,41 @@
...
+# Provision github actions runner
+# https://github.com/actions/runner/blob/main/docs/start/envlinux.md
+apt install -y liblttng-ust1t64 libkrb5-3 zlib1g libssl3 libicu78
+apt install -y docker.io docker-buildx jq python3-pip
+usermod -a -G docker ubuntu
+systemctl start docker
+sudo --login -u ubuntu bash -c 'mkdir actions-runner'
+
+# Note: the runner should self-update
+sudo --login -u ubuntu bash -c 'cd actions-runner && curl -o runner.tar.gz -L https://github.com/actions/runner/releases/download/v2.335.1/actions-runner-linux-x64-2.335.1.tar.gz'
```
Mark-Simulacrum, line 37: Let's leave this as-is, it should exercise self-updates, but I suspect we'll want to find a way to continuously bump this. Probably easier to do once we have a dedicated workflow for building the AMI in a separate repo, maybe we can get renovate to do something...

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/786/4502268b9e2b69d945ad9cebeaeebaeb0d5690ec..e473dce4b0e702f1e840fd50cf4676bd1c04004e)*

src/ec2/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+        .arg("terminate")
+        .arg("--launch-template")
+        // FIXME: use the latest version before pushing to production
+        .arg("LaunchTemplateName=gha-runner,Version=14")
+        .arg("--user-data")
+        .arg(script);
+
+    let output = run_command(&mut ec2_cli)
+        .await
+        .context("Cannot start ec2 instance")?;
```
Mark-Simulacrum, line 122: Will this "leak" the raw error from EC2 to github comments or similar? I think there shouldn't be anything too interesting there but maybe worth logging it just to bors's logs rather than externally.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/786/4502268b9e2b69d945ad9cebeaeebaeb0d5690ec..e473dce4b0e702f1e840fd50cf4676bd1c04004e)*

src/ec2/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    if !too_old_ids.is_empty() {
+        let too_old_ids = too_old_ids.join(",");
+        tracing::info!("Cancelling EC2 instance(s) {too_old_ids}");
+
+        run_command(
+            prepare_aws_cli()
+                .arg("ec2")
+                .arg("terminate-instances")
+                .arg("--instance-ids")
```
Mark-Simulacrum, line 202: Let's add `--force` and `--skip-os-shutdown` here. It's not entirely clear to me whether `--force` does us much good, but I don't think there's value in gracefully stopping these instances if we've exceeded the timeout.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/786/4502268b9e2b69d945ad9cebeaeebaeb0d5690ec..e473dce4b0e702f1e840fd50cf4676bd1c04004e)*

src/ec2/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+        .await
+        .context("Cannot start ec2 instance")?;
+    let launched: serde_json::Value = serde_json::from_str(&output)?;
+    tracing::info!(
+        "Launched {instance_type}: {}",
+        launched["Instances"][0]["InstanceId"]
+            .as_str()
+            .unwrap_or("unknown instance id")
+    );
+
```
Mark-Simulacrum, line 130: I wonder if it would be worth capping the number of instances we're willing to run somehow... not sure exactly how we'd do that though. I think we'll hit account quota limits at 640 vCPUs (~13 12xlarge instances) so maybe that's good enough? I think we can circle back to this if we run into issues in practice.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/786/4502268b9e2b69d945ad9cebeaeebaeb0d5690ec..e473dce4b0e702f1e840fd50cf4676bd1c04004e)*

src/config.rs
```diff
@@ -42,6 +42,41 @@ pub struct RepositoryConfig {
...
+#[serde(deny_unknown_fields)]
+pub struct Ec2RunnersConfig {
+    /// GitHub Actions self-hosted runner group ID in which should the EC2 runner be launched.
+    pub runner_group_id: u64,
+    /// Prefix of labels (the value of the `runs-on` field of a GitHub Actions job)
+    /// that will trigger the launch of an EC2 runner.
+    pub label_prefix: String,
+    /// In which AWS region should the runners be launched.
+    pub region: String,
+    /// Image types allowed to be launched.
```
Mark-Simulacrum, line 71: I think I would document this as "Human readable names => AMI SSM parameter name", and maybe give the example (`"ubuntu26.04": "/aws/service/canonical/ubuntu/server/26.04/stable/current/amd64/hvm/ebs-gp3/ami-id"`).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/786/4502268b9e2b69d945ad9cebeaeebaeb0d5690ec..e473dce4b0e702f1e840fd50cf4676bd1c04004e)*

src/config.rs
```diff
@@ -42,6 +42,41 @@ pub struct RepositoryConfig {
...
+    pub images: HashMap<String, String>,
+    #[serde(default)]
+    /// The kind of JIT runner config to create.
+    pub jit_runner: JitRunnerKind,
+    /// Allowed instance types that can be launched.
+    /// If empty, bors will launch any instance specified in the `runs-on` field.
+    #[serde(default)]
+    pub allowed_instances: Vec<String>,
```
Mark-Simulacrum, line 86: Hm, this is read from the main branch, right? It's a bit of a pain that we'd need to land a PR to allow experimentation with a new instance type (or use infra bypass to get the token like I did originally). It's probably OK though, we can either remove the restriction or figure out some way to relax it if we need to later.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/786/4502268b9e2b69d945ad9cebeaeebaeb0d5690ec..e473dce4b0e702f1e840fd50cf4676bd1c04004e)*

Mark-Simulacrum (approved):

src/ec2/mod.rs
Kobzol: The repo one also takes runner group ID, yeah. It basically only exists for easier local debugging. I'll add a comment.

src/ec2/mod.rs
Kobzol: No, background job errors are never posted as GitHub comments. Even errors during command handling do not expose the inner error details, those are only logged.

src/ec2/mod.rs
Kobzol: Even the graceful shutdown didn't help with GitHub getting stuck on the running GHA job :)

src/config.rs
Kobzol: We can experiment in `bors-kindergarten`, there a change of the config is a push to `main`, which is ~instantly propagated to bors (well, takes up to 15 minutes without restarting bors, but that's enough for experimentation, I think).

Kobzol:
In addition to this PR, we need to:
- Configure `Workflow job` webhook events for staging/production bors GitHub App.
- Configure organization permission `Self-hosted runners (write)` for staging/production bors GitHub App.

Can you configure it, please?

src/config.rs
Mark-Simulacrum: Kindergarten doesn't let us evaluate runner duration on jobs from r-l/r though, right?

Mark-Simulacrum:
Configured both staging and prod with the relevant permissions, I think.

Mark-Simulacrum (approved):

src/config.rs
Kobzol: Yeah, that's true. We can hardcode a set of default options later. Getting a PR merged here and redeploying production bors should be much faster than landing a `rust-lang/rust` PR :laughing:

Issue #690, opened 2026-02-10
Store jobs of pending workflows in memory


Kobzol:
It would be nice to show which CI *jobs* are still being computed (are pending) for the currently pending *workflow* on the queue page. However, given that this is just a nice-to-have feature, I don't want to complicate it by involving the DB. So I would just add a new webhook listener that will listen for job changes, and store their state in memory, in a non-persistent way, probably in the `GitHubRepositoryClient` struct (so that it is per repo). There should be some LRU cache or something, to avoid it consuming a lot of memory. In reality, we always care only about the jobs of a pending workflow, so we can simply dump all jobs of workflows that are already completed.

medzernik:
i have pushed some things in my repo but cant make it in time. as discussed, 

@rustbot unclaim

PR #788, opened 2026-07-23
Show expected time left for pending PRs

camelid:
Or show how long since the expected completion time for slow PRs.

### DIFF
--- a/web/templates/queue.html
+++ b/web/templates/queue.html
@@ -331,7 +331,7 @@ Average duration of recent auto builds: {{ format_duration(*average_build_durati
                         {%- else -%}
                           pending
                         {%- endif -%}
-                        <span class="elapsed-time-display"></span>
+                        <span class="time-display"></span>
                         </div>
                       {%- when QueueStatus::Failed(_, _) -%}
                         failed
@@ -739,7 +739,7 @@ Average duration of recent auto builds: {{ format_duration(*average_build_durati
     });
     {% endif %}
 
-    function formatElapsedTime(ms) {
+    function formatDuration(ms) {
         const totalSeconds = Math.floor(ms / 1000);
         const hours = Math.floor(totalSeconds / 3600);
         const remainingSeconds = totalSeconds % 3600;
@@ -758,10 +758,7 @@ Average duration of recent auto builds: {{ format_duration(*average_build_durati
         } else if (output.length === 0) {
             output = '<1m';
         }
-        if (output.length > 0) {
-            return ` (${output})`;
-        }
-        return '';
+        return output;
     }
 
     const AVERAGE_BUILD_DURATION_MS = {{ average_build_duration.as_millis() }};
@@ -772,9 +769,10 @@ Average duration of recent auto builds: {{ format_duration(*average_build_durati
             const createdAtMs = parseInt(td.dataset.createdAt, 10);
             if (createdAtMs > 0) {
                 const elapsedMs = now - createdAtMs;
-                const formattedTime = formatElapsedTime(elapsedMs);
-                const display = td.querySelector('.elapsed-time-display');
-                display.textContent = formattedTime;
+                const expectedMs = AVERAGE_BUILD_DURATION_MS - elapsedMs;
+                const formattedExpected = expectedMs < 0 ? `${formatDuration(-expectedMs)} overdue` : `${formatDuration(expectedMs)} left`;
+                const timeDisplay = td.querySelector('.time-display');
+                timeDisplay.textContent = ` (${formattedExpected})`;
 
                 // Update progress bar width and ARIA attribute
                 const progressBar = td.querySelector('.progress-bar-bg');
--- a/web/templates/queue.html
+++ b/web/templates/queue.html
@@ -455,7 +455,8 @@ Average duration of recent auto builds: {{ format_duration(*average_build_durati
             columns: [
                 { name: "select", orderable: false, className: "select-checkbox", width: "10px" },
                 { name: "number", className: "dt-right" },
-                { name: "status" },
+                // Ensure that the status doesn't have too much padding on the right side
+                { name: "status", width: "5%" },
                 { name: "mergeable" },
                 { name: "title", width: "40%" },
                 { name: "author" },


web/templates/queue.html
```diff
@@ -772,9 +769,11 @@ <h1>
             const createdAtMs = parseInt(td.dataset.createdAt, 10);
             if (createdAtMs > 0) {
                 const elapsedMs = now - createdAtMs;
-                const formattedTime = formatElapsedTime(elapsedMs);
-                const display = td.querySelector('.elapsed-time-display');
-                display.textContent = formattedTime;
+                const formattedElapsed = formatDuration(elapsedMs);
+                const expectedMs = AVERAGE_BUILD_DURATION_MS - elapsedMs;
+                const formattedExpected = expectedMs < 0 ? `~${formatDuration(-expectedMs)} overdue` : `~${formatDuration(expectedMs)} left`;
```
camelid: Not sure if "overdue" is the best wording. Open to suggestions. Also, is there some way to test the display of the queue locally without setting up a whole fake database etc.?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/788/4502268b9e2b69d945ad9cebeaeebaeb0d5690ec..cfef57d0bb5f6912ec775ee6dcc036d778398eac)*

Kobzol (commented):
Hi, thanks for the PR! I think that showing the remaining time is a good idea, but I'm not sure if we should format it like this, because it can become quite wide:

Image

What do you think about:
- Removing `pending` (since that is kinda implied)
- Keeping only the expected part, so that it will always show only "1h 3m left" or "5m overdue"
?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/788/4502268b9e2b69d945ad9cebeaeebaeb0d5690ec..cfef57d0bb5f6912ec775ee6dcc036d778398eac)*

camelid:
I also don't love this format (thanks for sending the preview btw -- I don't have a local setup for bors). However, I feel like we should still say pending or something similar. It's obvious to us but might not be to newbies. We could use icons instead of "pending" and "approved". Could be cool but I kinda like the simplicity of using words.

Another option is keeping expected if it's "1h 3m left", otherwise say how much is elapsed (and sidestep the whole overdue wording issue as well). E.g. it'd be either "pending (1h 3m left)" or "pending (5h so far)" if it's slow. Though then it's not obvious how much overdue it is.

It'd be nice to keep the absolute elapsed time rather than just the relative-to-average delta, though you might be right that just showing the delta makes more sense.

Kobzol:
By the way, is the progress bar not enough for showing the expected time left? :thinking:

camelid:
Not really because then you need to do mental arithmetic. Like if it's 2 hours done and the bar looks around 70% full, I pretty much have to do 2/0.7 - 2, then multiply by 60, if I want to know how many minutes are left. This is why progress bars in terminals and other apps usually show visually how much is done, _and_ the percentage done, _and_ the time spent so far, _and_ the expected time left. Better to make the information easily available.

Kobzol:
Showing everything (but still without %) doesn't look great:
image

I think that a compromise would be to show the expected remaining time, that might be a bit more useful than the elapsed time.

Kobzol:
Thanks, this should be fine visually (I'd add some more CSS to limit the width of the column, but that's a detail).

I still wonder about the shown message. We say "N minutes left", but the variability of our CI is 20-30 minutes, so it can be a bit misleading to say "left" and "overdue". But anyway, I think we can try it and see what others think.

Kobzol (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/788/b4434436072825b92291d1bc6c39e96d2d8f5af7..ee599e25ad59124b8510c32e1613cfaf5b878c45)*

Zalathar:
To me this seems more confusing and less reliable than just showing the elapsed time.

Now there's no way to see how long the current job has been running, which is important information.

And the estimated time remaining is simply not reliable, so I don't see it as useful.

camelid:
> Now there's no way to see how long the current job has been running, which is important information.

Note that you can still see this information by hovering over the "pending" entry (showing the HTML `title`).

> And the estimated time remaining is simply not reliable, so I don't see it as useful.

It is roughly accurate; there is just some error regarding the _exact_ time remaining.

I just don't think that showing only elapsed time is super useful since there isn't a reference point for how long a normal build is. The advantage of this PR too is now it's easier to identify builds that are slower than usual since they are marked as overdue. Open to any suggestions of better UIs for this.

Zalathar:
> Note that you can still see this information by hovering over the "pending" entry (showing the HTML `title`).

No, you can't. Try it. There is no way to see the elapsed time.

In addition, hovering is a huge pain compared to being able to *look at the page*, and doesn't work at all on mobile.

> It is roughly accurate; there is just some error regarding the _exact_ time remaining.

It's not accurate at all. The time estimator has no idea when the current job will stop, which could be anywhere between 0 hours and 6 hours from when it started. It can't possibly know.

> I just don't think that showing only elapsed time is super useful since there isn't a reference point for how long a normal build is. The advantage of this PR too is now it's easier to identify builds that are slower than usual since they are marked as overdue. Open to any suggestions of better UIs for this.

This change makes it way harder to identify slow builds, because it won't tell me how long the current build has been running!

We already had a good UI, and it was the previous one. I'm pretty upset that a successful design was made massively less useful for reasons that make no sense to me.

Kobzol:
Zalathar is doing a lot of rollup work and PR scheduling, so I will fully defer to their (and also [others'](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/bors.20queue.20not.20showing.20elapsed.20time/with/613798686)) preference on this, and revert this PR (sorry, Noah!).

That being said, I also find the expected remaining time to be kinda useful.

And this conversation just gave me an idea on how to:
1. Show both elapsed and expected remaining time
2. Resolve https://github.com/rust-lang/bors/issues/690
3. And also show try builds on the main queue page

With a single solution. I'll try to work on it today.

Kobzol:
Implemented this in https://github.com/rust-lang/bors/pull/795. It can be later extended to also show pending try builds.

camelid:
No worries, I like your new approach better anyway. I agree that elapsed time is useful, but like you I'd also like to see estimated time left.

PR #790, opened 2026-07-28
Added integration tests for sending events to Zulip and mock server

### LINKED ISSUE #785: Add integration tests for Zulip
https://github.com/rust-lang/bors/pull/784 added a Zulip client, but I haven't yet implemented tests for it. It would be nice to implement a mocked Zulip server, same as we do for GitHub, and use it to test the sent Zulip messages.

cuba0001:
Fixes: https://github.com/rust-lang/bors/issues/785

### DIFF
--- a/src/bors/handlers/review.rs
+++ b/src/bors/handlers/review.rs
@@ -622,7 +622,7 @@ mod tests {
     use crate::tests::{BorsTester, Commit};
     use crate::{
         bors::{RollupMode, handlers::trybuild::TRY_MERGE_BRANCH_NAME},
-        tests::{BorsBuilder, Comment, GitHub, Permissions, User, run_test},
+        tests::{BorsBuilder, Comment, GitHub, Permissions, User, ZulipMessage, run_test},
     };
 
     #[sqlx::test(migrator = "crate::MIGRATOR")]
@@ -1140,6 +1140,18 @@ approved = { modifications = ["+foo", "+baz"], unless = ["label1", "label2"] }
                 ctx.get_next_comment_text(()).await?,
                 @"Tree closed for PRs with priority less than 5."
             );
+            assert_eq!(
+                ctx.zulip_messages().await?,
+                vec![ZulipMessage::stream(
+                    242791,
+                    "Tree ops",
+                    format!(
+                        "The tree of `{}` was [closed](https://github.com/{}/pull/1#issuecomment-1) for PRs with priority below `5` by `default-user`.",
+                        default_repo_name(),
+                        default_repo_name()
+                    )
+                )]
+            );
 
             let repo = ctx.db().repo_db(&default_repo_name()).await?;
             assert_eq!(
@@ -1162,10 +1174,23 @@ approved = { modifications = ["+foo", "+baz"], unless = ["label1", "label2"] }
     #[sqlx::test(migrator = "crate::MIGRATOR")]
     async fn tree_closed_with_reason(pool: sqlx::PgPool) {
         run_test(pool, async |ctx: &mut BorsTester| {
-            ctx.post_comment("@bors treeclosed=5 CI is flaky").await?;
+            ctx.post_comment("@bors treeclosed=5 CI is flaky (@infra)")
+                .await?;
             insta::assert_snapshot!(
                 ctx.get_next_comment_text(()).await?,
                 @"Tree closed for PRs with priority less than 5."
+            );
+            assert_eq!(
+                ctx.zulip_messages().await?,
+                vec![ZulipMessage::stream(
+                    242791,
+                    "Tree ops",
+                    format!(
+                        "The tree of `{}` was [closed](https://github.com/{}/pull/1#issuecomment-1) for PRs with priority below `5` by `default-user`.\n\nReason: CI is flaky (infra)\n",
+                        default_repo_name(),
+                        default_repo_name()
+                    )
+                )]
             );
 
             let repo = ctx.db().repo_db(&default_repo_name()).await?;
@@ -1173,7 +1198,7 @@ approved = { modifications = ["+foo", "+baz"], unless = ["label1", "label2"] }
                 repo.unwrap().tree_state,
                 TreeState::Closed {
                     priority: 5,
-                    reason: Some("CI is flaky".to_string()),
+                    reason: Some("CI is flaky (@infra)".to_string()),
                     source: format!(
                         "https://github.com/{}/pull/1#issuecomment-1",
                         default_repo_name()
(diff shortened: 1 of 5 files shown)


src/bors/handlers/review.rs
```diff
@@ -1140,6 +1140,18 @@ approved = { modifications = ["+foo", "+baz"], unless = ["label1", "label2"] }
                 ctx.get_next_comment_text(()).await?,
                 @"Tree closed for PRs with priority less than 5."
             );
+            assert_eq!(
```
Kobzol: It gets super annoying to update these test assertions when they change, which is why I prefer to use `insta` as much as possible.

Could you please turn this into something like:
```rus
let messages = ctx.zulip_messages().await?;
insta::assert_debug_snapshot(messages, @"");
```

src/bors/handlers/review.rs
```diff
@@ -1162,18 +1174,31 @@ approved = { modifications = ["+foo", "+baz"], unless = ["label1", "label2"] }
...
     async fn tree_closed_with_reason(pool: sqlx::PgPool) {
         run_test(pool, async |ctx: &mut BorsTester| {
-            ctx.post_comment("@bors treeclosed=5 CI is flaky").await?;
+            ctx.post_comment("@bors treeclosed=5 CI is flaky (@infra)")
+                .await?;
             insta::assert_snapshot!(
                 ctx.get_next_comment_text(()).await?,
                 @"Tree closed for PRs with priority less than 5."
             );
+            assert_eq!(
```
Kobzol: Same here. Could you please add a comment here that says that we want to ensure that the message gets `@` stripped away? Thanks!

src/bors/handlers/review.rs
```diff
@@ -1214,6 +1283,7 @@ approved = { modifications = ["+foo", "+baz"], unless = ["label1", "label2"] }
                     ctx.get_next_comment_text(()).await?,
                     @"@default-user: :key: Insufficient privileges: not in review users"
                 );
+                assert!(ctx.zulip_messages().await?.is_empty());
```
Kobzol: Could you please add this to `BorsTester::finish`? To ensure that we didn't get any unexpected Zulip messages in tests. I think it could be added to `assert_empty_queues`.

To make this work, it would be good to change `async fn zulip_messages` to always *take* the received message, and flush it from the Zulip message queue. Otherwise we won't be able to recognize if a test read the message or not.

Kobzol (commented):
Hi, thank you, this looks good! One thing I'd like to modify is to remember the received messages "dynamically", so that we can check at the end of each test whether we got any expected messages or not. But for that we won't be able to use `received_requests` from `wiremock` anymore.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/790/4502268b9e2b69d945ad9cebeaeebaeb0d5690ec..fd3cff826bf2fc25f81a600ba180ccd2dd002600)*

cuba0001:
@Kobzol 

Done! Added queue for received messages, added a check for receiving unexpected messages, also it required to add a check of messages count to the existing tests (which also flushes received messages), replaced assertions with insta, comment that says a message gets `@` stripped away from reason.

src/tests/mock/zulip.rs
```diff
@@ -0,0 +1,181 @@
...
+    mock_server: MockServer,
+    received_messages: Arc<Mutex<Vec<anyhow::Result<ZulipMessage>>>>,
+}
+
+impl ZulipMockServer {
+    pub(super) async fn start() -> Self {
+        let mock_server = MockServer::start().await;
+        let received_messages = Arc::new(Mutex::new(Vec::new()));
+
+        Mock::given(any())
```
Kobzol: Could you please put back the explicit messages URL? It both helps the reader understand what is the test mocking, and it also ensures that we won't be mocking unexpected endpoints by accident.

src/tests/mock/zulip.rs
```diff
@@ -0,0 +1,181 @@
...
+            .respond_with({
+                let received_messages = received_messages.clone();
+
+                move |request: &Request| {
+                    received_messages
+                        .lock()
+                        .push(ZulipMessage::try_from(request));
```
Kobzol: Let's panic here if the message cannot be parsed, rather than putting results into the queue. The bors tester infrastructure should be able to detect panics within the mocked server.

src/tests/mock/zulip.rs
```diff
@@ -0,0 +1,181 @@
...
+        ZulipClient::new(
+            self.mock_server.uri(),
+            TEST_ZULIP_USERNAME.to_string(),
+            TEST_ZULIP_TOKEN.into(),
+        )
+        .unwrap()
+    }
+
+    pub(super) fn take_received_messages(&self) -> anyhow::Result<Vec<ZulipMessage>> {
+        let received_messages: Vec<_> = self.received_messages.lock().drain(..).collect();
```
Kobzol: (this code should be removed based on my previous suggestion, but here's a tip: you can collect a vec of results into a result of a vec)

```suggestion
        let received_messages: Vec<ZulipMessage> = self.received_messages.lock().drain(..).collect::<anyhow::Result<Vec<ZulipMessage>>()?;
```

src/tests/mock/zulip.rs
```diff
@@ -0,0 +1,181 @@
...
+            match message {
+                Ok(message) => messages.push(message),
+                Err(error) => return Err(error),
+            }
+        }
+
+        Ok(messages)
+    }
+
+    pub(super) fn assert_empty_queue(&self) -> anyhow::Result<()> {
```
Kobzol: This should do the same thing as the GitHub mock, to stop the mocked server (the function will have to become `async`):
```rust
self.mock_server.reset().await;
drop(self.mock_server);
```

Kobzol (commented):
Thank you! Left a few more comments based on the edits.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/790/d0f416f8742b7683c2334652ecb81eab28fdeea4..0a1cb3c4017a50799dfc5c7ad6ff3c1c3b6cec9b)*

cuba0001:
@Kobzol 
Sorry for the delay. I improved the code according to your comments.

Kobzol (approved):
Thank you, great work!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/790/e890af6f5ebf4500a13f69e8ebd0b1eec3cbc347..8d98216f49c278d82f5855ccfac1b7b4246be1ed)*

PR #795, opened 2026-07-31
Add table with pending builds to the queue page

### LINKED ISSUE #690: Store jobs of pending workflows in memory

Kobzol:
This PR adds a new table with pending builds to the queue page. The motivation is to have some space that could show extended information about pending builds (https://github.com/rust-lang/bors/pull/788, https://github.com/rust-lang/bors/issues/690) and eventually also (optionally) show running try builds.

When bors receives workflow job stated/completed webhooks, it will now cache them in-memory (not in the DB, as it is only best effort) and display them in the build table. After too many workflows are accumulated in memory, the oldest ones are removed.

We recognize oldest workflows by their run ID; bors assumes that the IDs are sequentially increasing. If that won't be true in practice, we can also store e.g. their started time.

The jobs might not always be accurate if we miss some webhooks (due to them being dropped or bors being redeployed). Bors could at startup load currently running workflows and load their jobs, but I didn't include that in this PR, it's already large enough.

The UI is not super pretty at the moment, improvements/suggestions are welcome :laughing: It is quite annoying to do changes here though, my fingers are really twitching to ports the whole page to Vue.

image

Fixes: https://github.com/rust-lang/bors/issues/690


Zalathar:
I really appreciate that the table can be hidden when not needed, and remembers its visibility across reloads.

Kobzol:
Yeah, I tried to make sure that it will be unobtrusive, if possible. The button is still pretty big, and the UI to hide/show the table could be likely much better, but doing this in pure HTML/JS/CSS is becoming quite annoying.

PR #806, opened 2026-08-04
Try to start EC2 instances on any branch

Kobzol:
Should fix EC2 instances not starting on the `try-perf` branch.

### DIFF
--- a/src/bors/handlers/workflow.rs
+++ b/src/bors/handlers/workflow.rs
@@ -155,6 +155,10 @@ pub(super) async fn handle_workflow_job_started(
     repo: Arc<RepositoryState>,
     payload: WorkflowJobStarted,
 ) -> anyhow::Result<()> {
+    if let Err(error) = try_start_ec2_instance(ctx, &db, &repo, &payload).await {
+        tracing::error!("Cannot start EC2 instance: {error:?}");
+    }
+
     let Some(build_kind) = get_build_kind_from_branch(&payload.branch) else {
         return Ok(());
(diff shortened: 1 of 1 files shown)


src/bors/handlers/workflow.rs
```diff
@@ -212,17 +227,19 @@ pub(super) async fn handle_workflow_job_started(
         }
     };
 
+    // If we don't know what kind of branch it is, we just assume that it is a try build
+    let build_kind = get_build_kind_from_branch(&payload.branch);
+
+    // We try to spawn EC2 instances for all kinds of jobs, even those outside of try/auto branches
```
Mark-Simulacrum, line 233: If we spawn an EC2 instance for a non-allowlisted branch it'll just sit idle for 6 hours and then get terminated, which doesn't seem great. Can we stop this from happening for PR builds at least? Not sure what the branch for those is.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/806/5678b3aeb12917a266895b6006d84b3549a04281..a8e5d03fc7b2c312b94d4035d380ed91ab0d6243)*

Kobzol: I can do that tomorrow, this is more of a hotfix. We don't currently have any PR jobs that would use EC2, so it shouldn't be a problem.

Mark-Simulacrum: Yeah, more defense in depth.

PR #814, opened 2026-08-11
Simplify startup script

Mark-Simulacrum:
Now that we are baking AMIs, we can remove most of the complexity from bors related to starting instances.

This should also speed up time waiting for the instance to come up, though it already wasn't particularly long.

I think this should work but we should obviously verify on staging.

### DIFF
--- a/src/ec2/ec2-runner-script.sh
+++ b/src/ec2/ec2-runner-script.sh
@@ -1,41 +1,9 @@
 #!/bin/bash
 
-export DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Thanks, let's try on staging.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/814/0ab7cc6e51c07606141c499e3360556c8cd3e4df..f5885e9f1fad534f6a9ee56897181cb3c9e06e09)*

PR #820, opened 2026-08-14
Correctly parse unrolled member build kind from EC2 instance tags

Kobzol:
And also do not log sensitive information.

### DIFF
--- a/src/ec2/mod.rs
+++ b/src/ec2/mod.rs
@@ -469,6 +469,7 @@ pub async fn get_ec2_instances(
                     .and_then(|build_kind| match build_kind.as_str() {
                         "try" => Some(BuildKind::Try),
                         "auto" => Some(BuildKind::Auto),
+                        "try-perf" => Some(BuildKind::UnrolledMember),
                         _ => None,
                     })?;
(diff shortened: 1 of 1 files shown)


src/ec2/mod.rs
```diff
@@ -572,17 +574,30 @@ fn prepare_aws_cli(
 }
 
 async fn run_command(cmd: &mut tokio::process::Command) -> anyhow::Result<String> {
+    fn format_command(cmd: &tokio::process::Command) -> String {
+        // The command probably has secrets in ENV, so only print the command-line arguments
```
Mark-Simulacrum, line 578: Aren't we passing user-data via arguments too? That has the JIT token, which is semi-sensitive.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/820/8ef3fcb2f120933b29d5446c837597e611c0dd30..8daae6fb96c7e33823e44c8a01b445473034e6e4)*

Kobzol: Good point, I'll try to remove it.

PR #822, opened 2026-08-16
Make gha-runner startup non-blocking in user-data

Mark-Simulacrum:
With changes in the AMIs (https://github.com/rust-lang/simpleinfra/pull/1170), this was showing up as two different units running:

```
[ ***  ] (2 of 2) Job gha-runner.service/start running (2min 59s / 3min)
[***   ] (1 of 2) Job cloud-final.service/start running (3min 13s / no limit)
```

which is needlessly verbose. It also might block some processing done by cloud-init after the user data (not sure).

If we ever added anything to run after the unit has started this might need tweaking but for now it seems pretty reasonable.

As a drive-by improvement this also adjusts the script to have a single JITCONFIG variable instance, we were replacing both previously.

I've confirmed the `--no-block` aspect works with my manual run scripts, but of course we should double check on staging after merging this. The output now has just a single unit:

```
[    **] Job gha-runner.service/start running (53s / 3min)
```

### DIFF
--- a/src/ec2/ec2-runner-script.sh
+++ b/src/ec2/ec2-runner-script.sh
@@ -1,9 +1,18 @@
 #!/bin/bash
 
-# $JITCONFIG is replaced by bors with the actual JIT token.
+# JITCONFIG is *textually* replaced by bors with the actual JIT token.
+# The JIT token is base64 encoded by GitHub, so it shouldn't have any single quotes in the string.
+#
+# This is read by the gha-runner service started below.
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Ah, so that's why the cloud-init stuff was interleaved with interesting stuff in the logs. Thanks! Also a good catch with the double replacement.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/822/32622f567b734e38b409e9bf9953c1bae1356846..197b2a1e32f128b6bb6b99e2562e5851183b6491)*

PR #828, opened 2026-08-22
Squash & Approve

### LINKED ISSUE #691: Implement `@bors r+ squash`
It is natural to combine `r+` and the `squash` operation, so it would be nice to add something like `@bors r+ squash[=<commit message>]`, which would first start the squash operation and once it successfully finishes, it would approve the newly squashed commit and send a comment.

medzernik:
Added a Squash and Approve function. This should fix #691. 

Please let me know anything I can do to improve the code if needed.

I would add that in
     https://github.com/medzernik/bors/blob/ab07984372498f757ab06438e926a6a6c0424c2b/src/bors/handlers/squash.rs#L25-L27

I wanted to use some more differentiated error states, but with anyhow and the general practice of returning `Ok(X)` even if an error technically occurs, I just wrapped the `sha` in an `Option<sha>`

Mentored by: @Kobzol


medzernik:
Thanks! I'll try to fix the remaining errors in tests and get it cleaned up today or tommorow.

src/bors/mod.rs
```diff
@@ -106,6 +106,7 @@ pub fn format_help() -> &'static str {
         BorsCommand::Retry => {}
         BorsCommand::Cancel => {}
         BorsCommand::Squash { .. } => {}
+        BorsCommand::SquashAndApprove { .. } => {}
```
Kobzol, line 109: The command should be documented in the help message below.

src/bors/handlers/mod.rs
```diff
@@ -668,6 +672,78 @@ async fn handle_comment(
...
+                            let db2 = database.clone();
+                            let pr_github = pr_github.clone();
+                            let comment_author = comment.author.clone();
+                            let merge_queue_tx = senders.merge_queue().clone();
+                            let callback: AfterSquashCallback = Box::new(move |sha: CommitSha| {
+                                Box::pin(async move {
+                                    let pr_db = db2
+                                        .get_pull_request(repo2.repository(), pr_github.number)
+                                        .await?
+                                        .expect("TODO");
```
Kobzol, line 696: If this fails, we should return an `Err` from the callback, saying that the PR was not found in the DB.

src/bors/handlers/squash.rs
```diff
@@ -881,6 +893,72 @@ also include this pls
...
+            ctx.approve(()).await?;
+            ctx.pr(()).await.expect_added_labels(&["approved"]);
+            Ok(())
+        })
+        .await;
+    }
+
+    #[sqlx::test(migrator = "crate::MIGRATOR")]
+    async fn squash_two_commits_and_approve(pool: sqlx::PgPool) {
+        let pool = Arc::new(pool);
```
Kobzol, line 915: Some leftover?

src/bors/handlers/squash.rs
```diff
@@ -881,6 +893,72 @@ also include this pls
...
+                <sqlx::PgPool as Clone>::clone(&*(pool.clone())),
+                squash_state(),
+            ),
+            async |ctx: &mut BorsTester| {
+                ctx.modify_pr_in_gh((), |pr| {
+                    pr.title = "Foobar".to_string();
+                    pr.reset_to_single_commit(Commit::from_sha("sha1"));
+                    pr.add_commits(vec![Commit::from_sha("sha2")]);
+                });
+                ctx.post_comment("@bors squash").await?;
```
Kobzol, line 927: This should be using `@bors r+ squash`, which is what we want to test, right?

src/bors/command/mod.rs
```diff
@@ -174,5 +174,20 @@ pub enum BorsCommand {
     /// Cancel an auto build currently running on a given PR (without removing it from the queue).
     Cancel,
     /// Squash all commits of a pull request into a single commit.
-    Squash { commit_message: SquashCommitMessage },
+    Squash {
+        /// Squash message for the commit
+        commit_message: SquashCommitMessage,
+    },
+    SquashAndApprove {
```
Kobzol, line 181: We need to actually parse the command 😆 I was surprised why it was doing nothing in tests.. 😆

Kobzol (commented):
Thanks! Left some comments.

There is one race condition that we'll need to handle in `handle_push_to_pull_request`. If we push the squashed commit, GitHub will send us a webhook about the push, which will unapprove the PR in `handle_push_to_pull_request`. Of course that is not ideal if we just approved it after the squash :)

I will push some unrelated refactoring towards that.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/828/43a7baa05df0e549fa730cd9f9500417aa266ff1..42cc6a83e562672c39809f117847d76b4546fb87)*

src/bors/handlers/squash.rs
medzernik: EDIT: i see now :D I was confused because the refaactor put the blame on you instead of me, and it was a while ago. I'll try to fix the test as it's a leftover indeed

src/bors/command/parser.rs
```diff
@@ -670,6 +676,53 @@ fn parser_squash(command: &CommandPart<'_>, parts: &[CommandPart<'_>]) -> ParseR
     }
 }
 
+/// Parses `@bors r+ squash` command.
+/// Supports specifying a commit message via `@bors r+ squash [msg|message]="message"`.
+fn parser_squash_approve(_: &CommandPart<'_>, parts: &[CommandPart<'_>]) -> ParseResult {
```
Kobzol, line 681: This should still allow parsing priority, note, rollup mode, etc. I'd suggest modifying the approve parser, rather than creating a new one, and remembering `Option<SquashData>` while parsing it. Then if the data is `Some` after parsing is finished, return `SquashApprove`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/828/43a7baa05df0e549fa730cd9f9500417aa266ff1..33f8e575fdd45be73fa47bb99917eb203516acd4)*

medzernik: Approached this in: [cbf2076](https://github.com/rust-lang/bors/pull/828/commits/cbf2076beef7d43d3eb6531ede8bef636f95332c) but needs more work on error handling. However, is this the direction you want me to take the implementation?

PR #840, opened 2026-09-10
Add rollup auto build merge conflict detection

### LINKED ISSUE #635: Detect if a rollup has merge conflicts with the currently pending job
My rollup (https://github.com/rust-lang/rust/pull/151062) was unmergable due to conflicts with the `rollup=never` PR that was in queue before it, which was only detected when the rollup actually started getting tested.

This is low priority as this is the first time I've seen this happen, it's not a big deal (just make a new rollup without the conflicting PR) and I'm also not sure how easy this is to implement.

Sakib25800:
Closes #635.

When creating a rollup, we now check whether it conflicts with the pending auto build for the same base branch.

After constructing the rollup normally, bors temporarily merges the pending build commit into the generated `rollup-*` branch. If it conflicts, the rollup is still created with a warning in its description. Otherwise, the branch is reset to its original rollup commit.

The check is best effort, but rollup creation is aborted if we cannot safely restore the branch.

### DIFF
--- a/src/database/operations.rs
+++ b/src/database/operations.rs
@@ -780,6 +780,51 @@ WHERE repository = $1
     .await
 }
 
+pub(crate) async fn find_pending_auto_build(
+    executor: impl PgExecutor<'_>,
+    repo: &GithubRepoName,
+    base_branch: &str,
+) -> anyhow::Result<Option<BuildModel>> {
+    measure_db_query("find_pending_auto_build", || async {
+        let build = sqlx::query_as!(
+            BuildModel,
+            r#"
+SELECT
+    auto_build.id,
+    auto_build.repository AS "repository: GithubRepoName",
+    auto_build.pr_number AS "pr_number: PullRequestNumber",
+    auto_build.branch,
+    auto_build.kind AS "kind: BuildKind",
+    auto_build.commit_sha,
+    auto_build.status AS "status: BuildStatus",
+    auto_build.parent,
+    auto_build.created_at AS "created_at: DateTime<Utc>",
+    auto_build.check_run_id,
+    auto_build.duration AS "duration: PgDuration"
+FROM pull_request AS pr
+JOIN build AS auto_build ON pr.auto_build_id = auto_build.id
+WHERE pr.repository = $1
+    AND pr.base_branch = $2
+    AND pr.status = 'open'
+    AND pr.approved_by IS NOT NULL
+    AND pr.approved_sha IS NOT NULL
+    AND auto_build.repository = pr.repository
+    AND auto_build.pr_number = pr.number
+    AND auto_build.kind = $3
+    AND auto_build.status = $4
+"#,
+            repo as &GithubRepoName,
+            base_branch,
+            BuildKind::Auto as BuildKind,
+            BuildStatus::Pending as BuildStatus,
+        )
+        .fetch_optional(executor)
+        .await?;
+        Ok(build)
+    })
+    .await
+}
+
 pub(crate) async fn update_build(
     executor: impl PgExecutor<'_>,
     build_id: PrimaryKey,
--- a/src/github/rollup.rs
+++ b/src/github/rollup.rs
@@ -463,10 +463,8 @@ async fn has_pending_auto_build_conflict(
     pending_auto_build: &BuildModel,
     pending_auto_pr_number: PullRequestNumber,
 ) -> Result<bool, RollupError> {
-    let merge_message = format!(
-        "Rollup compatibility check against pending auto build #{}",
-        pending_auto_pr_number
-    );
+    let merge_message =
+        format!("Rollup compatibility check against pending auto build #{pending_auto_pr_number}");
     // Merge the pending auto build's commit into the rollup branch
     match client
         .merge_branches(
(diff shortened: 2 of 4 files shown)


src/github/rollup.rs
```diff
@@ -383,6 +383,33 @@ async fn create_rollup(
         }
     }
 
+    match db.find_pending_auto_build(&base_repo, &base_branch).await {
+        Ok(Some(pending_auto_build)) => {
+            if let Some(rollup_head) = successes.last()
+                && let Some(pending_auto_pr_number) = pending_auto_build.pr_number
+                && has_pending_auto_build_conflict(
```
Sakib25800, line 390: Note: this ends up a quite deeply indented, but keeping the optional lookup and guard conditions together seemed the cleanest. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/840/1c873260e550cab7a762dc69af103ef7aeef4d1b..8bd6ed663aa451a75aa471e7522018aa214b563a)*

src/database/operations.rs
```diff
@@ -780,6 +780,51 @@ WHERE repository = $1
     .await
 }
 
+pub(crate) async fn find_pending_auto_build(
```
Kobzol: Rather than creating a new query for this, can you just reuse `get_pending_builds`? It should be fine to download all the pending builds, there are usually not many of them, and rollup creation is very infrequent.

src/github/rollup.rs
```diff
@@ -383,6 +383,33 @@ async fn create_rollup(
...
+                    &rollup_branch,
+                    &rollup_head.rolled_up_merge_sha,
+                    &pending_auto_build,
+                    pending_auto_pr_number,
+                )
+                .await?
+            {
+                write!(
+                    body,
+                    "\nThis rollup conflicts with pending auto build #{pending_auto_pr_number} and may need to be recreated if it succeeds.\n",
```
Kobzol:
```suggestion
                    "\nThis rollup conflicts with pending auto build #{pending_auto_pr_number} and may need to be recreated if the pending build succeeds.\n",
```

src/github/rollup.rs
```diff
@@ -428,6 +455,58 @@ async fn create_rollup(
     Ok(pr)
 }
 
+/// Temporarily merges the pending auto build's commit into the rollup branch
```
Kobzol, line 498:
```suggestion
/// Temporarily merges the pending auto build's commit into the rollup branch
/// Returns Ok(true) if there is a conflict between the current rollup branch and the pending build.
```

src/github/rollup.rs
```diff
@@ -428,6 +455,58 @@ async fn create_rollup(
...
+            &merge_message,
+        )
+        .await
+    {
+        Err(MergeError::Conflict) => {
+            tracing::info!(
+                pending_pr = %pending_auto_pr_number,
+                "Rollup conflicts with the pending auto build"
+            );
+            return Ok(true);
```
Kobzol: We should always revert the rollup to the original state, it should not contain the pending build code.

Kobzol (commented):
Thank you for the PR! Left a few comments.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/840/1c873260e550cab7a762dc69af103ef7aeef4d1b..24d1db7e7254eb6d2b7aae921a10d0f3f91c8671)*

src/database/operations.rs
Sakib25800: Yep

src/github/rollup.rs
Sakib25800: I figured that if a conflict occurs then there's nothing to revert?

Kobzol: Ah, of course, you're right. Nevermind!

Kobzol: That being said, I wouldn't be surprised if GitHub sometimes told us that there was an issue, but actually did a merge, so this is a safer approach.

src/github/rollup.rs
```diff
@@ -428,6 +454,102 @@ async fn create_rollup(
...
+    {
+        Err(MergeError::Conflict) => {
+            tracing::info!(
+                pending_pr = %pending_auto_pr_number,
+                "Rollup conflicts with the pending auto build"
+            );
+            true
+        }
+        Ok(_) | Err(MergeError::AlreadyMerged) => {
+            tracing::info!(
```
Kobzol, line 533: This log is misleading, because it is also fired for the `Ok(_)` case (found while testing locally). If the merge was successful, we should either not print anything, or print that it was OK.

Kobzol (commented):
I had some troubles getting this to work locally, before I noticed that I was trying to generate a conflict on two different files (damned invisible left-to-right UTF8 markers copied from GitHub, uhh).

It seems to work great, thank you!

Could you please put the conflict message in the rollup message into a warning block? To make it more visible.

> [!WARNING]
> This rollup conflicts with pending auto build #68 and may need to be recreated if the pending build succeeds.

```
> [!WARNING]
> This rollup conflicts with pending auto build #68 and may need to be recreated if the pending build succeeds.
```

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/840/1c873260e550cab7a762dc69af103ef7aeef4d1b..0408d42e70099f5b5694f10840fb24c042065388)*

Kobzol (approved):
Thank you very much, let's try this!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/840/1c873260e550cab7a762dc69af103ef7aeef4d1b..d8ce766e2da17cfae330f4c116787c3ad8267722)*

PR #841, opened 2026-09-13
Update dependencies & cleanup `Cargo.toml`

medzernik:
I made a small PR to:
1. lock in `octocrab` and group it with `jsonwebtoken` as `octocrab` requires the same version of internally used `jsonwebtoken` and does not reexport it (this may be just my mistaken view). Lock both to specific patch versions as the latest patch version of `octocrab` currently used version breaks tests
2. add repository metadata, I think it should be there for people to find the repo easier.
3. update Dockerfile to `FROM RUST:1.98.1` as when developing the version used is usually latest.
4. update insta tests to new format, eliminating the errors
5. update the `rust-team` dependency with a breaking change - added `google_workspace_saml_group: None` for now, please let me know if this needs to be set to smth else
6. finally, update all other dependencies via cargo update, which with the currently locked deps should make it easier to keep everything non-breaking up to date :)

Hope this is not too much or too broad, will gladly change things around, but thought this should help things run smoothly as well!

### DIFF
--- a/src/github/rollup.rs
+++ b/src/github/rollup.rs
@@ -1049,9 +1049,7 @@ also include this pls"
 
             [View changes since this unapproval](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/borstest/3/main-sha1..pr-3-sha)
             ");
-            insta::assert_snapshot!(ctx.get_next_comment_text(4).await?, @"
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Thank you! :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/841/1c873260e550cab7a762dc69af103ef7aeef4d1b..46541ec507f3d2517adec934252699ce4b1fa6be)*

PR #848, opened 2026-09-18
Fix concurrent unroll processing

Sakib25800:
The `unroll_concurrent_bors_instances` test was flaky because concurrent instances could read stale member state, so we need to move the advisory lock before the query.


Kobzol (approved):
Hehe, I noticed the flaky test, but didn't yet get around to fixing it yet, and now you did it in the meantime. Wonderful ❤️ Good find, and thank you.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/bors/848/6550997bc3cc1e12f78853bc8f20dc953270290e..567e36cca7b650fb8862f85efbaf567af9c2bb62)*

